PR #66, opened 2020-06-22
Feature/gha workflow base

vimanyu:
- "Windows only" github workflow for x64 Release desktop build using Visual Studio 2019
- Separate workflow for "linux/mac/win" desktop builds (x64 Release)
- using vcpkg as a cross-platform package manager for those dependencies that are not downloaded during cmake generation process
- using github actions caching for packages installed using vcpkg (saves around 10 to 20 mins depending on platform)
- ccache speed up for linux/mac builds (dramatically reduces build times 10 mins -> 2 mins on linux and 10 mins -> 5 mins on mac)

### DIFF
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -138,7 +138,7 @@ jobs:
             -B build \
             -G "${{ matrix.generator }}" \
             -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
-            -DCMAKE_TOOLCHAIN_FILE=vcpkg/scripts/buildsystems/vcpkg.cmake \
+            -DCMAKE_TOOLCHAIN_FILE=external/vcpkg/scripts/buildsystems/vcpkg.cmake \
             -DVCPKG_TARGET_TRIPLET=${{ matrix.vcpkg_triplet }} \
 
       - name: Build
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -51,6 +51,7 @@ jobs:
           echo "::set-env name=VCPKG_RESPONSE_FILE::external/vcpkg_${{ matrix.vcpkg_triplet }}_response_file.txt"
           echo "::set-env name=PIP_REQUIREMENTS_FILE::external/pip_requirements.txt"
           echo "::set-env name=MATRIX_UNIQUE_NAME::${{ matrix.os }}-${{ matrix.build_type }}-${{ matrix.architecture }}-${{ matrix.python_version }}-${{ matrix.generator }}"
+          echo "test if ccache cache works"
 
       - name: Install Ninja (windows)
         if: startsWith(matrix.generator, 'Ninja') && startsWith(matrix.os, 'windows')
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -76,7 +76,7 @@ jobs:
 
       - name: Install system level applications (linux)
         if: startsWith(matrix.os, 'ubuntu')
-        id: ubuntu-brew-install
+        id: ubuntu-apt-install
         run: |
           sudo apt install ccache
 
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -51,7 +51,6 @@ jobs:
           echo "::set-env name=VCPKG_RESPONSE_FILE::external/vcpkg_${{ matrix.vcpkg_triplet }}_response_file.txt"
           echo "::set-env name=PIP_REQUIREMENTS_FILE::external/pip_requirements.txt"
           echo "::set-env name=MATRIX_UNIQUE_NAME::${{ matrix.os }}-${{ matrix.build_type }}-${{ matrix.architecture }}-${{ matrix.python_version }}-${{ matrix.generator }}"
-          echo "test if ccache cache works"
 
       - name: Install Ninja (windows)
         if: startsWith(matrix.generator, 'Ninja') && startsWith(matrix.os, 'windows')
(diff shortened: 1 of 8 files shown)


.github/workflows/desktop.yml
```diff
@@ -0,0 +1,160 @@
...
+  CCACHE_DIR: ${{ github.workspace }}/ccache_dir
+
+jobs:
+  build:
+    name: ${{ matrix.os }}-${{ matrix.build_type }}-${{ matrix.architecture }}-${{ matrix.python_version }}-${{ matrix.generator }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [windows-latest, ubuntu-latest, macos-latest]
```
DellaBitta, line 17: is "-latest" required here?   Are we going to run on older versions that aren't latest?  If so maybe we can just tag those as older versions instead of marking all of these with "-latest".  It makes the matrix output quite verbose.

.github/workflows/desktop.yml
```diff
@@ -0,0 +1,160 @@
...
+
+      - name: Install system level applications (mac)
+        if: startsWith(matrix.os, 'macos')
+        id: macos-brew-install
+        run: |
+          brew install ccache
+
+      - name: Install system level applications (linux)
+        if: startsWith(matrix.os, 'ubuntu')
+        id: ubuntu-brew-install
```
DellaBitta, line 79: should this be "apt" not "brew"?

.github/workflows/desktop.yml
```diff
@@ -0,0 +1,160 @@
...
+          key: dev-vcpkg-${{ matrix.vcpkg_triplet }}-${{ hashFiles(format('{0}', env.VCPKG_RESPONSE_FILE)) }}-${{ hashFiles('.git/modules/external/vcpkg/HEAD') }}
+
+      - name: Install vcpkg packages (windows)
+        if: steps.cache_vcpkg.outputs.cache-hit!='true' && startsWith(matrix.os, 'windows')
+        run: |
+          ${{ github.workspace }}/external/vcpkg/bootstrap-vcpkg.bat -disableMetrics
+          ${{ github.workspace }}/external/vcpkg/vcpkg install @${{env.VCPKG_RESPONSE_FILE}}
+
+      - name: Install vcpkg packages (mac and linux)
+        if: steps.cache_vcpkg.outputs.cache-hit!='true' && startsWith(matrix.os, 'windows')!='true'
```
DellaBitta, line 97: Right now we're not doing any cross compilation, but will doing a check on the OS for 'not windows' be a problem later if we add other OS targets like Android and iOS that might be running on a host Windows VM?  Or maybe there should be a host_os and target_os in the matrix?

.github/workflows/desktop.yml
```diff
@@ -0,0 +1,160 @@
...
+            -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
+            -DCMAKE_TOOLCHAIN_FILE=external/vcpkg/scripts/buildsystems/vcpkg.cmake \
+            -DVCPKG_TARGET_TRIPLET=${{ matrix.vcpkg_triplet }} \
+
+      - name: Build
+        shell: bash
+        run: cmake --build build --config ${{ matrix.build_type }} -j 8
+
+      - name: Stats for ccache (mac and linux)
+        if: startsWith(matrix.os, 'macos') || startsWith(matrix.os, 'ubuntu')
```
DellaBitta, line 150: Instead of checking for the OSes here and assuming that cmake is enabled / installed, maybe make the "- name: Install system level applications (linux)" and "- name: Install system level applications (macos)" steps (above) set a flag indicating that cmake WAS installed and that this check to just be for that flag.  Is that possible?

.github/workflows/desktop.yml
vimanyu: So, github has labels for various machine configurations for their runners. I think "-latest" is on relatively more stable and tested configurations. For example, Ubuntu 18 is ubuntu-latest but they also have Ubuntu-20. I think at the very least, we should be using all the "latest" ones and maybe specifically add any newer versions if need be.

https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners

.github/workflows/desktop.yml
vimanyu: Yup! Good catch :)! I copy/pasted the mac block and forgot to fix the "id"

.github/workflows/desktop.yml
vimanyu: Excellent point. Yes, let me look into this. I am certain I can set a global environment variable during the installation but I am not sure if I can use that in an "if" expression. Going to give this a try.

.github/workflows/desktop.yml
vimanyu: Excellent point. Yes, let me look into this. I am certain I can set a global environment variable during the installation but I am not sure if I can use that in an "if" expression. Going to give this a try.

.github/workflows/desktop.yml
DellaBitta: Gotcha, ok, thanks.

.github/workflows/desktop.yml
vimanyu: So I was planning to have a separate workflow for Android and iOS. We should discuss how we organize these github workflows. Because of similarity in steps, I was thinking of organizing them as follows,

- Desktop workflow (builds for windows/mac/linux desktop)
- Android workflow (windows/linux/mac builds for Android)
- iOS workflow (mac build for iOS)

DellaBitta: Ah right, this file itself is in the "desktop" classifcation.  Got it

.github/workflows/desktop.yml
vimanyu: So this experiment was successful.

```
        # Setting an env variable right after installing ccache that can be reused in later stages

      - name: Install system level applications (linux)
        if: startsWith(matrix.os, 'ubuntu')
        id: ubuntu-apt-install
        run: |
          sudo apt install ccache
          echo "::set-env name=CCACHE_INSTALLED::1"

      - name: Cache ccache files
        if: env.CCACHE_INSTALLED
        id: cache_ccache
```

PR #71, opened 2020-06-26
Add encrypted google services files.

cynthiajoan:
Add encrypted files for integration test


vimanyu:
+1 to a document detailing the expected workflow.
I noticed a decrypt file and am guessing the passphrase is stored as a secret on Github. So through Github actions, we should be able decrypt these files. It might be worth testing if a forked repo will have access to these secrets?

paulb777:
A forked repo does NOT have access. See examples for addressing this in https://github.com/firebase/firebase-ios-sdk/tree/master/.github/workflows

cynthiajoan:
> A forked repo does NOT have access. See examples for addressing this in https://github.com/firebase/firebase-ios-sdk/tree/master/.github/workflows

Is this the commit that addresses this? https://github.com/firebase/firebase-ios-sdk/commit/de91df623b082db270be1029078a8c4cf13d0ddd

paulb777:
@cynthiajoan That commit is to stop running push jobs on the private GitHub mirror since they were mostly redundant and especially expensive.

Managing forks or not is done by checking the secret environment variable like https://github.com/firebase/firebase-ios-sdk/blob/master/.github/workflows/auth.yml#L100

vimanyu:
Just a note for when we incorporate this into a Github workflow. We might have to install "gpg" in a separate step. I think the iOS repo doesnt run into this because out of all the runners, mac is the only one with "gpg" pre-installed. 

https://help.github.com/en/actions/reference/software-installed-on-github-hosted-runners

PR #88, opened 2020-07-17
Feature/chk fix callback test

chkuang-g:
Fix the racy callback test which causes test failure on Windows platform
https://github.com/firebase/firebase-cpp-sdk/runs/846731160


chkuang-g:
@googlebot I fixed it.

vimanyu:
I merged Shawn's changes into my local branch and ran the Github workflows twice to test this and I didn't see this error in the 2 runs.

chkuang-g:
I ran 10000 times using msvc framework and did not see any crash :)

PR #100, opened 2020-07-22
Regenerate secret files and add two more

cynthiajoan:


cynthiajoan:
> Why are some of these binary and some ascii?

Regenerate to make sure all files are binary

PR #101, opened 2020-07-23
Feature/prereq and build scripts

vimanyu:
Python script to install prerequisites (python and cpp dependencies) and to build the cpp sdk. 
This greatly simplifies all the branching and special conditions in Github workflow file.

### DIFF
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -85,7 +85,7 @@ jobs:
         run: |
           python scripts/install_prereqs.py --ccache --generator "${{ matrix.generator}}"
 
-      - name: CMake configure and build
+      - name: Build SDK
         run: |
           python scripts/build.py --build_tests
 
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -68,6 +68,14 @@ jobs:
           path: external/vcpkg/installed
           key: dev-vcpkg-${{ matrix.vcpkg_triplet }}-${{ hashFiles(format('{0}', env.VCPKG_RESPONSE_FILE)) }}-${{ hashFiles('.git/modules/external/vcpkg/HEAD') }}
 
+      - name: Cache ccache files
+        if: startsWith(matrix.os, 'ubuntu') || startsWith(matrix.os, 'macos')
+        id: cache_ccache
+        uses: actions/cache@v2
+        with:
+          path: ccache_dir
+          key: dev-test-ccache-${{ env.MATRIX_UNIQUE_NAME }}
+
       - name: Setup python
         uses: actions/setup-python@v2
         with:
@@ -78,35 +86,12 @@ jobs:
         run: |
           python scripts/install_prereqs.py --ccache --generator "${{ matrix.generator}}"
 
-      - name: Cache ccache files
-        if: env.CCACHE_INSTALLED
-        id: cache_ccache
-        uses: actions/cache@v2
-        with:
-          path: ccache_dir
-          key: dev-test-ccache-${{ env.MATRIX_UNIQUE_NAME }}
-
-      - name: Configure 
-        shell: bash
+      - name: CMake configure and build
         run: |
-          mkdir build
-          cmake \
-            -S . \
-            -B build \
-            -G "${{ matrix.generator }}" \
-            -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
-            -DCMAKE_TOOLCHAIN_FILE=external/vcpkg/scripts/buildsystems/vcpkg.cmake \
-            -DVCPKG_TARGET_TRIPLET=${{ matrix.vcpkg_triplet }} \
-            -DFIREBASE_CPP_BUILD_TESTS=ON \
-            -DFIREBASE_INCLUDE_FIRESTORE=OFF \
-            -DFIREBASE_FORCE_FAKE_SECURE_STORAGE=ON
-
-      - name: Build
-        shell: bash
-        run: cmake --build build --config ${{ matrix.build_type }} -j 8
+          python scripts/build.py --build_tests
 
       - name: Stats for ccache (mac and linux)
-        if: env.CCACHE_INSTALLED
+        if: startsWith(matrix.os, 'ubuntu') || startsWith(matrix.os, 'macos')
         run: ccache -s
 
       - name: Print built libraries
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -70,7 +70,7 @@ jobs:
 
       - name: Install prerequisites(cpp and python dependencies)
         run: |
-          python scripts/install_prereqs.py --ccache --generator ${{ matrix.generator}}
+          python scripts/install_prereqs.py --ccache --generator "${{ matrix.generator}}"
 
       - name: Cache ccache files
         if: env.CCACHE_INSTALLED
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -3,6 +3,7 @@ name: Desktop Builds
 on: 
   pull_request:
     types: [opened, reopened, synchronize]
+  push:
 
 env:
   CCACHE_DIR: ${{ github.workspace }}/ccache_dir
@@ -60,35 +61,6 @@ jobs:
         run: |
           echo "::set-env name=BUILD_GENERATOR_NO_SPACES::$(echo -e "${{matrix.generator}}" | tr -d '[:space:]')"
 
-      - name: Install Ninja (windows)
-        if: startsWith(matrix.generator, 'Ninja') && startsWith(matrix.os, 'windows')
-        run: |
-          choco install ninja
-
-      - name: Install Ninja (mac)
-        if: startsWith(matrix.generator, 'Ninja') && startsWith(matrix.os, 'macos')
-        run: |
-          brew install ninja
-
-      - name: Install Ninja (linux)
-        if: startsWith(matrix.generator, 'Ninja') && startsWith(matrix.os, 'ubuntu')
-        run: |
-          sudo apt install ninja-build
-
-      - name: Install system level applications (mac)
-        if: startsWith(matrix.os, 'macos')
-        id: macos-brew-install
-        run: |
-          brew install ccache
-          echo "::set-env name=CCACHE_INSTALLED::1"
-
-      - name: Install system level applications (linux)
-        if: startsWith(matrix.os, 'ubuntu')
-        id: ubuntu-apt-install
-        run: |
-          sudo apt install ccache
-          echo "::set-env name=CCACHE_INSTALLED::1"
-
       - name: Cache vcpkg C++ dependencies
         id: cache_vcpkg
         uses: actions/cache@v2
@@ -96,40 +68,9 @@ jobs:
           path: external/vcpkg/installed
           key: dev-vcpkg-${{ matrix.vcpkg_triplet }}-${{ hashFiles(format('{0}', env.VCPKG_RESPONSE_FILE)) }}-${{ hashFiles('.git/modules/external/vcpkg/HEAD') }}
 
-      - name: Install vcpkg packages (windows)
-        if: steps.cache_vcpkg.outputs.cache-hit!='true' && startsWith(matrix.os, 'windows')
+      - name: Install prerequisites(cpp and python dependencies)
         run: |
-          ${{ github.workspace }}/external/vcpkg/bootstrap-vcpkg.bat -disableMetrics
-          ${{ github.workspace }}/external/vcpkg/vcpkg install @${{env.VCPKG_RESPONSE_FILE}}
-
-      - name: Install vcpkg packages (mac and linux)
-        if: steps.cache_vcpkg.outputs.cache-hit!='true' && startsWith(matrix.os, 'windows')!='true'
-        run: |
-          ${{ github.workspace }}/external/vcpkg/bootstrap-vcpkg.sh -disableMetrics
-          ${{ github.workspace }}/external/vcpkg/vcpkg install @${{env.VCPKG_RESPONSE_FILE}}
-
-      - name: Setup python
-        uses: actions/setup-python@v2
-        with:
-          python-version: ${{ matrix.python_version }}
-          architecture: ${{ matrix.architecture }}
-
-      - name: Get pip cache dir
-        id: get_pip_cache_dir
-        run: |
-          echo "::set-output name=cachedir::$(pip cache dir)"
-
-      - name: pip cache
-        id: cache_pip
-        uses: actions/cache@v2
-        with:
-          path: ${{ steps.get_pip_cache_dir.outputs.cachedir }}
-          key: dev-pip-${{ runner.os }}-${{ matrix.python_version }}-${{ matrix.architecture }}-${{ hashFiles(format('{0}', env.PIP_REQUIREMENTS_FILE)) }}
-
-      - name: Install pip packages
-        run: |
-          python -m pip install --upgrade pip
-          pip install -r ${{ env.PIP_REQUIREMENTS_FILE }}
+          python scripts/install_prereqs.py
 
       - name: Cache ccache files
         if: env.CCACHE_INSTALLED
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -18,7 +18,7 @@ jobs:
         os: [windows-latest, ubuntu-latest, macos-latest]
         build_type: ["Release", "Debug"]
         architecture: ["x64",]
-        python_version: [2.7]
+        python_version: [3.6]
         generator: ["Visual Studio 16 2019", "Unix Makefiles", "Ninja"]
         exclude:
         - os: ubuntu-latest
@@ -67,6 +67,12 @@ jobs:
         with:
           path: external/vcpkg/installed
           key: dev-vcpkg-${{ matrix.vcpkg_triplet }}-${{ hashFiles(format('{0}', env.VCPKG_RESPONSE_FILE)) }}-${{ hashFiles('.git/modules/external/vcpkg/HEAD') }}
+
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python_version }}
+          architecture: ${{ matrix.architecture }}
 
       - name: Install prerequisites(cpp and python dependencies)
         run: |
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -3,7 +3,6 @@ name: Desktop Builds
 on: 
   pull_request:
     types: [opened, reopened, synchronize]
-  push:
 
 env:
   CCACHE_DIR: ${{ github.workspace }}/ccache_dir
@@ -118,7 +117,7 @@ jobs:
         run: |
           rm -rf ~/Library/Logs/DiagnosticReports/*
           
-      - name: Run tests (windows & macos)
+      - name: Run unit tests (windows & macos)
         if: startsWith(matrix.os, 'windows') || startsWith(matrix.os, 'macos')
         env:
           LANG: en_US
@@ -126,7 +125,7 @@ jobs:
           cd build
           ctest --verbose
 
-      - name: Run tests (linux)
+      - name: Run unit tests (linux)
         # Linux exists as its own standalone execution step in order to invoke
         # platform-specific `ulimit` to enable crash collection.  The ulimit
         # command must be invoked in same shell instance of that runs the
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -87,7 +87,7 @@ jobs:
 
       - name: Build SDK
         run: |
-          python scripts/build.py --build_tests
+          python scripts/build.py --build_tests --arch "${{ matrix.architecture }}" --config "${{ matrix.build_type }}"
 
       - name: Stats for ccache (mac and linux)
         if: startsWith(matrix.os, 'ubuntu') || startsWith(matrix.os, 'macos')
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -70,7 +70,7 @@ jobs:
 
       - name: Install prerequisites(cpp and python dependencies)
         run: |
-          python scripts/install_prereqs.py --cmake --generator ${{ matrix.generator}}
+          python scripts/install_prereqs.py --ccache --generator ${{ matrix.generator}}
 
       - name: Cache ccache files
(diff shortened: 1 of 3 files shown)


.github/workflows/desktop.yml
```diff
@@ -60,106 +60,37 @@ jobs:
...
-      - name: pip cache
-        id: cache_pip
-        uses: actions/cache@v2
-        with:
-          path: ${{ steps.get_pip_cache_dir.outputs.cachedir }}
-          key: dev-pip-${{ runner.os }}-${{ matrix.python_version }}-${{ matrix.architecture }}-${{ hashFiles(format('{0}', env.PIP_REQUIREMENTS_FILE)) }}
+          python scripts/install_prereqs.py --ccache --generator "${{ matrix.generator}}"
 
-      - name: Install pip packages
+      - name: CMake configure and build
```
jonsimantov, line 88: nit: Just "Build SDK" or similar is fine, no need to specify cmake here since we're just running the build script.

.github/workflows/desktop.yml
```diff
@@ -60,106 +60,37 @@ jobs:
...
-            -DCMAKE_TOOLCHAIN_FILE=external/vcpkg/scripts/buildsystems/vcpkg.cmake \
-            -DVCPKG_TARGET_TRIPLET=${{ matrix.vcpkg_triplet }} \
-            -DFIREBASE_CPP_BUILD_TESTS=ON \
-            -DFIREBASE_INCLUDE_FIRESTORE=OFF \
-            -DFIREBASE_FORCE_FAKE_SECURE_STORAGE=ON
-
-      - name: Build
-        shell: bash
-        run: cmake --build build --config ${{ matrix.build_type }} -j 8
+          python scripts/build.py --build_tests
```
jonsimantov, line 90: Here is really where I'd like to see this be much more explicit. Something like...

build.py --platform=windows --arch=x64 --type=debug --runtime=md

(where obviously those come from the matrix, similar to how the ios build script is run in my PR).

jonsimantov: (because right now how does it even get those options?)

scripts/build.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
vimanyu: I was keeping the structure consistent with `firebase-ios-sdk` so there is some sort of familiarity between different firebase repos.

scripts/build.py
```diff
@@ -0,0 +1,213 @@
...
+    return sys.platform.startswith('linux')
+
+
+def is_mac_os():
+    """Check if the current OS is mac
+
+    Returns:
+        (bool): True if windows, False otherwise
+    """
+    return sys.platform.startswith('darwin')
```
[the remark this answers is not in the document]
vimanyu: I really contemplated doing that but that adds another "dependency" and these scripts are not standalone anymore. My goal was to have these scripts as completely standalone and users could execute it from any subdirectory in the repo and it just works.

scripts/build.py
```diff
@@ -0,0 +1,213 @@
...
+
+def main():
+    args = parse_cmdline_args()
+
+    step_cmake_configure(args)
+    step_cmake_build(args)
+
+
+def parse_cmdline_args():
+    parser = argparse.ArgumentParser(description='Install Prerequisites for building cpp sdk')
```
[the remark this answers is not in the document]
vimanyu: Unfortunately, we wouldn't be able to since it isnt available when the python interpretor starts. absl-py is installed as part of this script.

scripts/install_prereqs.py
```diff
@@ -0,0 +1,453 @@
...
+    step_install_python_packages(args)
+    step_install_cpp_packages(args)
+
+
+def parse_cmdline_args():
+    parser = argparse.ArgumentParser(description='Install Prerequisites for building cpp sdk')
+    parser.add_argument('-A', '--arch', default='x64', help='Platform architecture (x64, x86)')
+    parser.add_argument('-G', '--generator', help='Build generator to use (install ninja if needed)')
+    parser.add_argument('-c', '--ccache', action='store_true', help='Install ccache using system package manager'\
+                                                                    '(only supported on mac and linux)')
```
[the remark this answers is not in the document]
vimanyu: So I tried to keep it as simple as possible but this is the bare minimum required for prereqs. The reasons are,
- architecture determines the "vcpkg triplet". Basically vcpkg build step needs to know which architecture/platform it is building for.
-  --generator and --ccache : These were included only to support Ninja and ccache. If we dont plan to use Ninja, we can get rid of "--generator". The prereqs script will install these two tools via a system package manager because they are not available in vcpkg. 

Having said this, the script can run without any command line flags whatsoever and it will work. (it assumes default arch of x64 and it just won't install ninja and ccache)

scripts/install_prereqs.py
```diff
@@ -0,0 +1,453 @@
...
+    parser.add_argument('-p', '--no_vcpkg', action='store_true', help='Use system package manager (brew, apt, choco) for dependencies(instead of vcpkg)')
```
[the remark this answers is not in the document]
vimanyu: So, we could keep it simple and always use "vcpkg" but I was thinking more from a user using these scripts locally. Didnt want to force using vcpkg incase they have already installed cpp dependencies in some other way. Am expecting these scripts to be used in dev workflows by users but we could keep it simple and stick to vcpkg.

scripts/install_prereqs.py
```diff
@@ -0,0 +1,453 @@
...
+        found_vcpkg_executable = os.path.exists(get_vcpkg_executable_file_path())
+
+        # Install packages and only install vcpkg if its not built already
+        install_packages_with_vcpkg(args.arch, build_vcpkg=not found_vcpkg_executable)
+
+    else:
+        # Do not install dependencies automatically as we want users to have control
+        # on their system level package installations
+        print('Please install packages listed in {0}'\
+              'with your system package manager'.format(get_vcpkg_response_file_path(get_vcpkg_triplet(args.arch))))
```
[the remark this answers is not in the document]
vimanyu: I agree things would be simple if we decide to just use vcpkg.
Most of the code is an optimization. It will shave off some time (few minutes) on Github runners and super helpful for local dev workflows.

.github/workflows/desktop.yml
```diff
@@ -17,7 +17,7 @@ jobs:
         os: [windows-latest, ubuntu-latest, macos-latest]
         build_type: ["Release", "Debug"]
         architecture: ["x64",]
-        python_version: [2.7]
+        python_version: [3.6]
         generator: ["Visual Studio 16 2019", "Unix Makefiles", "Ninja"]
```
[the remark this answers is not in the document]
vimanyu: "Ninja" was there just to make sure that things work with Ninja incase users are using it. 
If we decide to not run our workflows against "Ninja", we can get rid of this `generator`. And we dont have to do aynthing in the build script because cmake will pick the default generator based on the platform.

.github/workflows/desktop.yml
vimanyu: Makes sense.

.github/workflows/desktop.yml
vimanyu: Good catch. I forgot to add --arch=<architecture> and --config=<config> to this step. 
Its working right now because,
- platform is easy to detect in the script
- arch has a default value of 'x64' if its not specified (but we should keep it explicit here)
- type defaults to Debug in cmake (but we should specify it here)
- runtime (I havent gotten that far to support runtime in Desktop builds yet. Yes, we should have the runtime specified too)

.github/workflows/desktop.yml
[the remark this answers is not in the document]
vimanyu: Oh and I was curious to know if Ninja builds faster than other generators. So yeah, we could technically take Ninja off the list and that simplifies the build script too. Will still need an additional step of installing ccache though, else we could do everything locally in the repo with vcpkg and not touch any system level packages whatsoever.

scripts/build.py
```diff
@@ -0,0 +1,213 @@
+#!/usr/bin/env python
+
+# Copyright 2019 Google
```
DellaBitta, line 3: This should be 2020.

scripts/build.py
```diff
@@ -0,0 +1,213 @@
...
+import argparse
+
+
+def run_command(cmd, cwd=None, as_root_user=False, collect_output=False):
+    """Run a command
+
+    Args:
+        cmd (:obj:`list` of :obj:`str`): Command to run as a list object
+                                         Eg: ['ls', '-l']
+        cwd (str): Directory to execute the command from
```
DellaBitta, line 47: nit: all of these comments should all end with a period (.)

scripts/build.py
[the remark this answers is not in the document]
DellaBitta: I went into this review thinking that this WAS the scripting file that all platforms would use and its behavior is dicated by command line flags.  Is this not the case?

scripts/build.py
```diff
@@ -0,0 +1,213 @@
...
+    """Check if the current OS is mac
+
+    Returns:
+        (bool): True if windows, False otherwise
+    """
+    return sys.platform.startswith('darwin')
+
+
+def get_vcpkg_triplet(arch):
+    """ Get vcpkg target triplet (platform definition)
```
DellaBitta, line 117: We should distinguish between target and host platforms since we'll soon be cross compiling from various host desktop platforms to various target Android platforms.  Applies to the next file, too.

scripts/build.py
```diff
@@ -0,0 +1,213 @@
...
+    Returns:
+        (str): Triplet name
+               Eg: "x64-windows-static"
+    """
+    triplet_name = [arch]
+    if is_windows_os():
+        triplet_name.append('windows')
+        # For windows, default is to build dynamic. Hence we specify static.
+        # For mac/linux, default is to build static libraries
+        triplet_name.append('static')
```
DellaBitta, line 134: Could we output these decisions to the console so that any automatic configurations are evident in the build results?

scripts/build.py
```diff
@@ -0,0 +1,213 @@
...
+        cmd.append('-DFIREBASE_FORCE_FAKE_SECURE_STORAGE=ON')
+    # If you want to avoid vcpkg altogether and have cpp dependencies installed elsewhere
+    if not args.no_vcpkg:
+        vcpkg_toolchain_file_path = os.path.join(REPO_ROOT_DIR, 'external', 'vcpkg', 
+                                                'scripts', 'buildsystems', 'vcpkg.cmake')
+        
+        cmd.append('-DCMAKE_TOOLCHAIN_FILE={0}'.format(vcpkg_toolchain_file_path))
+        cmd.append('-DVCPKG_TARGET_TRIPLET={0}'.format(get_vcpkg_triplet(args.arch)))
+    
+    cmd.append('-DFIREBASE_INCLUDE_FIRESTORE=OFF')
```
DellaBitta, line 174: Shouldn't this be a parameter we pass into the script?  Why is a lack of vkcpkg explicit to a lack of vcpkg, and why would a user know if/when Firestore is distabled?  Seems like a warning message on the console is warranted at the very least.

scripts/build.py
```diff
@@ -0,0 +1,213 @@
...
+    step_cmake_build(args)
+
+
+def parse_cmdline_args():
+    parser = argparse.ArgumentParser(description='Install Prerequisites for building cpp sdk')
+    parser.add_argument('-A', '--arch', default='x64', help='Platform architecture (x64, x86)')
+    parser.add_argument('-G', '--generator', help='Build generator to use')
+    parser.add_argument('--build_dir', default='build', help='Output build directory')
+    parser.add_argument('--build_tests', action='store_true', help='Build unit tests too')
+    parser.add_argument('--no_vcpkg', action='store_true', help='Do not use vcpkg toolchain')
```
DellaBitta, line 205: Is it just the toolchain or all of the other dependencies, too?  Maybe this be something more akin to why it would be used and not expiclity name vcpkg.  Something like --custom_tool_config or something.,

scripts/install_prereqs.py
```diff
@@ -0,0 +1,453 @@
+#!/usr/bin/env python
+
+# Copyright 2019 Google
```
DellaBitta, line 3: 2020

scripts/build.py
```diff
@@ -0,0 +1,213 @@
...
+    parser.add_argument('--config', help='Release/Debug config')
+    parser.add_argument('--target', nargs='+', help='A list of CMake build targets (eg: firebase_app firebase_auth)')
+    parser.add_argument('-j', default=multiprocessing.cpu_count(), help='Number of CPU cores to use for build (default=max)')
+    args = parser.parse_args()
+    return args
+
+if __name__ == '__main__':
+    main()
```
DellaBitta, line 213: nit:  add a new line here.

scripts/build.py
vimanyu: Ignore the Firestore bit. That is a temporary placeholder for my test workflows to run without errors. Once we iron out the last few unit test errors, Firestore should be enabled by default.  I should have made a comment. 
re: vcpkg, this was placed to make the script more flexible incase uesrs didnt want to use vcpkg.

scripts/build.py
vimanyu: So this option was to avoid using any of vcpkg stuff and have users just build without it. This assumes users used some other package manager to install cpp dependencies and everything is accessible by the build system.

.github/workflows/desktop.yml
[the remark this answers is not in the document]
jonsimantov: Yeah I think since Ninja doesn't make a difference, i'm fine with just testing against the default. If a developer wants to use Ninja they can be responsible for whether it works themselves, I'd rather not have to support more than 1 generator per platform.

scripts/build.py
jonsimantov: Again, I think this script's primary purpose is to install prereqs on GHA runners and build on those runners. I don't think it should support every use case. If a developer wants to NOT use vcpkg they can go ahead and set up the prereqs themselves rather than running our own prereqs script.

scripts/install_prereqs.py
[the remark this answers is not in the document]
jonsimantov: architecture: The idea of breaking up the script into prereq / build is that you install_prereqs once per system, and then can build multiple times for different architectures. This means that if vcpkg has to be run separately for each architecture, I think we should either a) install it for all architectures in this prereqs script   or b) run vcpkg in the build script.

generator: As previously commented, let's get rid of Ninja and just use the default.

ccache: Just always install it when it's available (i.e. not on Windows).

anonymous-akorn:
General question: do these scripts need to be runnable with Python 2? There are a few simplifications that can be made if we only care about Python 3.

scripts/install_prereqs.py
```diff
@@ -0,0 +1,453 @@
...
+
+    Raises:
+        ValueError: If the current OS could not be identified.
+
+    Returns:
+        (str): Package manager for current platform.
+    """
+    if is_windows_os():
+        if not is_tool_installed(PackageManager.WINDOWS):
+            raise ValueError('Did not find {0} on the system. \
```
anonymous-akorn, line 237: This will capture the big block of whitespace in front of "Please install" and "try setting" and produce something like this:

```
"Did not find hi on the system.                                 Please install it or                                 try setting install_if_not_found=True"
```

Python will concatenate consecutive strings automatically, so you can write something like this:

```
raise ValueError('a'
                 ' b'
                 ' c'.format(stuff))
```

I've seen this kind of thing in a couple of places, and concatenation (implicit, as above, or explicit, with +) is strongly recommended over using \

scripts/install_prereqs.py
```diff
@@ -0,0 +1,453 @@
...
+
+def install_system_packages(package_names):
+    """ Install packages using system package manager.
+    brew for mac, apt for linux, choco for windows
+
+    Args:
+        package_names (:obj:`list` of :obj:`str`): List of packages to install.
+                                                    Eg: ['ninja', 'ccache'].
+    Raises:
+        (ValueError): When either package manager is not found or packages list is empty.
```
anonymous-akorn, line 268: Looks like it prints and returns if it's empty.

scripts/install_prereqs.py
```diff
@@ -0,0 +1,453 @@
...
+    cmd = [package_manager, 'install'] + package_names
+    if is_linux_os():
+        run_command(cmd, as_root_user=True)
+    else:
+        run_command(cmd)
+
+
+def install_python_packages():
+    """ Install python packages needed as dependencies for cpp sdk build.
+    """
```
anonymous-akorn, line 288: Nit: For single line docstrings, the terminating """ should be on the same line.

scripts/install_prereqs.py
```diff
@@ -0,0 +1,453 @@
...
+    if args.generator == 'Ninja':
+        if is_tool_installed('ninja'):
+            print('Ninja requested but is already installed. Skipping installation...')
+        elif is_linux_os():
+            system_packages_to_install.append('ninja-build')
+        elif is_mac_os:
+            system_packages_to_install.append('ninja')   
+
+    # ccache is supported only on mac and linux and check if it is already installed
+    if args.ccache and (is_linux_os() | is_mac_os()) and not is_tool_installed('ccache'):
```
anonymous-akorn, line 375: A couple of things here. First, I'm assuming the '|' is supposed to be 'or'?

Second, I might be misreading this since it's late on a Friday, but let's say
a = args.ccache
b = is_linux_os() or is_mac_os()
c = is_tool_installed('ccache')

Then the outer conditional is "if a and b and not c"

The first inner conditional is "if c", which is always false at this point.

The second inner conditional is "if b", which is always true at this point.

scripts/install_prereqs.py
```diff
@@ -0,0 +1,453 @@
...
+    installed_packages = {pkg.key for pkg in pkg_resources.working_set}
+
+    # Get the list of packages we intend to install
+    requirements_file_path = get_pip_requirements_file_path()
+    packages_to_install = None
+    with open(requirements_file_path, 'r') as requirements_file:
+        packages_to_install = {line.strip('\n') for line in requirements_file.readlines()}
+
+    # Check if there is any work to do by comparing these lists
+    if packages_to_install and not (packages_to_install-installed_packages):
```
anonymous-akorn, line 401: Nit: should be a space around the operator.

scripts/install_prereqs.py
```diff
@@ -0,0 +1,453 @@
...
+
+    Args:
+        args (`argparse.Namespace`): Parsed command line arguments.
+    """
+    # Check for packages installed in this python environment
+    installed_packages = {pkg.key for pkg in pkg_resources.working_set}
+
+    # Get the list of packages we intend to install
+    requirements_file_path = get_pip_requirements_file_path()
+    packages_to_install = None
```
anonymous-akorn, line 396: This initialization is unnecessary, as scope in Python is a bit more permissive. You can initialize a value inside of an intended block (such as 'with', 'if', etc) and reference it outside of the block.

scripts/install_prereqs.py
```diff
@@ -0,0 +1,453 @@
...
+    with open(requirements_file_path, 'r') as requirements_file:
+        packages_to_install = {line.strip('\n') for line in requirements_file.readlines()}
+
+    # Check if there is any work to do by comparing these lists
```
anonymous-akorn, line 400: Is there any harm in running pip install on the requirements file if all the stuff has already been installed? Pip will find that the requirement is already satisfied and not do any real work. Seems to me that we could remove this code to compare the requirements file with what's already installed.

scripts/build.py
```diff
@@ -0,0 +1,217 @@
...
+            output = subprocess.check_output(cmd, cwd=cwd)
+            return output.decode('utf-8').strip()
+        else:
+            subprocess.call(cmd, cwd=cwd)
+    except subprocess.CalledProcessError:
+        raise IOError('Error executing {0}'.format(cmd_string))
+
+
+def get_repo_root_dir():
+    """ Get root directory of current git repo.
```
anonymous-akorn, line 73: Nit: no space after the """

scripts/build.py
```diff
@@ -0,0 +1,217 @@
...
+    Raises:
+        IOError: Current directory is not in a git repo.
+
+    Returns:
+        (str): Root directory of git repo.
+    """
+    return run_command(['git', 'rev-parse', '--show-toplevel'], collect_output=True)
+
+
+REPO_ROOT_DIR = get_repo_root_dir()
```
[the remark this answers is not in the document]
vimanyu: Yeah, I wanted to cache the value as its used in multiple places. (wanted to use a `lru_cache` from Python3 but that broke Python2 and writing a cache function just for this would have made it look complicated).
Re: why use 'git rev-parse', I thought it will be nice to allow execution of this python script from any subdirectory in the repo and it just works. I could assume users are always at the root but when I am debugging, I am sometimes in the directory containing python script and it will be convenient if I could just run it.

scripts/build.py
```diff
@@ -0,0 +1,217 @@
...
+
+def is_windows_os():
+    """Check if the current OS is Windows.
+
+    Returns:
+        (bool): True if windows, False otherwise.
+    """
+    return sys.platform.startswith('win')
```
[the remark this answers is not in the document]
vimanyu: This is good to know! Much nicer return values.

scripts/build.py
```diff
@@ -0,0 +1,217 @@
...
+import sys
+import multiprocessing
+import argparse
+
+
+def run_command(cmd, cwd=None, as_root_user=False, collect_output=False):
+    """Run a command.
+
+    Args:
+        cmd (:obj:`list` of :obj:`str`): Command to run as a list object.
```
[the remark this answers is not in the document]
vimanyu: TBH, this is the first time I used this notation too. This was based off this one article I found while looking for google docstrings example
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html
I have always used "list(str)" or "iterable(str)" before.

scripts/install_prereqs.py
vimanyu: No harm but I was saving microseconds here by not incurring the cost of running pip. Makes it easier to read if I take this optimization out though.

scripts/install_prereqs.py
vimanyu: Good to know!

vimanyu:
I went with PEP-8 indentation standards. 
https://www.python.org/dev/peps/pep-0008/#indentation

Can move it to 2 spaces if we want to stick to google style guide though.

scripts/install_prereqs.py
vimanyu: My bad. I folded the nested conditions into one condition and didnt remove the old blocks. Thanks for catching this!

vimanyu:
Created a new branch more in line with design requirements. Separate PR for that. Closing this one.

PR #102, opened 2020-07-25
Feature/prereqs build scripts redesign

vimanyu:
Redesigned the prereq and build scripts according to this principle,
- prereqs should be platform aganostic and run once
- all platform dependent steps should be done in build (vcpkg now happens here)

Both these scripts are using Python3 goodness now.


scripts/build.py
```diff
@@ -0,0 +1,243 @@
...
+
+# Build only firebase_app and firebase_auth with default options and also build unit tests
+python build.py --build_tests --arch x64 --targets firebase_app firebase_auth
+
+"""
+
+import subprocess
+import os
+import sys
+import argparse
```
jonsimantov, line 40: Since we're installing absl-py anyway and we're trying to stick to the google python style guide wherever possible, WDYT about using absl.flags instead of argparse?

scripts/install_prereqs.py
```diff
@@ -0,0 +1,98 @@
...
+        cmd = ['apt', 'install', 'ccache']
+        run_command(cmd, as_root_user=True)
+
+    elif platform.system() == 'Darwin':
+        cmd = ['brew', 'install', 'ccache']
+        run_command(cmd)
+
+  # Install required python dependencies
+  python3_cmd = 'python3' if is_command_installed('python3') else 'python'
+  cmd = [python3_cmd, '-m', 'pip', 'install', '-r', 'external/pip_requirements.txt', '--user']
```
jonsimantov, line 93: nit, can just be in one line

scripts/build.py
```diff
@@ -0,0 +1,243 @@
...
+    - initialize/update the vcpkg git submodule
+    - build vcpkg executable
+    - install packages via vcpkg.
+  Args:
+    arch (str): Architecture (eg: 'x86', 'x64').
+  """
+  # vcpkg is a submodule of the main cpp repo
+  # Ensure that the submodule is initialized and updated
+  run_command(['git', 'submodule', 'init'])
+  run_command(['git', 'submodule', 'update'])
```
jonsimantov, line 148: I'd be fine seeing this in main() since ultimately we may end up with other submodules.

scripts/install_prereqs.py
```diff
@@ -0,0 +1,98 @@
...
+def download_file(url, file_path):
+  """Download from url and save to specified file path."""
+  with urllib.request.urlopen(url) as response, open(file_path, 'wb') as out_file:
+    shutil.copyfileobj(response, out_file)
+
+
+def unpack_files(archive_file_path, output_dir=None):
+  """Unpack/extract an archive to specified output_directory"""
+  shutil.unpack_archive(archive_file_path, output_dir)
+
```
jonsimantov, line 78: I still think these should all be in a module that is imported, to keep this script clean.

In Python 3 I don't think it'll be that difficult:  https://stackoverflow.com/questions/4142151/how-to-import-the-class-within-the-same-directory-or-sub-directory

I'd like to do that in build.py as well, for the cmake_* functions, run_command, is_{platform}, and even the vcpkg path helper functions (not the main vcpkg function though).

scripts/install_prereqs.py
```diff
@@ -0,0 +1,98 @@
...
+
+def main():
+  # Install ccache on linux/mac
+  if not is_command_installed('ccache'):
+    if platform.system() == 'Linux':
+        cmd = ['apt', 'install', 'ccache']
+        run_command(cmd, as_root_user=True)
```
jonsimantov, line 85: Again, I'd be fine seeing all these run_command calls just looking like run_command(['apt', 'install', 'ccache'], as_root=True);

(also you may want to add command-line arguments to ensure apt doesn't prompt for anything, -y or -yy or whatever it is.)

scripts/build.py
```diff
@@ -0,0 +1,243 @@
...
+"""
+
+import subprocess
+import os
+import sys
+import argparse
+import platform
+
+
+def run_command(cmd, cwd=None, as_root_user=False, collect_output=False):
```
jonsimantov, line 44: as_root is shorter and pithier ;)

scripts/build.py
```diff
@@ -0,0 +1,243 @@
...
+ """Run a command.
+
+ Args:
+  cmd (list(str)): Command to run as a list object.
+          Eg: ['ls', '-l'].
+  cwd (str): Directory to execute the command from.
+  as_root_user (bool): Run command as root user with admin priveleges (supported on mac and linux).
+  collect_output (bool): Get output from the command. The string returned from this function
```
jonsimantov, line 52: I don't really love the return value EITHER being the return code of the subprocess OR the collected output. What about always returning the error code and the output in a tuple? (return_code, output_string)

scripts/build.py
```diff
@@ -0,0 +1,243 @@
...
+
+def main():
+  args = parse_cmdline_args()
+  
+  # Install platform dependent cpp dependencies with vcpkg 
+  install_cpp_dependencies_with_vcpkg(args.arch)
+  # CMake configure
+  cmake_configure(args.build_dir, args.arch, args.build_tests, args.config)
+  # CMake build
+  cmake_build(args.build_dir, args.target)
```
jonsimantov, line 228: Actually now that I'm thinking about it, why have functions for cmake_* at all?

Couldn't the second one just be run_command(['cmake', '--build', build_dir, '-j', str(os.cpu_count()] + ['--target', args.target] if args.target else [])

It's like 5x larger than it needs to be because of docstrings, etc.

scripts/build.py
```diff
@@ -0,0 +1,243 @@
...
+
+ Args:
+  cmd (list(str)): Command to run as a list object.
+          Eg: ['ls', '-l'].
+  cwd (str): Directory to execute the command from.
+  as_root_user (bool): Run command as root user with admin priveleges (supported on mac and linux).
+  collect_output (bool): Get output from the command. The string returned from this function
+       call will have the output from command execution.
+ Raises:
+  (subprocess.CalledProcessError): If command errored out.
```
jonsimantov, line 55: There should be an option to ignore errors, possibly useful in prereqs. (though if we move this function to a shared module, we can add functionality to it as we need it).

scripts/build.py
```diff
@@ -0,0 +1,243 @@
...
+ Returns:
+  (str|None): If collect_output is provided, a string is returned
+     None if output is not requested.
+ """
+ if as_root_user and (is_mac_os() or is_linux_os()):
+  cmd.insert(0, 'sudo')
```
jonsimantov, line 61: What happens if sudo prompts for a password? (as it might on a desktop, just not on GHA runners).

scripts/build.py
anonymous-akorn: If were assuming access to absl-py, that also gives us access to absl's logging module. Instead of print, we can use logging.info, logging.warning and logging.error which will label the message with a timestamp, line number, and log level (I, W, E). Having that extra info can be quite helpful in a number of situations.

For consistency, we probably should decide whether tools in this repo should make use of absl-py or not.

scripts/build.py
```diff
@@ -0,0 +1,243 @@
...
+   targets (list(str)): Specific build targets.
+             If not specified, cmake will build all targets.
+  """   
+  cmd = ['cmake', '--build', build_dir, '-j', str(os.cpu_count())]
+  if targets:
+    targets_cmd_line_arg = ['--target']
+    targets_cmd_line_arg.extend(args.target)
+    cmd.extend(targets)
+
+  run_command(cmd)
```
anonymous-akorn, line 217: There are a couple of issues with this method.

1. cpu_cores is no longer an arg.
2. 'args.target' is referenced, but args does not exist here.
3. 'targets_cmd_line_arg' is defined but not used. 

I'm guessing targets_cmd_line_arg is meant to be extended with 'targets' instead of args.target, and then cmd extended with targets_cmd_line_arg? Though you could skip the intermediate list and just do something like this:

cmd.append('--target')
cmd.extend(targets)

scripts/build.py
```diff
@@ -0,0 +1,243 @@
...
+  
+  # If generator is not specifed, default for platform is used by cmake, else 
+  # use the specified value
+  if config:
+    cmd.append('-DCMAKE_BUILD_TYPE={0}'.format(config))
+  if build_tests:
+    cmd.append('-DFIREBASE_CPP_BUILD_TESTS=ON')
+    cmd.append('-DFIREBASE_FORCE_FAKE_SECURE_STORAGE=ON')
+
+  vcpkg_toolchain_file_path = os.path.join(os.getcwd(), 'external', 'vcpkg', 'scripts', 'buildsystems', 'vcpkg.cmake')
```
anonymous-akorn, line 191: Do we care about sticking to the 80 character line width limit prescribed by the style guide? There are quite a few lines that exceed it. 

Though I'm not going to lie, having longer lines would be awfully convenient.

scripts/build.py
```diff
@@ -0,0 +1,243 @@
...
+ if as_root_user and (is_mac_os() or is_linux_os()):
+  cmd.insert(0, 'sudo')
+
+ cmd_string = ' '.join(cmd)
+ print('Running cmd: {0}\n'.format(cmd_string))
+ if collect_output:
+  output = subprocess.check_output(cmd, cwd=cwd)
+  return output.decode('utf-8').strip()
+ else:
+  subprocess.call(cmd, cwd=cwd)
```
anonymous-akorn, line 69: Your comment mentions using "Python 3" goodness but is still written in a Python 2 friendly way.

If we're fine with requiring Python 3, there is a unified subprocess method called subprocess.run, which has args to reproduce the various behaviours of old subprocess methods. check, capture_output, etc. You can also specify the text or encoding arg instead of decoding the output explicitly.

https://docs.python.org/3/library/subprocess.html#using-the-subprocess-module

scripts/install_prereqs.py
anonymous-akorn: Yeah, shoving functionality into library files and then importing them should be a non-issue, and shouldn't impose any additional burden on those running the scripts.

scripts/build.py
jonsimantov: Yeah let's use that, the GHA runners have Python 3, and I'm okay with listing it as a dependency if users want to run the scripts manually.

scripts/build.py
jonsimantov: I hate to say it, but let's try to stick with the style guide. At least the [public one](https://google.github.io/styleguide/pyguide.html). It has a link to a python formatter (yapf) that we should use.

We should probably put auto code formatting hooks into the git repo.

scripts/build.py
vimanyu: I am on the fence about this. For simple scripts like this, we should try to keep them as lean as possible. Python has standard library modules for command line argument parsing and logging (not the best but is workable and can set formatting and log levels). Also, since we are trying to create something very close to a shell script, we probably wont use logging much because it adds verbosity. Google python style guide doesn't enforce its usage and taking a quick glance at firebase-ios-sdk, they are using standard logging and arparse modules in their python scripts too.

scripts/install_prereqs.py
vimanyu: Will be bundling them under `scripts/build/*.py`.

scripts/install_prereqs.py
vimanyu: Good point. Should test this out.

scripts/build.py
vimanyu: Definitely agree with this. I kind of don't like returning tuples because on the consumer end, we have to extra careful about the return values and unpacking these tuples. Maybe create 2 separate functions in our utility module?

scripts/build.py
vimanyu: If we are maintaining parity with a shell script, we probably shouldn't ignore the error and stop when a command fails. Also, ignoring on the consumer end could be easily done with a try-except block .

scripts/build.py
vimanyu: My guess is that our subprocess call is blocking and will wait for user input and the command to finish. So we should be fine but I will still confirm this.

scripts/build.py
vimanyu: Thanks for catching this. Yes, I changed the signature to remove `args` as an argument and missed this.

scripts/build.py
vimanyu: Good to know. Yeah, aside from this, I used some other Python3 "only" constructs,
- downloading urls
- unpacking files
- os.cpu_count instead of multiprocess.cpu_count

So yeah, Python3 is definitely required for these scripts.

scripts/build.py
anonymous-akorn: I don't think absl.flags vs argparse makes a huge difference, but I think there is value in picking one and using that for all of our scripts to minimize the difficulty in contributing/maintaining the repo. (Mind you I'm guilty of violating this - my PR has absl.flags for almost everything, but then argparse for the restore_secrets script. I plan to rewrite one of them once we decide on which way to go with this).  

I'm not sure I follow the shell script argument though. Adding a bit of extra verbosity for useful diagnostic information seems like a good tradeoff. If nothing else, it seems very valuable to know how long each step is taking, which the timestamps give us. 

I don't think it's a huge deal either way, but I think it makes sense to use absl-py, for both flags and logging.

scripts/build.py
anonymous-akorn: Python at least has nice syntax for unpacking tuples

x, y = func_that_returns_tuple()

Though in this case, it looks like you're not returning the return code? The return value is the output, which is possibly none (based on an arg). Returning the return code is unnecessary, since if it's anything other than 0, a CalledProcessError will occur. Thus if the function completes, it's safe to assume a return code of 0.

vimanyu: Actually `subprocess.run` that I learnt from you is very convenient. "check" and "capture_output" are exactly the things we need. It also returns an object which contains the output of the command in an attribute called ".stdout". 
So yeah, this function is now just a wrapper over that.

scripts/build.py
jonsimantov: Generally we run shell scripts with -e so they will exit on any error. We can tweak this later though.

vimanyu: cool. Python 3.7 has a convenient way to deal with this (check=True argument to subprocess.run will raise an exception, else it wont). Apologies but since I restructured the code, it will show up as a brand new file.

.github/workflows/desktop.yml
```diff
@@ -50,44 +39,7 @@ jobs:
         run: |
           echo "::set-env name=VCPKG_RESPONSE_FILE::external/vcpkg_${{ matrix.vcpkg_triplet }}_response_file.txt"
           echo "::set-env name=PIP_REQUIREMENTS_FILE::external/pip_requirements.txt"
```
DellaBitta, line 41: is PIP_REQUIREMENTS_FILE still required?

vimanyu: Nope, it wasn't. Thanks for catching it. Have removed it now.

scripts/gha/build_desktop.py
```diff
@@ -0,0 +1,145 @@
...
+
+# Build all targets with default options and also build unit tests
+python scripts/gha/build_desktop.py --build_tests --arch x64
+
+# Build only firebase_app and firebase_auth 
+python scripts/gha/build_desktop.py --target firebase_app firebase_auth
+
+"""
+
+import argparse
```
jonsimantov, line 37: I think we should land on absl.flags for all arg parsing in all our scripts. If you want to do this in a follow-up PR that's fine.

scripts/gha/build_desktop.py
```diff
@@ -0,0 +1,145 @@
...
+  cmake_configure(args.build_dir, args.arch, args.build_tests, args.config)
+
+  # CMake build 
+  # cmake --build build -j 8 
+  cmd = ['cmake', '--build', args.build_dir, '-j', str(os.cpu_count())] 
+  if args.target:
+    # Example:  cmake --build build -j 8 --target firebase_app firebase_auth
+    cmd.append('--target')
+    cmd.extend(args.target)
+  utils.run_command(cmd)
```
jonsimantov, line 130: not cmd = ['cmake', '--build', args.build_dir, '-j', str(os.cpu_count())] + ['--target', args.target] if args.target else []?

Maybe I'm biased as a former Perl programmer ;)

vimanyu: So we will have to put parantheses around the second part and that was making it harder to read. 
```
l = ['a', 'b'] + ['c', 'd'] if something else []
# something = True
# l = ['a', 'b', 'c', 'd']
# something = False
# l = []

l = ['a', 'b'] + (['c', 'd'] if something else [])
```

scripts/build.py
vimanyu: Will be creating another PR incase we decide to go absl.

scripts/gha/build_desktop.py
vimanyu: Yup! Will do it in a follow up PR.

PR #103, opened 2020-07-27
Add tools for C++ integration tests

anonymous-akorn:
Adds the tools needed to build and run C++ integration tests.

These derive from the corresponding tools in the internal repo, with some changes and exceptions:
- build_testapps.py and build_desktop_testapps.py have been merged into a single, simpler build_testapps.py.
- restore_secrets.py is a new script that decrypts the secret files in scripts/gha-encrypted and restores all the files/values that were stripped out when pushing the integration source code to Github. Except for the reverse ids in the Info.plists needed for iOS, which are not present yet.
- xcodebuild.py, provisioning.py and xcode_tool.rb are all for iOS, which is not yet supported until signing is figured out.

This also includes a change to .gitignore to ignore temporary Python files, as well as unencrypted secrets.

There are still a few things I think I could improved, as we take this stuff out of the internal repo. 
- The main reason for the JSON file (as opposed to putting that data in Python) was to allow configuration changes without needing to get Python readability. That's no longer the case, so we could just do it all in Python and get rid of the config_reader nonsense (which isn't implemented as well as I'd like anyway).
- The reason for refactoring the logic for xcodebuild, provisioning, etc. out into separate modules was to share it between both the Unity and C++ projects. It's unclear how/whether we'll do this for the github repos, or whether we really need to.

I'm open to making other large scale changes as well. 

Another question is where we should be putting this stuff. testing/integration_testing might not be the best place, especially for restore_secrets.py, which should probably go into scripts/

### DIFF
--- a/testing/integration_testing/config_reader.py
+++ b/testing/integration_testing/config_reader.py
@@ -1,0 +1,183 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""A utility for working with testapp builder JSON files.
+
+This module handles loading the central configuration file for a testapp
+builder, as well as offering simplified access to the loaded in-memory
+object.
+
+The main motivation for this module is to provide descriptive error messages,
+rather than an unhelpful KeyError that one would get from working with the
+loaded JSON (a dictionary) directly. Instead, errors will dump a readable form
+of the section of JSON being read, alongside the error.
+
+Example of such a configuration file:
+
+{
+  "apis": [
+    {
+      "name": "analytics",
+      "full_name": "FirebaseAnalytics",
+      "bundle_id": "com.google.ios.analytics.testapp",
+      "ios_target": "integration_test",
+      "testapp_path": "firebase/analytics/client/cpp/integration_test",
+      "frameworks": [
+        "firebase_analytics.framework",
+        "firebase.framework"
+      ],
+      "provision": "Google_Development.mobileprovision"
+    },
+    {
+      "name": "admob",
+      "full_name": "FirebaseAdmob",
+      "bundle_id": "com.google.ios.admob.testapp",
+      "ios_target": "integration_test",
+      "testapp_path": "firebase/admob/client/cpp/integration_test",
+      "frameworks": [
+        "firebase_admob.framework",
+        "firebase.framework"
+      ],
+      "provision": "Google_Development.mobileprovision"
+    }
+  ],
+  "dev_team": "ABCDEFGHIJK"
+}
+
+Available methods:
+
+    read_config reads the config into memory.
+    read_general_config provides access to the root level config
+    read_api_config provides access to an API's config, and requires its name.
+        This must occur in a root level "apis" list of objects.
+    get_api_names returns a sequence of the names of configured APIs, for use
+        with read_api_config.
+
+"""
+
+import json
+import os
+import pathlib
+
+# List of api-specific configurations.
+_APIS_KEY = "apis"
+
+_NAME_KEY = "name"
+
+# These are used to contextualize where a value is being read within the config
+# to provide better error messages.
+_ROOT_CONFIG = "root config "
+_API_CONFIG = "config in 'apis' list"
+
+
+def read_config(path=None):
+  """Creates an in-memory config object out of a testapp config file.
+
+  Args:
+    path: Path to a testapp builder config file. If not specified, will look
+        for 'build_testapps.json' in the same directory as this file.
+
+  Returns:
+    An in-memory config object that can be used with the other methods in
+    this module to extract data from the config passed in.
+
+  """
+  if not path:
+    directory = pathlib.Path(__file__).parent.absolute()
+    path = os.path.join(directory, "build_testapps.json")
+  with open(path, "r") as config:
+    return json.load(config)
+
+
+def read_general_config(config, key, optional=False):
+  """Reads a configuration value not tied to a particular API.
+
+  Args:
+    config: Configuration created by read_config.
+    key: Key whose value is being read.
+    optional: Return a None if the key is not found.
+
+  Returns:
+    Value associated to the given key. Could be any type supported by
+    Python's JSON module.
+
+  """
+  return _get_from_json(config, key, _ROOT_CONFIG, optional)
+
+
+def read_api_config(config, api_name, key, optional=False):
+  """Reads a configuration value for a particular API.
+
+  Args:
+    config: Configuration created by read_config.
+    api_name: Name of the API whose configuration is being read.
+    key: Key whose value is being read.
+    optional: Return a None if the key is not found.
+
+  Returns:
+    Value associated to the given key. Could be any type supported by
+    Python's JSON module.
+
+  """
+  api_dicts = _get_from_json(config, _APIS_KEY, _ROOT_CONFIG)
+  api_dict = [api for api in api_dicts if api[_NAME_KEY] == api_name][0]
+  return _get_from_json(api_dict, key, _API_CONFIG, optional)
+
+
+def get_api_names(config):
+  """Returns a list of all the APIs present in this config."""
+  apis = _get_from_json(config, _APIS_KEY, _ROOT_CONFIG)
+  return [_get_from_json(api, _NAME_KEY, _API_CONFIG) for api in apis]
+
+
+def _get_from_json(json_dict, key, json_description, optional=False):
+  """Attempts to retrieve key from the json dictionary.
+
+  If the key is not found and is not marked as optional, an error
+  will be raised with detailed information including the key, a
+  pretty-printed dump of the given json, and a description of the json
+  for context.
+
+  Args:
+    json_dict: A dictionary representation of a JSON object.
+    key: The key whose value is being extracted from the JSON dictionary.
+    json_description: A description of the json object to clarify where
+        the object came from. Will be included in an error message, if the key
+        is not found in the dict. Especially important for nested objects.
+        e.g. "API object from 'APIS' list in main config".
+    optional: Return a None if the key is not found.
+
+  Raises:
+    ValueError: Key is None or empty, json_dict is not a dictionary.
+    RuntimeError: Non-optional Key not found in the dictionary.
+
+  Returns:
+    The value of the key in the dictionary, equivalent to json_dict[key].
+
+  """
+  if not key:
+    raise ValueError("Must supply a valid, non-empty key.")
+  try:
+    return json_dict[key]
+  except TypeError:  # json_dict is not a dict
+    raise ValueError(
+        "Expected dictionary (JSON object) from %s, received %s: %s instead." %
+        (json_description, type(json_dict), json_dict))
+  except KeyError:  # key not found
+    if optional:
+      return None
+    formatted_json = json.dumps(
+        json_dict, sort_keys=True, indent=4, separators=(",", ":"))
+    raise RuntimeError(
+        "%s not found in %s:\n%s" % (key, json_description, formatted_json))
--- a/testing/integration_testing/config_reader.py
+++ b/testing/integration_testing/config_reader.py
@@ -72,7 +72,7 @@ def read_config(path=None):
         look for 'build_testapps.json' in the same directory as this file.
 
   Returns:
-    A 'Config' containing all the testapp builder's configuration.
+    Config: All of the testapp builder's configuration.
 
   """
   if not path:
@@ -100,7 +100,7 @@ def read_config(path=None):
         compilers=config["compiler_dict"])
   except (KeyError, TypeError, IndexError):
     # The error will be cryptic on its own, so we dump the JSON to
-    # offer context, then reraise it the error.
+    # offer context, then reraise the error.
     print(
         "Error occurred while parsing config. Full config dump:\n"
         + json.dumps(config, sort_keys=True, indent=4, separators=(",", ":")))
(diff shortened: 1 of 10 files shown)


testing/integration_testing/restore_secrets.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+def _decrypt(encrypted_file, passphrase):
+  """Generates a decrypted file with same path minus the '.gpg' extension."""
+  print("Decrypting %s" % encrypted_file)
+  # Note: if setting check=True, be sure to catch the error and not rethrow it
+  # or print a traceback, as the message will include the passphrase.
+  result = subprocess.run(
+      args=[
+          "gpg",
```
vimanyu, line 107: Should we install gpg in`install_prereqs` scripts? Wondering if it makes sense to just install the python gpg library instead of invoking system package manager to install this?

testing/integration_testing/requirements.txt
```diff
@@ -0,0 +1,2 @@
+absl-py
```
vimanyu, line 1: Should we name this file `pip_requirements.txt` to be clearer on what it is?

jonsimantov:
The reversed client ID can be extracted from GoogleService-Info.plist and google-services.json, if you want to just do that.

testing/integration_testing/requirements.txt
anonymous-akorn: From what I can tell, it's a soft convention in the Python community to call it requirements.txt.

testing/integration_testing/restore_secrets.py
anonymous-akorn: We should install gpg somewhere, assuming it's not already available on GHA machines. Though it's worth noting that we only want to install it for ourselves, and on CI, not for external contributors or users, since it's used for decrypting secrets that only we should have access to.

Looking at the Python module, it appears to be a wrapper for the gpg tool itself?
"Apart from a recent-enough version of Python, in order to use this module you need to have access to a compatible version of the GnuPG executable."

In which case we would still need it installed.
https://pythonhosted.org/python-gnupg/

testing/integration_testing/xcodebuild.py
```diff
@@ -0,0 +1,144 @@
...
+      "xcodebuild",
+      "-sdk", _get_ios_env_from_target(ios_sdk),
+      "-scheme", scheme,
+      "-configuration", configuration,
+      "-quiet",
+      "DEVELOPMENT_TEAM=" + dev_team,
+      "BUILD_DIR=" + output_dir
+  ]
+
+  if path is None:
```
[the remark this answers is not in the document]
anonymous-akorn: It'll get caught by the final else check regardless, but "if not path" is better anyway, as the error message (path not supplied) will be more specific than the one given by the else condition (lacking an appropriate extension).

testing/integration_testing/config_reader.py
```diff
@@ -0,0 +1,183 @@
...
+    key: Key whose value is being read.
+    optional: Return a None if the key is not found.
+
+  Returns:
+    Value associated to the given key. Could be any type supported by
+    Python's JSON module.
+
+  """
+  api_dicts = _get_from_json(config, _APIS_KEY, _ROOT_CONFIG)
+  api_dict = [api for api in api_dicts if api[_NAME_KEY] == api_name][0]
```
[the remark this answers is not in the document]
anonymous-akorn: This module is actually annoying me for a couple of reasons, so I just made it read the entire config into a simple data class, handling the validation upfront: eliminated a lot of this weird looking stuff.

testing/integration_testing/config_reader.py
```diff
@@ -0,0 +1,183 @@
...
+# to provide better error messages.
+_ROOT_CONFIG = "root config "
+_API_CONFIG = "config in 'apis' list"
+
+
+def read_config(path=None):
+  """Creates an in-memory config object out of a testapp config file.
+
+  Args:
+    path: Path to a testapp builder config file. If not specified, will look
```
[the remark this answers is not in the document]
anonymous-akorn: The style guide does in fact prescribe this, so I'll add types in the docstrings.

I do find it weird to specify types in a dynamic language though. Paths can be strings, Path objects, for a few other things. 'File' objects can be a dozen different types. etc.

anonymous-akorn:
> The reversed client ID can be extracted from GoogleService-Info.plist and google-services.json, if you want to just do that.

Yeah that's a good call, simple to do and results in one fewer thing to maintain.

testing/integration_testing/restore_secrets.py
anonymous-akorn: Looked into it, and GHA documentation states that GPG is installed on MacOS and Linux, but it's not listed for Windows.

Though I tested it out, and found GPG installed and on the path for all three platforms.

testing/integration_testing/requirements.txt
vimanyu: Sure, we can leave it as it is. In my case, there are a bunch of these files for vcpkg too and am using this naming scheme it to distinguish between them. (vcpkg_* vs pip_*).

testing/integration_testing/provisioning.py
```diff
@@ -76,7 +76,7 @@ def get_provision_id(provision_path):
   Can only be run on macOS.
 
   Args:
-    provision_path: Path to a .mobileprovision file.
+    provision_path (str): Path to a .mobileprovision file.
 
   Returns:
     A string corresponding to a UUID.
```
vimanyu, line 82: As per google docstrings, returns should have a type too. 
```
Returns:
    str: UUID
```

testing/integration_testing/provisioning.py
```diff
@@ -0,0 +1,141 @@
...
+This should only be run on MacOS, as it uses a mac-specific
+application.
+
+"""
+
+import plistlib
+import subprocess
+
+
+def patch_provisioning_profile(plist_path, bundle_id, uuid):
```
jonsimantov, line 34: Might be better to have this take a named list of parameters in case we add more patches in the future?

testing/integration_testing/provisioning.py
```diff
@@ -0,0 +1,141 @@
...
+    raise ValueError("Path is not a .mobileprovision file: %s" % provision_path)
+
+  plist = _extract_plist(provision_path)
+  uuid = _extract_uuid(plist)
+  if not uuid:
+    raise RuntimeError("No UUID found in provision %s." % provision_path)
+  return uuid
+
+
+def _extract_plist(provision_path):
```
jonsimantov, line 102: Just a note that a lot of this mobileprovision-related stuff won't be usable on GHA as-is. It might be worth stripping all these out until we have a solution.

testing/integration_testing/restore_secrets.py
```diff
@@ -0,0 +1,161 @@
...
+"""Script for restoring secrets into the integration test projects.
+
+Usage:
+
+python restore_secrets.py --passphrase <passphrase> --repo_dir <path_to_repo>
+
+repo_dir refers to the C++ SDK github repository. Defaults to current directory.
+
+As an alternative to passing the passphrase as a flag, the password can be
+stored in a file and passed via the --passphrase_file flag.
```
jonsimantov, line 24: nit: Let's list these in a more structured way:
python restore_secrets.py --passphrase <passphrase> [--repo_dir <path>]
python restore_secrets.py --passphrase_file <path> [--repo_dir <path>]

--passphrase <passphrase>: Passphrase to decrypt the files. This option is insecure on a multi-user machine; use the --passphrase_file option instead.
--passphrase_file <path>: Specify a file to read the passphrase from (only reads the first line).
--repo_dir <path>: Path to the C++ SDK github repository. Defaults to current directory.

jonsimantov: (also you should make sure it only reads the first line of passphrase_file

testing/integration_testing/restore_secrets.py
```diff
@@ -0,0 +1,161 @@
...
+- Google Service files (plist and json) will be restored into the
+  integration_test directories.
+- The server key will be patched into the Messaging project.
+- The uri prefix will be patched into the Dynamic Links project.
+- The reverse id will be patched into all Info.plist files, using the value from
+  the decrypted Google Service plist files as the source of truth.
+
+"""
+
+import argparse
```
jonsimantov, line 37: Let's switch this to absl.flags in expectation of using that for all python scripts.

testing/integration_testing/restore_secrets.py
```diff
@@ -0,0 +1,161 @@
...
+  uri_path = os.path.join(secrets_dir, "dynamic_links", "uri_prefix.txt.gpg")
+  uri_prefix = _decrypt(uri_path, passphrase)
+  dlinks_project = os.path.join(repo_dir, "dynamic_links", "integration_test")
+  _patch_main_src(dlinks_project, "REPLACE_WITH_YOUR_URI_PREFIX", uri_prefix)
+
+  print("Attempting to patch Messaging server key.")
+  server_key_path = os.path.join(secrets_dir, "messaging", "server_key.txt.gpg")
+  server_key = _decrypt(server_key_path, passphrase)
+  messaging_project = os.path.join(repo_dir, "messaging", "integration_test")
+  _patch_main_src(messaging_project, "REPLACE_WITH_YOUR_SERVER_KEY", server_key)
```
jonsimantov, line 94: Hmm, I wonder if we should actually formalize and factor this out. Scan everything for "REPLACE_WITH_YOUR_* and replace with the string in the relevant text file, if present. WDYT? (just a thought, not a strong suggestion).

testing/integration_testing/restore_secrets.py
jonsimantov: This workflow will run install_prereqs for all platforms, yes? (once ios/android are moved to python). We could have the prereqs scripts *verify* some prereqs and error out if they aren't present. e.g. gpg. So if the runners change we know exactly why by seeing the prereqs script output.

testing/integration_testing/xcode_tool.rb
```diff
@@ -0,0 +1,188 @@
...
+  @project.save
+end
+
+def make_changes_for_auth
+  puts 'Auth testapp detected.'
+  add_entitlements
+  set_reverse_id
+  add_system_framework('UserNotifications')
+  puts 'Finished making auth-specific changes.'
+end
```
jonsimantov, line 97: I'd love to minimize the actual project-specific logic in this script and have it just be a very dumb way for the Python script to set up the project. WDYT?

testing/integration_testing/provisioning.py
anonymous-akorn: I'm inclined to go with YAGNI on this one, especially since there's a non-trivial detail of where in the plist to patch in the information: in this case, we're patching in a key-value pair to a root-level 'provisioningProfile' dict, but other data would likely need to go elsewhere.

testing/integration_testing/provisioning.py
anonymous-akorn: Right, I was somewhat ambivalent about whether to include the iOS logic (which is half the code in this PR). My main reason for including it is to provide a starting/reference point for when this is modified to handle whatever solution we come up with for iOS, especially since that person might not be me (since Mou is looking into signing, he might also wire things up for this), and also since there are quite a few changes here relative to what's in google3.

testing/integration_testing/xcode_tool.rb
anonymous-akorn: Yeah I strongly agree with that actually, to treat this more like a wrapper for the relevant bits of the xcodeproj library we need and reduce the cases where it needs to be modified. Also to get rid of the error-prone logic to infer the API from the project path, which is just asking for trouble.

testing/integration_testing/restore_secrets.py
anonymous-akorn: I think there's some merit to that idea, though the possibility of matching something it's not supposed to and patching a secret into an unexpected location (possibly a README file mentioning the placeholder, for example) makes me nervous. Or running into some other unexpected issue that I'm not currently predicting, like running into a file it doesn't have permission to read.

testing/integration_testing/xcode_tool.rb
jonsimantov: Agreed - will you make that change here or in a future PR when we get iOS working?

anonymous-akorn: I put in a few comments in this PR, but will leave the rest to a future PR - probably shouldn't hold up this week's iteration for this.

testing/integration_testing/restore_secrets.py
jonsimantov: Yeah, fair. I would like to see this more data-driven though. But let's file that in the backlog.

testing/integration_testing/provisioning.py
jonsimantov: Yeah I think that makes sense. Let the record show that it doesn't work as-is now. :)

testing/integration_testing/provisioning.py
jonsimantov: Even if we won't need other things to be patched in the provisioning profile, it's still easier to read the call if it's:
patch_provisioning_profile(plist_file, bundle_id=___, uuid=___)? There's not really a good way to differentiate it in the call otherwise. I don't feel super strongly about it so I approved regardless.

anonymous-akorn: Oh I misunderstood your suggestion. When you said named list, I thought you meant taking in a list of things to patch. 

Calling with named args does not requiring modifying the method definition, though: if a method has the signature 
def method(a, b):

Then the caller can pass in named args at their discretion. Both are valid:
method('duck', 'quack')
method(a='duck', b='quack')

jonsimantov: Can it require it though?

anonymous-akorn: If the signature specifies a value for something, then that will just serve as a default value. i.e. this:
def method(a=3)

Means the following are all valid:
method()  # a will be 3
method(4)
method(a=4)

There actually **is** a way to force named args in Python, though I think most people won't recognize the syntax when they see it, and it's not trivial to look up. Putting in an asterisk as an arg means the following args all have to be called as named args:

def method(a, *, b, c):

method(3, b=4, c=5)  is valid
method(3, 4, 5)  is invalid

The error also might confuse people:
"TypeError: method() takes 1 positional argument but 3 were given"

jonsimantov: Ah ok. Weird. I guess I'm fine to keep it as-is unless the style guide disagrees.

PR #105, opened 2020-07-29
Android GHA workflow

DellaBitta:
Add Android as a GHA CI worfklow.  
 - added android.yml to our workflows
 - fixed an analytics build issue which now news to include play-services-base as an implementation dependency.
 - there's a lint failure in a firestore dependency.  Disabling abortOnError for that gradle build target for now. I've added an internal issue to the backlog to circle back on this issue later.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -1,0 +1,79 @@
+name: Android Builds
+
+on: 
+  pull_request:
+    types: [opened, reopened, synchronize]
+
+env:
+  CCACHE_DIR: ${{ github.workspace }}/ccache_dir
+
+jobs:
+  build:
+    name: android-${{ matrix.os }}-${{ matrix.build_type }}-${{ matrix.architecture }}-${{ matrix.python_version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-latest, macos-latest]
+        architecture: ["x64",]
+        python_version: [3.7]
+        include:
+        - os: ubuntu-latest
+          architecture: "x64"
+        - os: macos-latest
+          architecture: "x64"
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: Set env variables for subsequent steps (all)
+        run: |
+          echo "::set-env name=PIP_REQUIREMENTS_FILE::external/pip_requirements.txt"
+          echo "::set-env name=MATRIX_UNIQUE_NAME::${{ matrix.os }}-${{ matrix.architecture }}-${{ matrix.python_version }}"
+      
+      - name: Install system level applications (mac)
+        if: startsWith(matrix.os, 'macos')
+        id: macos-brew-install
+        run: |
+          brew install ccache
+          echo "::set-env name=CCACHE_INSTALLED::1"
+      
+      - name: Install system level applications (linux)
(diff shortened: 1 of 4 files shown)


.github/workflows/android.yml
```diff
@@ -0,0 +1,79 @@
...
+        id: cache_ccache
+        uses: actions/cache@v2
+        with:
+          path: ccache_dir
+          key: dev-test-ccache-${{ env.MATRIX_UNIQUE_NAME }}
+
+      - name: Setup python
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python_version }}
```
jonsimantov, line 60: I think we should just remove python_version from the matrix entirely.

.github/workflows/android.yml
```diff
@@ -0,0 +1,79 @@
...
+        run: |
+          brew install ccache
+          echo "::set-env name=CCACHE_INSTALLED::1"
+      
+      - name: Install system level applications (linux)
+        if: startsWith(matrix.os, 'ubuntu')
+        id: ubuntu-apt-install
+        run: |
+          sudo apt install ccache
+          echo "::set-env name=CCACHE_INSTALLED::1"
```
jonsimantov, line 48: Could we incorporate this into the prereqs script? and run the ccache step afterwards?

.github/workflows/android.yml
```diff
@@ -0,0 +1,79 @@
...
+          architecture: "x64"
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: Set env variables for subsequent steps (all)
+        run: |
+          echo "::set-env name=PIP_REQUIREMENTS_FILE::external/pip_requirements.txt"
```
jonsimantov, line 33: Still used?

.github/workflows/android.yml
DellaBitta: Will make it an option in the script.  The hitch is that if we reuse these scripts for users to run the builds locally then we won't want to install ccache for them.

.github/workflows/android.yml
DellaBitta: Done.

.github/workflows/android.yml
jonsimantov: Yeah, that's a secondary goal. The primary goal is for these scripts to work on GHA (and be testable locally). So yeah, an option makes sense. (I'm rewriting the scripts in Python anyway so don't worry too much how it's specified here.)

.github/workflows/android.yml
DellaBitta: Removed.

.github/workflows/android.yml
DellaBitta: I'm basing it on an environment variable that we can set in GHA.

.github/workflows/android.yml
vimanyu: I think its good to have a Python version specified because of multiple reasons,
- Latest install_preqreqs and build scripts are using a feature of `subprocess.run` that is only available in Python 3.7+
- Github runners have Python 3.6 installed by default 
- Installing a fresh python environment starts us from a clean slate and is a better way of verifying our work

So yeah, we could remove it from the matrix but we should leave the "Setup Python" step in the workflow.

DellaBitta:
> Looks good, though just a heads up I may change the way it decides whether to install ccache as part of my change. :)

That's ok!

PR #107, opened 2020-07-31
Add integration test workflow

anonymous-akorn:
Most of this workflow is copied verbatim from the workflow to build the SDK. The duplication means that most changes to the SDK build workflow will need to be mirrored into this one. 

There are two new parts for the integration tests:

- Manual dispatch with input parameters that can be specified when triggering.
- Two new steps ('Prepare for integration tests' and 'Build and run integration tests').

This PR is dependent on two other PRs:
https://github.com/firebase/firebase-cpp-sdk/pull/106
https://github.com/firebase/firebase-cpp-sdk/pull/103

This workflow is not set to run automatically. For now, it will only run on manual trigger. Which APIs to build/test, which platforms to build for, and which operating systems to run on can be specified when triggering. 

Parameterizing the operating systems required a bit of hackery, as that requires modifying the job matrix. The input is supplied as a JSON list, and then fromJson converts it into valid YML. If and when Github implements a first party solution for this kind of thing, this should be easiest to migrate to that solution. Two alternatives were considered:

1. Supply normal CSV input, then run a job to convert CSV to a JSON list. Has the benefit of a cleaner triggering interface, at the expense of more boilerplate, integration complexity, and an extra job in the workflow. 
2. Hard-code the three operating systems in the job matrix, then use a job-level "if" condition to see if the matrix.os value is contained in the input (which can be a normal CSV). This does not work as the context available when the condition is evaluated does not appear to include the values in the matrix. 

Once the two dependent PRs are in, Desktop (build + test) should work. Android (build) *might* work. iOS definitely won't work for the foreseeable future.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -1,0 +1,91 @@
+name: Integration tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      platforms:
+        description: 'CSV of Desktop, Android and/or iOS'
+        default: 'Desktop'
+        required: true
+      apis:
+        description: 'CSV of apis to build and test'
+        default: 'admob,analytics,auth,database,dynamic_links,firestore,functions,instance_id,messaging,remote_config,storage'
+        required: true
+      operating_systems:
+        description: 'JSON list of VMs to run on'
+        default: '["ubuntu-latest","windows-latest","macos-latest"]'
+        required: true
+
+env:
+  CCACHE_DIR: ${{ github.workspace }}/ccache_dir
+
+jobs:
+  build:
+    name: ${{ matrix.os }}-${{ matrix.build_type }}-${{ matrix.architecture }}-${{ matrix.python_version }}
+    runs-on: ${{ matrix.os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        os: ${{ fromJson(github.event.inputs.operating_systems) }}
+        build_type: ["Debug"]
+        architecture: ["x64",]
+        python_version: [3.7]
+        include:
+        - os: windows-latest
+          architecture: "x64"
+          vcpkg_triplet: "x64-windows-static"
+          msbuild_platform: "x64"
+        - os: ubuntu-latest
+          architecture: "x64"
+          vcpkg_triplet: "x64-linux"
+        - os: macos-latest
+          architecture: "x64"
+          vcpkg_triplet: "x64-osx"
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          submodules: true
+
+      - name: Set env variables for subsequent steps (all)
+        run: |
+          echo "::set-env name=VCPKG_RESPONSE_FILE::external/vcpkg_${{ matrix.vcpkg_triplet }}_response_file.txt"
+          echo "::set-env name=MATRIX_UNIQUE_NAME::${{ matrix.os }}-${{ matrix.build_type }}-${{ matrix.architecture }}-${{ matrix.python_version }}"
+      - name: Cache vcpkg C++ dependencies
+        id: cache_vcpkg
+        uses: actions/cache@v2
+        with:
+          path: external/vcpkg/installed
+          key: dev-vcpkg-${{ matrix.vcpkg_triplet }}-${{ hashFiles(format('{0}', env.VCPKG_RESPONSE_FILE)) }}-${{ hashFiles('.git/modules/external/vcpkg/HEAD') }}
+
+      - name: Cache ccache files
+        if: startsWith(matrix.os, 'ubuntu') || startsWith(matrix.os, 'macos')
+        id: cache_ccache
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,91 @@
...
+on:
+  workflow_dispatch:
+    inputs:
+      platforms:
+        description: 'CSV of Desktop, Android and/or iOS'
+        default: 'Desktop'
+        required: true
+      apis:
+        description: 'CSV of apis to build and test'
+        default: 'admob,analytics,auth,database,dynamic_links,firestore,functions,instance_id,messaging,remote_config,storage'
```
vimanyu, line 12: I can't think of a good way but it is unfortunate we have to deal with a list of strings in 2 different ways (for apis and operating_systems). In an ideal world, CSV values is the most intuitive but looks like we cant avoid json representation. Some ideas that come to mind,
- Accept both as json strings but have the python script deal with json string input (not intuitive at all if we use python script on the command line directly).
- Accept both as json string and use `fromJson` and `join` functions to create a comma separated list and pass it to the python script (makes the workflow file unreadable but there is some consistency in passing arguments to the workflow file and the way its dealing with those)

Feel free to ignore if we are not planning on passing these inputs to the workflow manually often.

.github/workflows/integration_tests.yml
```diff
@@ -0,0 +1,91 @@
+name: Integration tests
+
+on:
+  workflow_dispatch:
+    inputs:
+      platforms:
+        description: 'CSV of Desktop, Android and/or iOS'
+        default: 'Desktop'
+        required: true
+      apis:
```
vimanyu, line 10: nit: you definitely have more context about this but I noticed its called "apis" here (and in several places in documentation) but python script flag is "--testapps".

.github/workflows/integration_tests.yml
anonymous-akorn: We can in fact make the operating systems input a normal CSV at the expense of some added complexity. 

My reason for thinking the JSON isn't a big deal is because of the default value. If you want to manually trigger it for just  windows and mac, for example, you'll hit a button to trigger the workflow, and the field for operating systems will be pre-populated with:
["ubuntu-latest","windows-latest","macos-latest"]

At that point you just delete the pieces you don't want and end up with:
["windows-latest","macos-latest"]

Having said that, I don't feel strongly about this either way, so if we feel strongly that having simple CSVs for all inputs is worth some additional workflow complexity, then I can make that change. It would look something like this:

on:
  workflow_dispatch:
    inputs:
      operating_systems:
        default: 'ubuntu-latest,windows-latest,macos-latest'
jobs:
  prepare_matrix:
    runs-on: ubuntu-latest
    outputs:
      matrix_os: ${{ steps.set-matrix-os.outputs.matrix_os }}
    steps:
    - id: set-matrix-os
      run: echo '::set-output name=matrix_os::[${{github.event.inputs.operating_systems}}]'
  test:
    needs: prepare_matrix
    name: ${{ matrix.os }}
    runs-on: ${{ matrix.os }}
    strategy:
      fail-fast: false
      matrix: 
        os: ${{fromJson(needs.prepare_matrix.outputs.matrix_os)}}

Except we'd need to write some bash in the prepare_matrix step to convert 'a,b,c' -> '["a","b","c"]'

.github/workflows/integration_tests.yml
anonymous-akorn: Yeah, 'testapp' is a carryover for historical reasons, but 'api' seems more intuitive. 

Then again, when triggering, the default value will make it immediately clear what it is, so perhaps it's better to go with that for consistency.

.github/workflows/integration_tests.yml
vimanyu: Ah ok. I feel that this is not that much harder to read and my vote would be to make the interface to the workflow consistent (CSVs). Also, "dispatch" events can be sent via REST API. These inputs will be part of the payload and will be easier if the format is the same.

.github/workflows/integration_tests.yml
DellaBitta: I personally have been confused in the past as to what the test app is.  Is it an integration test, a unit test, etc.   I know it's contextualized by the fact that the workflow is integration_test but I think prefer "api" myself.  Just my $0.02.

.github/workflows/integration_tests.yml
anonymous-akorn: I've made the change. First uses sed to replace , with "," then shoves it into ["$(value)"] with command substitution: a,b,c -> a","b",c  -> ["a",b","c"]

`OS_JSON=[\"$(echo ${{ github.event.inputs.operating_systems }} | sed 's/,/","/g')\"]`

Simplest solution in bash I could think of.

vimanyu: Nice! Thanks for doing this :).

PR #111, opened 2020-08-05
Enable Android integration test building

anonymous-akorn:
This makes the changes needed to build Android integration tests.

Different platforms are set to run as separate jobs, so that platform-specific issues don't affect other platforms, and to speed things up. This is consistent with the jobs to build SDKs.

The integration test job for Android runs the setup script for building Android SDKs, mainly to download the Android NDK.

A non-trivial consideration in the workflows is the caching. It makes sense to use the same caches for the vcpkg dependencies as the SDK build jobs, since those dependencies are the same. But this removes the ccache step for intermediate cmake objects since different cmake targets are being built. We could use separate caches for the integration tests, but we're likely already getting close to the repo limit of 5gb for the cache (each configuration for the Android SDK build job uses 500mb by itself). Since the integration tests are not running on PR, speeding these jobs up is less of a priority than the workflows running on PR. Caching this job properly might also be more complex due to the input parameters: as an example, we don't want to replace the cache for "analytics,auth,etc" with the artifacts from a job that just builds "analytics", nor do we want a cache for every combination of apis. So we'd need parameterized caching logic for each api.

This includes a small change to the restore secrets script, to log a sanitized error message when decryption fails. Previously the error message was suppressed to avoid logging the passphrase, but this makes debugging more difficult. We can just remove the passphrase from the string before logging it.

.gitignore is also changed to ignore intermediate gradle objects that appear in the repo when building for Android.


.github/workflows/integration_tests.yml
```diff
@@ -63,42 +67,38 @@ jobs:
...
       - name: Prepare for integration tests
         run: |
           pip install -r testing/integration_testing/requirements.txt
-          python testing/integration_testing/restore_secrets.py --passphrase ${{ secrets.TEST_SECRET }}
+          python testing/integration_testing/restore_secrets.py --passphrase "${{ secrets.TEST_SECRET }}"
           
       - name: Build and run integration tests
+        # The set up script for Android will download the NDK here.
+        env:
+          NDK_ROOT: '/tmp/android-ndk-r16b'
```
vimanyu, line 102: Is this workflow supported on Windows? Could this cause an issue?

anonymous-akorn: It's not supported on Windows, no. The SDK build workflow for Android is only for Linux and MacOS at the moment, so this matches that.

The build setup script downloads the NDK and puts it in this directory, so that's why I'm setting this env var accordingly.  When the SDK build workflow is updated to also work on Windows I'll update this to use whatever solution is used for that.

testing/integration_testing/build_testapps.py
```diff
@@ -444,6 +460,55 @@ def _run(args, timeout=1200):
   return subprocess.run(args=args, timeout=timeout, check=True)
 
 
+@attr.s(frozen=True, eq=False)
+class VCPKG(object):
```
vimanyu, line 464: Would be easier to read if class names were camel case. 
Reading on the client side, I initially confused this with a constant.

.github/workflows/integration_tests.yml
jonsimantov: The android prereqs script I wrote already fetches NDK-r16b in the script itself. (it's still bash using curl, but will be switched to Python asap). We should just do that, not use additional workflows - as little as possible should be in the yml files.

jonsimantov: Oh I misunderstood this, this is just using the NDK that the script has already downloaded. Never mind. But yes it will be updated to Python and support Windows soon.

jonsimantov:
> Is there a reason why the `build_testapps.py` script can't be placed in `scripts/gha`?
> Will be nice to collect all python scripts in one place and place common functions used in various build/test scripts in `scripts/gha/utils.py`.
> @jonsimantov will have more input on this but for the build scripts, we tried to stick as close as possible to resembling a shell script and avoided classes.

Agreed, we should keep all the GHA scripts in one place. I am trying to get them to read mostly like shell scripts for ease of reading and updating.

jonsimantov:
What about the option of building the tests against a binary SDK instead of against the source SDK? (for release verirfication)

anonymous-akorn:
> What about the option of building the tests against a binary SDK instead of against the source SDK? (for release verification)

Most of the integration test workflow is to reproduce the environment created to build the SDK. For testing against the binary SDK, it would be substantially simpler, as we'd just need the last two steps in the workflow ("Prepare for integration tests" and "Build and run integration tests"). 

My tentative assumption as that we'll have this (test against binary SDK) in the same workflow file as the workflow(s) responsible for prepping the SDK for release, as it will depend on those artifacts. 

> Agreed, we should keep all the GHA scripts in one place. I am trying to get them to read mostly like shell scripts for ease of reading and updating.

Moving them in with the rest of the scripts should be fine. For the library files likely to be remain specific to the testapp builder, I figure it makes sense to shove in a subdirectory, to avoid crowding up the directory.

> we tried to stick as close as possible to resembling a shell script and avoided classes.

I'm assuming the point here is to avoid the abstractions and complexity that often come with application code rather than making them literally look like shell scripts. Simple, immutable classes can be useful for readability and maintainability.

testing/integration_testing/build_testapps.py
anonymous-akorn: Yeah I wasn't sure how to deal with the acronym, but I've seen a sensible rule of thumb that says 1-2 letters gets upper case (e.g. IO), while 3+ just capitalize the first letter (e.g. Unesco).

It also occurred to me while writing this that pkg is not an acronym. 

Regardless, I'll just go with Vcpkg.

jonsimantov:
> > What about the option of building the tests against a binary SDK instead of against the source SDK? (for release verification)
> 
> Most of the integration test workflow is to reproduce the environment created to build the SDK. For testing against the binary SDK, it would be substantially simpler, as we'd just need the last two steps in the workflow ("Prepare for integration tests" and "Build and run integration tests").

I think that's fine - my point is that the workflow itself should do as little as possible and the bulk of the work should be done by the scripts.

So basically, the workflow could run in one of two ways, and the integration test script should support both cases (probably via command line arguments):

Version A:
- On each platform
  - run "install prereqs" script for this platform
  - run "prepare integration tests" for this platform
  - run "build integration test" script for this platform (implicitly builds the SDK)
  - (optional) save built integration tests as artifact
  - run "run integration test" script for this platform

Version B (binary SDK):
- On each platform
  - run "install prereqs" script for this platform
  - run "build SDK" for this platform
  - save built SDK for platform as artifact
- On Linux after all of the above
  - grab built SDKs from artifacts for all platforms
  - run "package binary SDK" script (this is a slight simplification, as the script requires one or two steps to be done on Mac on the interim)
  - save packaged binary SDK as artifact
- On each platform
  - run "install prereqs" script for this platform
  - grab binary SDK from artifacts
  - run "prepare integration tests" for this platform
  - run "build integration tests" for this platform
  - save integration test artifacts
  - run "run integration tests" script for this platform

> 
> My tentative assumption as that we'll have this (test against binary SDK) in the same workflow file as the workflow(s) responsible for prepping the SDK for release, as it will depend on those artifacts.
> 
> > Agreed, we should keep all the GHA scripts in one place. I am trying to get them to read mostly like shell scripts for ease of reading and updating.
> 
> Moving them in with the rest of the scripts should be fine. For the library files likely to be remain specific to the testapp builder, I figure it makes sense to shove in a subdirectory, to avoid crowding up the directory.
> 
> > we tried to stick as close as possible to resembling a shell script and avoided classes.
> 
> I'm assuming the point here is to avoid the abstractions and complexity that often come with application code rather than making them literally look like shell scripts. Simple, immutable classes can be useful for readability and maintainability.

Yes, correct. Also having the classes be in an imported module to make the script easier to follow.

PR #116, opened 2020-08-19
Feature/gha include firestore clean

vimanyu:
Since we decided to support just Python3 for building our SDK, firestore's nanopb dependency had to be updated to 0.3.9.6 as 0.3.9.5 has build errors in Python3. nanopb is updated across all firebase open source sdks and google3.
This commit specifically fixes issues with finding modules with Python3.
nanopb/nanopb@63f8e5e

Relevant PRs,
google/nanopb-podspec#14
firebase/firebase-ios-sdk#6214
cl/326682471

### DIFF
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -103,7 +103,7 @@ jobs:
           LANG: en_US
         run: |
           cd build
-          ctest --verbose
+          ctest --verbose -E "firestore_(objc_spec|objc_integration)_test"
 
       - name: Run unit tests (linux)
         # Linux exists as its own standalone execution step in order to invoke
@@ -116,7 +116,7 @@ jobs:
         run: |
           ulimit -c unlimited 
           cd build
-          sudo ctest --verbose
+          sudo ctest --verbose -E "firestore_(objc_spec|objc_integration)_test"
       
       - name: Prep bins for achive (linux)
         # Copies all of the binary files into one directory for ease in
--- a/scripts/gha/build_desktop.py
+++ b/scripts/gha/build_desktop.py
@@ -36,6 +36,7 @@ python scripts/gha/build_desktop.py --target firebase_app firebase_auth
 
 import argparse
 import os
+import shutil
 import utils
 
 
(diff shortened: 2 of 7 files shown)


scripts/gha/build_desktop.py
```diff
@@ -69,6 +70,17 @@ def install_cpp_dependencies_with_vcpkg(arch):
   utils.run_command([vcpkg_executable_file_path, 'install',
                      '@' + vcpkg_response_file_path, '--disable-metrics'])
 
+  vcpkg_root_dir_path = utils.get_vcpkg_root_path()
+
+  # Clear temporary directories and files created by vcpkg buildtrees
+  # could be several GBs and cause github runners to run out of space
+  buildtrees_dir_path = os.path.join(vcpkg_root_dir_path, 'buildtrees')
+  if os.path.exists(buildtrees_dir_path):
+    shutil.rmtree(buildtrees_dir_path)
```
jonsimantov, line 79: Maybe put this in utils as a utility function?

vimanyu: Good call. Done!

.github/workflows/desktop.yml
```diff
@@ -103,7 +103,7 @@ jobs:
           LANG: en_US
         run: |
           cd build
-          ctest --verbose
+          ctest --verbose -E "firestore_(objc_spec|objc_integration)_test"
```
a-maurice, line 106: I think it would be better to exclude them via https://github.com/firebase/firebase-cpp-sdk/blob/master/cmake/CTestCustom.cmake.

Either way, putting a comment explaining why it is being excluded would be good.

vimanyu: Yeah, I was contemplating where to put this. Ended up going with this approach because it is more visible and will be a reminder on every workflow update/run. But I noticed that there are few tests already explicitly disabled in CTestCustom.cmake. So maybe having one list of all disabled tests is better. What do you think?

vimanyu: Done!

PR #117, opened 2020-08-21
Add tool for sending mobile integration tests to FTL.

anonymous-akorn:
The main change is the addition of the Firebase Test Lab tool for testing mobile testapps. The third commit (c0f3bc7) has a detailed commit message explaining the tool and its differences from its counterpart in google3.

Another important change is to remove intermediate deps/objects after each integration test. These were found to contribute 400-1200 megabytes per project and are not needed after building. This should address the issue with Windows running out of disk space for the integration test workflow.


scripts/gha/test_lab.py
```diff
@@ -0,0 +1,334 @@
...
+
+  all_success = _report_results(tests, code_platform)
+  return 0 if all_success else 1
+
+
+def _get_base_results_dir():
+  """Defines the object used on GCS for all tests in this run."""
+  # We generate a unique directory to store the results by appending 4
+  # random letters to a timestamp. Using a timestamp is useful so that the
+  # the directory for different runs get sorted base on when it was run.
```
vimanyu, line 150: nit: gets sorted based

scripts/gha/test_lab.py
```diff
@@ -0,0 +1,334 @@
...
+  """Converts a relative GCS path to a GS URI understood by gsutil."""
+  # e.g. <results_dir> -> gs://<project_id>/<results_dir>
+  return "gs://%s/%s" % (_PROJECT_ID, path)
+
+
+def _gcs_list_dir(gcs_path):
+  """Recursively returns a list of contents for a directory on GCS."""
+  args = ["gsutil", "ls", "-r", gcs_path]
+  logging.info("Listing GCS contents: %s", " ".join(args))
+  result = subprocess.run(args=args, capture_output=True, text=True, check=True)
```
vimanyu, line 250: We have a function to run commands in `utils.py`. Though its trivial here, it would be nice to use common functions from a maintenance point of view (unless it doesnt meet your requirements).

scripts/gha/test_lab.py
```diff
@@ -0,0 +1,334 @@
+r"""Tool for sending mobile testapps to Firebase Test Lab for testing.
```
DellaBitta, line 1: should this r be here?

scripts/gha/build_testapps.py
```diff
@@ -457,10 +472,24 @@ def _run_setup_script(root_dir, testapp_dir):
     logging.info("setup_integration_tests.py not found")
 
 
-def _run(args, timeout=1200):
+def _run(args, timeout=2400, capture_output=False, text=None, check=True):
```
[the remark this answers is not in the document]
anonymous-akorn: Something similar to _rm_dir_safe already exists in utils, in the form of remove_directory. Uses the same underlying function (rmtree), but doesn't log anything if the directory doesn't exist, which I want to do here. Depending on application, the directory not existing may or may not warrant a warning. We could add arguments to control this kind of behaviour, but I think it makes more sense to use the primitives directly. 

As for _run, you bring up the method that already exists in utils in another comment, so I'll address it in that comment. Though even if we used that, we would still put a wrapper here to set the default timeout (which isn't universally appropriate for other files that might want to use utils).

scripts/gha/integration_testing/test_validation.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+def validate_results_unity(log_text):
+  """Determines the results in the log output of a Unity testapp."""
+  results_match = re.search(
+      r"PASS: (?P<pass>[0-9]+), FAIL: (?P<fail>[0-9]+)", log_text)
+  match_dict = results_match.groupdict()
+  if results_match:
+    # After the result string comes a list of failing testapps.
+    summary = log_text[results_match.start()]
```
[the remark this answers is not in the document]
anonymous-akorn: Whoops. Good catch.

scripts/gha/integration_testing/test_validation.py
```diff
@@ -0,0 +1,90 @@
...
+  """Determines the results in the log output of a testapp."""
+  if platform == UNITY:
+    return validate_results_unity(log_text)
+  elif platform == CPP:
+    return validate_results_cpp(log_text)
+  else:
+    raise ValueError("Invalid platform: %s." % platform)
+
+
+def validate_results_unity(log_text):
```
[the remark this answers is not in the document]
anonymous-akorn: Done

scripts/gha/test_lab.py
```diff
@@ -0,0 +1,334 @@
...
+
+Requires Cloud SDK installed with gsutil. Can be checked as follows:
+  gcloud --version
+
+This tool will use the games-auto-release-testing GCS storage bucket. To
+be authorized, it's necessary have the key file (JSON) (from valentine) and
+supply it with the --key_file flag.
+
+https://valentine.corp.google.com/#/show/1561166657633900
```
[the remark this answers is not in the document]
anonymous-akorn: I'd think it's fine, but I'll remove the url. If nothing else, might prevent someone outside of Google from punching this url into their browser and getting confused.

scripts/gha/test_lab.py
anonymous-akorn: Also two consecutive "the"s. Three mistakes in one sentence, must have been one of those days.

scripts/gha/test_lab.py
anonymous-akorn: From a maintenance point of view I think it's debatable in cases like this (invoking a single function with minimal additional logic): adding that bit of indirection means debugging requires tracking logic across another file and then unpacking the logic in that method. The only benefit would be if we needed to make changes to all uses of subprocess.run (if Python changed the api, for example), which wouldn't be a hard change regardless. 

As it so happens, after looking at the implementation in utils, it actually won't serve in all of my use cases since it always sets text to True if capture_output is True, which will cause the output to be encoded and normalized. This is usually fine, but not in all cases, namely when reading binary files like archives or encrypted files.

scripts/gha/test_lab.py
anonymous-akorn: That designates the string as a raw string, so that it ignores escape characters (such as the continuation in the example shell command). Similar to @ in C#.

DellaBitta: Gotcha. Thanks!

scripts/gha/build_testapps.py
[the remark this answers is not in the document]
vimanyu: Sounds good.

scripts/gha/test_lab.py
vimanyu: Another benefit is that it would standardize the way we are calling commands in various python scripts. 
We could address the `text` argument by setting its default to `None` and later setting it to `captureOutput` only if caller hasn't specified anything. 

I think desktop/android/ios(possibly) will probably be sharing these functions as much as possible.  Let me know what you think and I am okay with ignoring this and coming back to it if we feel it might be useful to have it in a central place.

anonymous-akorn: I don't feel too strongly about it, but adding a dependency on a custom module just for this feels a bit icky to me. I say we kick this bucket down the road - we should do a pass when everything's migrated over to see what we can refactor and clean up anyway.

vimanyu: Makes sense!

PR #124, opened 2020-09-01
Feature/fix database integration tests

vimanyu:
This should fix SEGFAULTS that were seen on running database integration tests. Issue was that "app_name" is usually a string with a "/" in it and mkdir calls on Mac and Windows were not creating directories recursively.

### DIFF
--- a/database/src/desktop/util_desktop_macos.mm
+++ b/database/src/desktop/util_desktop_macos.mm
@@ -23,48 +23,40 @@ namespace firebase {
 namespace database {
 namespace internal {
 
-// Given a directory path, and a directory creation mode with an option delimiter,
-// recursively create directories. mkdir doesn't create directories recursively
-// This is similar to `mkdir -p` on the command line
-static int mkdir_recursive(const char* dir_path, const mode_t mode) {
-  std::string dir_path_string(dir_path);
-  const char path_delimiter = '/';
-
-  // This index will be used to track the positions of path_delimiters in the path string
+// Split a string based on specified character delimiter into constituent parts
+static std::vector<std::string> split_string(const std::string& s,
+                                             const char delimiter='/') {
   size_t pos = 0;
-  // This index is used as the starting index to search the path_delimiters from.
-  size_t path_delimiter_search_start = 0;
-  // Skip any leading path_delimiters
-  while(dir_path_string[path_delimiter_search_start] == path_delimiter) {
-    path_delimiter_search_start++;
+  // This index is used as the starting index to search the delimiters from.
+  size_t delimiter_search_start = 0;
+  // Skip any leading delimiters
+  while(s[delimiter_search_start] == delimiter) {
+    delimiter_search_start++;
   }
-  size_t len = dir_path_string.size();
-  // Invalid path if it consists of just path_delimiters
+
+  std::vector<std::string> split_parts;
+  size_t len = s.size();
+  // Cant proceed if input string consists of just delimiters
   if (pos >= len) {
-    return -1;
+    return split_parts;
   }
-  std::string sub_directory;
-  int retval;
-  while( (pos = dir_path_string.find(path_delimiter, path_delimiter_search_start))
-                                                             != std::string::npos) {
-    sub_directory = dir_path_string.substr(0, pos);
-    retval = mkdir(sub_directory.c_str(), mode);
-    if (retval != 0 && errno != EEXIST) return -1;
-    while(dir_path_string[pos] == path_delimiter && pos<len) {
+
+  while( (pos = s.find(delimiter, delimiter_search_start)) != std::string::npos) {
+    split_parts.push_back(s.substr(delimiter_search_start, pos-delimiter_search_start));
+
+    while(s[pos] == delimiter && pos<len) {
       pos++;
-      path_delimiter_search_start = pos;
+      delimiter_search_start = pos;
     }
   }
 
-  // If the path does not have trailing delimiter, we still have to create the
-  // last subdirectory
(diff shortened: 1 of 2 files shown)


database/src/desktop/util_desktop_macos.mm
```diff
@@ -17,11 +17,47 @@
...
 namespace internal {
 
+// Split a string based on specified character delimiter into constituent parts
+static std::vector<std::string> split_string(const std::string& s,
+                                             const char delimiter='/') {
+  size_t pos = 0;
+  // This index is used as the starting index to search the delimiters from.
+  size_t delimiter_search_start = 0;
+  // Skip any leading delimiters
+  while(s[delimiter_search_start] == delimiter) {
```
DellaBitta, line 33: Can you use std::string.find here?

database/src/desktop/util_desktop_macos.mm
```diff
@@ -30,8 +66,17 @@
     int retval;
     retval = mkdir(path.c_str(), 0700);
     if (retval != 0 && errno != EEXIST) return "";
-    retval = mkdir((path + "/" + app_name).c_str(), 0700);
-    if (retval != 0 && errno != EEXIST) return "";
+    // App name might contain path separators. Split it to get list of subdirs
+    std::vector<std::string> app_name_parts = split_string(app_name, '/');
+    if(app_name_parts.empty()) return "";
```
DellaBitta, line 71: if's have spaces after if, as in "if (app_name_parts.empty()) return "";

database/src/desktop/util_desktop_macos.mm
```diff
@@ -30,8 +66,17 @@
...
+    std::string dir_path = path;
+    // Recursively create entire tree of directories
+    for (std::vector<std::string>::const_iterator it = app_name_parts.begin();
+                                            it != app_name_parts.end(); it++) {
+          dir_path = dir_path + "/" + *it;
```
DellaBitta, line 76: This tabbing looks off, as if there's one too many indents.

database/src/desktop/util_desktop_windows.cc
```diff
@@ -26,6 +26,41 @@ namespace firebase {
 namespace database {
 namespace internal {
 
+// Split a string based on specified character delimiter into constituent parts
```
DellaBitta, line 29: Formatting comments above apply here, too.

database/src/desktop/util_desktop_macos.mm
```diff
@@ -17,11 +17,47 @@
...
+  }
+
+  std::vector<std::string> split_parts;
+  size_t len = s.size();
+  // Cant proceed if input string consists of just delimiters
+  if (pos >= len) {
+    return split_parts;
+  }
+
+  while( (pos = s.find(delimiter, delimiter_search_start)) != std::string::npos) {
```
DellaBitta, line 44: while ((

database/src/desktop/util_desktop_macos.mm
```diff
@@ -17,11 +17,47 @@
...
+    while(s[pos] == delimiter && pos<len) {
+      pos++;
+      delimiter_search_start = pos;
+    }
+  }
+
+  // If the input string doesn't end with a delimiter we need to push the last
+  // token into our return vector
+  if (delimiter_search_start != len)
+    split_parts.push_back(s.substr(delimiter_search_start, len-delimiter_search_start));
```
DellaBitta, line 56: recommend putting this in brackets, like the if (pos >= len) above.

vimanyu: Done!

database/src/desktop/util_desktop_macos.mm
vimanyu: Done!

database/src/desktop/util_desktop_windows.cc
vimanyu: Done!

database/src/desktop/util_desktop_macos.mm
vimanyu: Done!

database/src/desktop/util_desktop_macos.mm
vimanyu: Done!

database/src/desktop/util_desktop_macos.mm
vimanyu: I feel for this purpose, it will make the code real hard to follow given the edge cases.
From `find`, we can find the first position of delimiter but it could be a leading one or not. 
For the leading case, we will have to issue multiple `find`(s) repeatedly until we have found the last consecutive delimiter (Eg: ///Library/foo/bar)

DellaBitta: Gotcha.

DellaBitta (dismissed):

database/src/desktop/util_desktop_windows.cc
```diff
@@ -26,6 +26,42 @@ namespace firebase {
...
+static std::vector<std::string> split_string(const std::string& s,
```
a-maurice, line 30: If I'm reading this right, there is no difference between this and the above one. Should probably be moved to the shared util_desktop.cc/h

a-maurice (changes requested):

database/src/desktop/util_desktop_windows.cc
vimanyu: Good call. This function went from a platform dependent `mkdir` to a simple string split that should be in a common location. Moved it.

a-maurice (approved):

PR #125, opened 2020-09-02
Feature/fix firestore integration tests build

vimanyu:
Fixes linker symbol missing errors discovered when building firestore's integration tests. This dependency `absl_types` is used just by the C++ part of Firestore and hence was not pushed to "firebase-ios-sdk" repo. 

We were seeing errors like these,
undefined reference to `absl::lts_2020_02_25::variant_internal::ThrowBadVariantAccess()

### DIFF
--- a/firestore/CMakeLists.txt
+++ b/firestore/CMakeLists.txt
@@ -179,13 +179,13 @@ target_link_libraries(firebase_firestore
   PUBLIC
     firebase_app
     firebase_auth
-    absl_variant
 )
 
 if(FIRESTORE_USE_EXTERNAL_CMAKE_BUILD)
   target_link_libraries(firebase_firestore
     PRIVATE
       firestore_core
+      absl_variant
   )
 endif()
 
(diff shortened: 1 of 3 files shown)


firestore/CMakeLists.txt
```diff
@@ -179,6 +179,7 @@ target_link_libraries(firebase_firestore
   PUBLIC
     firebase_app
     firebase_auth
+    absl_variant
```
a-maurice, line 182: Pretty sure this should only be needed with firestore_core, so it'd be better to do it in the thing below. There are some platforms that firestore_core is not used, which that FIRESTORE_USE_EXTERNAL_CMAKE_BUILD flag captures

a-maurice (changes requested):
Seems like the database fixes accidentally got included in this

firestore/CMakeLists.txt
vimanyu: Makes sense.
Had to create another PR because this branch had some stuff accidentally merged in.
https://github.com/firebase/firebase-cpp-sdk/pull/133
Closing this one.

PR #132, opened 2020-09-03
Fix Storage Long Download Pause n' Cancel Integration Test

DellaBitta:
Fixes for the Storage LongDownloadSuspendAndCancel test which was completing too quickly on GHA for the pause and cancel to take place.  Reworked the test to detect when transfers have started instead of waiting xyz milliseconds.


storage/integration_test/src/integration_test.cc
```diff
@@ -773,4 +788,4 @@ TEST_F(FirebaseStorageTest, TestInvalidatingReferencesWhenDeletingApp) {
   EXPECT_FALSE(metadata.is_valid());
 }
 
-}  // namespace firebase_testapp_automated
+}  // namespace firebase_testapp_automated
```
vimanyu, line 791: nit: no newline at end of file

storage/integration_test/src/integration_test.cc
```diff
@@ -588,7 +588,10 @@ TEST_F(FirebaseStorageTest, TestLargeFilePauseResumeAndDownloadCancel) {
...
     ASSERT_TRUE(controller.is_valid());
 
-    ProcessEvents(500);
+    while(controller.bytes_transferred() == 0) 
+    {
+      ProcessEvents(1);
+    }
     // After waiting a moment for the operation to start, pause the operation
     // and verify it was successfully paused. Note that pause might not take
     // effect immediately, so we give it a few moments to pause before
```
anonymous-akorn, line 597: Do the comments need to be updated to reflect the new strategy?

vimanyu (commented):
Looks good except a super minor nit.

storage/integration_test/src/integration_test.cc
DellaBitta: Nope, this is the same behavior as before.

anonymous-akorn: hmm I guess we're just being smarter about how we're "waiting a moment". Fair.

DellaBitta: Though I don't think that comment was accurate even in the previous implementation.  updating it.

storage/integration_test/src/integration_test.cc
DellaBitta: Done!

anonymous-akorn (approved):

vimanyu (approved):
LGTM

PR #134, opened 2020-09-08
Fix for Windows Storage Integration Test timeout

DellaBitta:
Fixes for the Storage integration test.  The test would timeout on Windows when attempting to sign in a user anonymously, though it would work for the first sign in attempt, just not subsequent ones.

I pulled in the auth teardown logic.

Additionally removed an unneeded check for app initialization in the teardown method.  This check is already being done by the sub-methods.


vimanyu (approved):
LGTM!

PR #139, opened 2020-09-09
Add Android integration testing to integration test workflow

anonymous-akorn:
This modifies the integration test workflow to use the test_lab.py script to send testapps to Firebase Test Lab, and validates the results.

This requires a GCP service account. To authorize its use, we need a key file, which is added to the encrypted files in gha-encrypted. Note that this is not mirrored in google3, since policy prohibits this kind of data from being placed in google3. Instead, the unencrypted file is found in valentine, its valentine id mentioned in the test_lab.py file.

To install the Cloud SDK, we use GCP's setup-gcloud action. I have not found a particularly convenient way to automate the installation of the Cloud SDK so we use this as a (possibly temporary) convenience. In the future, it may be replaced by custom logic so that we can simplify local setup.


.github/workflows/integration_tests.yml
```diff
@@ -87,9 +99,16 @@ jobs:
...
           
-      - name: Build and run integration tests
+      - name: Build integration tests (and run Desktop tests)
         # The set up script for Android will download the NDK here.
         env:
           NDK_ROOT: '/tmp/android-ndk-r16b'
         run: |
           python scripts/gha/build_testapps.py --t ${{ github.event.inputs.apis }} --p ${{ matrix.target_platform }} --output_directory ${{ github.workspace }} --use_vcpkg --execute_desktop_testapp --noadd_timestamp
+      - name: Install Cloud SDK for mobile integration tests
+        if: matrix.target_platform != 'Desktop' && ( success() || failure() )
```
DellaBitta, line 109: What's the ( success() || failure() ) about?

DellaBitta (approved):
Approved but with a comment.

.github/workflows/integration_tests.yml
anonymous-akorn: Those are recognized by GHA. 

success() will be true if there were no failed steps prior to this step (all steps implicitly use success() by default), failure() will be true if any prior step failed. This ensures the tests will run on any integration tests that built successfully, even if others failed (which causes the build step to report failure), instead of the tests being aborted.

There's also an always(), but that includes cancelled(), which we don't want.

DellaBitta: Would !cancelled() work?  Seems like it would provide more information about the intention of the check and not confused ignorant people like me who think that things either succeed or fail.

anonymous-akorn: It would, yes. The behaviour could diverge in the future, and I'm not sure which is more desirable: "not cancelled" would pick up new statuses that Github might add, which could be good or bad. But given the uncertainty, might as go with the more readable option.

vimanyu (approved):
LGTM

PR #140, opened 2020-09-10
Fix Database integration test on Windows 

DellaBitta:
Added enforced auth signout on every test teardown.  Signout must be done after the database cleanup is done.


DellaBitta:
[Integration tests passing on Desktops](https://github.com/firebase/firebase-cpp-sdk/actions/runs/248171872).

anonymous-akorn (approved):

PR #141, opened 2020-09-11
CMake build Framework for iOS

sunmou99:
Currently, CMake builds library for iOS. This PR modifies the CMake to build Framework for iOS.

The [build cmd ](https://docs.google.com/document/d/18FEitEefi6vaK1VtzU7FtJjDnj6Xj8ijb0JqIRpUVB8/edit?usp=sharing)is the same as it use to be.

[Design Doc](https://docs.google.com/document/d/18FEitEefi6vaK1VtzU7FtJjDnj6Xj8ijb0JqIRpUVB8/edit?usp=sharing)


app/CMakeLists.txt
```diff
@@ -374,3 +456,26 @@ cpp_pack_public_headers()
 if (NOT ANDROID AND NOT IOS)
   cpp_pack_library(flatbuffers "deps/app/external")
 endif()
+
+if (IOS)
+  set_target_properties(firebase_app PROPERTIES
+    FRAMEWORK TRUE
+  )
+
+  # Generate Framewrok Headers
```
vimanyu, line 465: nit: Framework

There is quite a bit of post processing happening on the file names. Will be awesome to have a comment explaining the transformations.

app/CMakeLists.txt
```diff
@@ -288,6 +288,88 @@ else()
...
+    "${FIREBASE_SOURCE_DIR}/analytics/ios_headers/FIRUserPropertyNames.h"
+    "${analytics_generated_headers_dir}/user_property_names.h"
+  )
+  set(analytics_HDRS
+    "${FIREBASE_SOURCE_DIR}/analytics/src/include/firebase/analytics.h"
+    "${analytics_generated_headers_dir}/event_names.h"
+    "${analytics_generated_headers_dir}/parameter_names.h"
+    "${analytics_generated_headers_dir}/user_property_names.h")
+
+  file(GLOB auth_HDRS
```
vimanyu, line 329: Though GLOB makes it easy to read and follow the CMake logic, it can have surprising side effects. If an explicit list is not crazy long, I prefer explicitly listing down the files.

https://stackoverflow.com/questions/32411963/why-is-cmake-file-glob-evil/32412044
Example from CMakeLists in our cpp sdk repo,
https://github.com/firebase/firebase-cpp-sdk/blob/master/database/CMakeLists.txt#L97

Not a big deal if you want to keep it the same as it is now.

vimanyu (commented):
Looks good in general. Minor points to think about but dont necessarily require changes.

PR #145, opened 2020-09-17
Add iOS frameworks build

sunmou99:
Added ios build script (replaced the old one). The generated frameworks locate at dir "/ios_build" by default.
Added ios build workflow.

Tried to build different architectures in parallel, but it's more likely to break.


build_scripts/ios/build.sh
```diff
@@ -1,37 +1,130 @@
...
 
-cmake -GXcode "-DCMAKE_TOOLCHAIN_FILE=$toolchain" -S "${sourcepath}" -B "${buildpath}"
-cmake --build "${buildpath}"
+frameworks=$(ls arm64)
+for framework in ${frameworks}; do
+    mkdir -p universal/${framework}
+    libbasename="$(basename "${framework}" .framework)"
+    libsubpath="${framework}/${libbasename}"
+    if [[ -f "arm64/${libsubpath}" ]]; then
+        lipo -create "arm64/${libsubpath}" "armv7/${libsubpath}" "x86_64/${libsubpath}" "i386/${libsubpath}" \
```
vimanyu, line 125: Could this list be prepared from SUPPORTED_ARCHITECTURES instead of hard coding the names?

app/CMakeLists.txt
```diff
@@ -464,15 +464,13 @@ if (IOS)
 
   # Generate Framewrok Headers
   foreach(hfile ${internal_HDRS})
-    message(STATUS "hfile: ${hfile}")
     string(FIND ${hfile} "firebase/" pos_firebase)
```
vimanyu, line 467: As mentioned in the other PR, will be great to have a simple example in comments on various string transformations.

vimanyu (commented):
Looks very clean overall! 
Super minor comments.

paulb777 (commented):
It could be worth sync'ing up with the iCore team before going too far on developing scripts to generate frameworks.  Apple is deprecating frameworks in favor of xcframeworks with the onset of Catalyst and arm64 macs.

It also might make sense to collaborate more on tooling since we already have some tooling for transforming Firebase sources to binaries at https://github.com/firebase/firebase-ios-sdk/tree/master/ZipBuilder

sunmou99:
@paulb777 Thanks for calling out. 
I talked with Gran, it seems that ZipBuilder is not the tool we could use here. 
xcframeworks sounds good, but CMake doesn't support xcframeworks. Also, I don't know if xcframeworks will affect the test codes. 
Who (from iCore) should I reach out to for more information?

a-maurice (dismissed):

a-maurice (approved):

PR #148, opened 2020-09-17
Build & Test iOS integration test against repo

sunmou99:
Workflow to build & test iOS integration test
 - add build iOS framework against repo
 - change to build unsigned .ipa and delete signing code.
 - build vcpkg only for Desktop platform 
 - fix bug b/163338992
 - exclude [iOS,ubuntu-latest] [iOS,windows-latest] from matrix
 - install gcloud components "beta" 
 - add function to set bundle id.


anonymous-akorn:
Could you add some context on this PR? It looks like you're merging one of your feature branches into another one of your feature branches, which doesn't require approval. Are you just looking for some early input on these particular changes?

sunmou99:
Talked with akorn about the context on this PR.

scripts/gha/build_testapps.py
```diff
@@ -195,15 +184,17 @@ def main(argv):
   else:
     timestamp = ""
 
+  config = config_reader.read_config()
+
+  if _IOS in platforms:
+    _build_ios_framework_from_repo(FLAGS.sdk_dir, testapps, config)
```
anonymous-akorn, line 190: If any framework fails to build, the entire script will stop, which is not what we want: errors in iOS should not affect desktop or Android, and a failure for database should not affect auth. 

It would be better if the frameworks needed for a given API were built within the try block (line 268) for the iOS build of that platform.

anonymous-akorn (changes requested):
Looks mostly good at a high level.

scripts/gha/integration_testing/xcodebuild.py
```diff
@@ -27,29 +27,24 @@
 
 The methods support either device or simulator builds. For simulator
 builds, it suffices to use gets_args_for_build to create a .app that can be
-used with simulators. For device builds, it's necessary to use both a
-get_args_for_archive step and a get_args_for_export step. The archive step
-generates an .xcarchive, and the export step 'exports' this archive to create
-an .ipa that can be installed onto devices.
+used with simulators. For unsigned device builds, it's necessary to 
+get_args_for_unsigned_build step and then package the .app to .ipa.
```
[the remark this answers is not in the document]
sunmou99: Will do

scripts/gha/integration_testing/xcodebuild.py
```diff
@@ -58,18 +53,17 @@ def get_args_for_archive(
...
   """
   xcode_args = get_args_for_build(
-      path, scheme, output_dir, ios_sdk, dev_team, configuration)
+      path, scheme, output_dir, ios_sdk, configuration)
   additional_device_args = [
-      "PROVISIONING_PROFILE=" + uuid,
-      "CODE_SIGN_STYLE=Manual",
-      "-archivePath", archive_path,
-      "archive"
+      "CODE_SIGN_IDENTITY=\"\"",
```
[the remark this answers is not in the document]
sunmou99: But it will looks inconsistent with other args.

scripts/gha/integration_testing/xcode_tool.rb
```diff
@@ -123,7 +125,22 @@ def add_entitlements
   puts 'Added entitlement to xcode project.'
 end
 
-# This only involves patching a plist file, which should be moved to the
+# This only involves patching a plist file, which could be moved to the
+# Python tool.
+def set_bundle_id
```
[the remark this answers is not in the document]
sunmou99: Not all the xcode projects have this field set: 
https://github.com/firebase/firebase-cpp-sdk/blob/feature/gha-ios-integration-tests/messaging/integration_test/integration_test.xcodeproj/project.pbxproj#L206

I also remember FPL has a script to set PRODUCT_BUNDLE_IDENTIFIER in google3.

scripts/gha/build_testapps.py
```diff
@@ -407,7 +420,7 @@ def _build_ios(
   entitlements_path = os.path.join(
       project_dir, api_config.ios_target + ".entitlements")
   xcode_tool_path = os.path.join(
-      testapp_builder_dir, "integration_testing", "xcode_tool.rb")
+      root_dir, "scripts", "gha", "integration_testing", "xcode_tool.rb")
```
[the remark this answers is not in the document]
sunmou99: They are not the same path. I remember testapp_builder_dir is a subdirectory in project_dir.

scripts/gha/integration_testing/xcodebuild.py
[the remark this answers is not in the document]
anonymous-akorn: More readable, though. PEP8 explicitly prescribes doing this to avoid escape characters in the string:
https://www.python.org/dev/peps/pep-0008/#string-quotes

scripts/gha/test_lab.py
```diff
@@ -295,7 +295,7 @@ def _gcloud_command(self):
     if self.platform == _ANDROID:
       cmd = ["gcloud", "firebase", "test", "android", "run"]
     elif self.platform == _IOS:
-      cmd = ["gcloud", "beta", "firebase", "test", "ios", "run"]
+      cmd = ["gcloud", "--quiet", "beta", "firebase", "test", "ios", "run"]
```
anonymous-akorn, line 298: Should the quiet flag be added to the Android command as well?

scripts/gha/integration_testing/xcodebuild.py
```diff
@@ -142,3 +83,19 @@ def _get_ios_env_from_target(ios_sdk):
     return "iphonesimulator"
   else:
     raise ValueError("Unrecognized ios_sdk: %s" % ios_sdk)
+
+
+def generate_unsigned_ipa(build_dir):
```
anonymous-akorn, line 88: There are a few undocumented assumptions/behaviours here, which should either be explained in the docstring or exposed as arguments.

- There's a subdirectory called "Debug-iphoneos" in the build dir, which implicitly assumes we ran xcodebuild with "Debug" configuration.
- "build_dir" is the directory containing Debug-iphoneos, and not Debug-iphoneos itself.
- The resulting ipa will be called "integration_test.ipa", and will be placed within Debug-iphoneos, not build_dir.

scripts/gha/build_testapps.py
```diff
@@ -195,15 +184,18 @@ def main(argv):
   else:
     timestamp = ""
 
+  config = config_reader.read_config()
+
+  github_repo = False
```
anonymous-akorn, line 189: This variable is misleading - it will only be set to True if we're building from repo **and** building for iOS. If someone decides to add any logic that depends on whether we're building from repo for Desktop or Android, they'll run into problems.

It would make more sense to move the check out of _generate_makefiles_from_repo, and only run that method if github_repo is true and we're building for iOS. We should also log it (whether we've detected that we're building from repo) for debugging purposes.

scripts/gha/build_testapps.py
```diff
@@ -386,9 +374,56 @@ def _validate_android_environment_variables():
...
+        "-s", sdk_dir,
+        "-c", "false"
+    ]
+    _run(framework_builder_args)
+    return True
+  
+  return False
+
+
+# If sdk_dir is Github repo, then build ios frameworks first
```
anonymous-akorn, line 399: Is this identical comment supposed to be before both _build_ios_framework_from_repo and _generate_makefiles_from_repo?

anonymous-akorn (changes requested):

scripts/gha/test_lab.py
sunmou99: There is no interactive prompts for android run. But probably yes.

scripts/gha/integration_testing/xcodebuild.py
sunmou99: Will add document, and change the variables' name.

scripts/gha/test_lab.py
sunmou99: Change the components installation into yml.

.github/workflows/integration_tests.yml
```diff
@@ -108,6 +118,10 @@ jobs:
       - name: Install Cloud SDK for mobile integration tests
         if: matrix.target_platform != 'Desktop' && !cancelled()
         uses: GoogleCloudPlatform/github-actions/setup-gcloud@master 
+      - name: Install Google Cloud beta components for iOS integration tests
+        if: matrix.target_platform == 'iOS'
+        run: |
+          gcloud --quiet components install beta
```
anonymous-akorn, line 124: Is there any harm in putting this command in test_lab.py? This will be needed both locally and on GHA, so it seems sensible to put it in the script instead.

anonymous-akorn (changes requested):
Looks like there's a merge conflict? Also some of these files look fairly outdated, like they're missing changes from dev. Might need to merge from dev.

.github/workflows/integration_tests.yml
sunmou99: There is no harm putting it in test_lab.py, except it require more code and code looks not good(an extra variable to make it only for iOS) in test_lab.py.
To run it locally, there is interactive prompts. Just input "Y" to install the components.

sunmou99:
> Looks like there's a merge conflict? Also some of these files look fairly outdated, like they're missing changes from dev. Might need to merge from dev.

Yes. I noticed it too. But I have resolved the merge conflict more than 3 times. So, I decide not to resolve it until feature/package-ios-frameworks get merged. 
The files are rebased/merge from dev this Monday, not quite outdated. Again, I will rebased/merge from dev after feature/package-ios-frameworks get merged.

.github/workflows/integration_tests.yml
anonymous-akorn: Having to include this step in every workflow that uses the script is a significant downside. 

To avoid having to add another flag, the logic can choose to run this command if we find any iOS testapps, i.e. could set a boolean here:
https://github.com/firebase/firebase-cpp-sdk/blob/dev/scripts/gha/test_lab.py#L119

sunmou99: This is what I mean the code looks not good. It is because we build iOS and then we got IPAs. It's strange that we say we have IPAs, thus we are building for iOS (especially in a loop).
1. To write it into the script, I prefer adding a flag to test_lab.py itself. Like: --p ${{ matrix.target_platform }}
2. Or we could put it into step "Install Cloud SDK for mobile integration tests":
`      - name: Install Cloud SDK for mobile integration tests
        if: matrix.target_platform != 'Desktop' && !cancelled()
        uses: GoogleCloudPlatform/github-actions/setup-gcloud@master 
        run: |
          gcloud --quiet components install beta
`
   Thus, this step installs both gcloud and gcloud compoent beta. Though it installs gcloud compoent beta for Android, it only takes about 30 seconds.

anonymous-akorn: The logic is "we passed a directory with testapps to this tool; one or more of these testapps is for iOS; thus we need to install tools for testing iOS". That's intuitive, no? 

The flag would add complexity to its usage, and is redundant since we need to install these tools if and only if the directory of testapps include iOS testapps, information we already have available to us in the context of the tool.

anonymous-akorn (approved):

PR #154, opened 2020-10-06
Feature/cmake ios framework

sunmou99:
Currently, CMake builds library for iOS. This PR modifies the CMake to build Framework for iOS.

The [build cmd](https://github.com/firebase/firebase-cpp-sdk#building-with-cmake-for-ios) is the same as it use to be.

[Design Doc](https://docs.google.com/document/d/18FEitEefi6vaK1VtzU7FtJjDnj6Xj8ijb0JqIRpUVB8/edit#heading=h.xzptrog8pyxf)


a-maurice (approved):
There might be a better way to do the headers, but this is a good start.

PR #156, opened 2020-10-06
Add Windows integration tests to workflow

anonymous-akorn:
This contains the following high level changes:

1) Adding the setup-msbuild step, as used by the workflow to build the SDK.
2) Working around an issue with the gCloud SDK on Windows, by setting an environment variable (CLOUDSDK_PYTHON)
3) Fixing some issues with using gradle on Windows: build_testapps.py patches a path into gradle.settings, but gradle does not handle Windows path separators. As part of this, all paths provided through flags are passed through a function to normalize them and convert relative paths to full paths. Additionally, we now use gradlew.bat on Windows, instead of gradlew.
4) Fixed an issue with gcloud and gsutil not being found on Windows. subprocess.run does not check the system path, so e.g. subprocess.run(["gcloud"]) was failing with a FileNotFoundError, despite gcloud being in the environment's path. This was fixed by using full paths to the Google Cloud command line tools.


.github/workflows/integration_tests.yml
```diff
@@ -105,6 +110,11 @@ jobs:
           NDK_ROOT: '/tmp/android-ndk-r16b'
         run: |
           python scripts/gha/build_testapps.py --t ${{ github.event.inputs.apis }} --p ${{ matrix.target_platform }} --output_directory ${{ github.workspace }} --use_vcpkg --execute_desktop_testapp --noadd_timestamp
+      # Workaround for https://github.com/GoogleCloudPlatform/github-actions/issues/100
```
DellaBitta, line 113: Nit: add a new line above

DellaBitta (approved):
Approved with a nit.

DellaBitta (approved):

PR #165, opened 2020-10-16
Messaging IntegrationTest min Android sdk updated to v26

DellaBitta:
Updated Messaging Integration Test build.gradle file to require Android SDK 26. 

`src/android/java/LibraryManifest` declares minSDK version 26 as a dependency due to its new  use of JobIntentService. As of Android O, background services can no longer spawn other background services, they must instead create JobIntentServices, which can queue up work for the OS to perform when it is ready.


vimanyu (approved):
LGTM

vimanyu (approved):

PR #169, opened 2020-10-19
Feature/update podfile

sunmou99:
Update integration testapps' Podfiles to match the SDK's Podfile.

Script reference: https://github.com/firebase/firebase-ios-sdk/blob/master/scripts/localize_podfile.swift


anonymous-akorn:
I'm curious about the choice to use swift for this script. It looks like the logic in the script boils down to this:

- Read a file.
- Look for lines of the form " pod '<api>', '<version>'", and store a mapping of api -> version.
- Read another file.
- Look for lines of the same form, and replace their version with the version in the mapping.

Given that we're defaulting to Python for our scripts, would it make sense to write this logic in Python instead? Or is there a benefit to leaving it in swift?

sunmou99:
> Given that we're defaulting to Python for our scripts, would it make sense to write this logic in Python instead? Or is there a benefit to leaving it in swift?

I was thinking about Python in the first place, but @granluo told me that iCore team uses swift script to update podfile.
So I use swift for the sake of consistency. And Gran probably gonna have a similar swift script in his 'cpp ios sdk unrelease testing project'.

anonymous-akorn:
> I was thinking about Python in the first place, but @granluo told me that iCore team uses swift script to update podfile.
> So I use swift for the sake of consistency. And Gran probably gonna have a similar swift script in his 'cpp ios sdk unrelease testing project'.

The Games team will be the ones maintaining this, though - leaving it in swift means another language they need to use. Though the script is unlikely to change in the future, so it's probably not a big deal. I'd suggest getting someone on the Games team to review the PR, since their approval is more relevant than mine for this. 

Beyond that, LGTM.

granluo:
I think using Python could be an option for the script. The iOS team use Swift because Swift is their preferred language for command line tools. I would recommend Python since 1) Python is the major language for infra tools in Games team, and 2) it would be easier if other Python scripts want to reuse this tool. Unreleased testing could apply this tool and reviews for updates of this tool might be required.

sunmou99:
@granluo Updated to python script.

scripts/gha/integration_testing/update_podfile.py
```diff
@@ -0,0 +1,66 @@
...
+    raise app.UsageError("Too many command-line arguments.")
+
+  sdk_podfile = FLAGS.sdk_podfile
+  app_podfile = FLAGS.app_podfile
+
+  # dict to store podName & podVersion from sdk's podfile
+  pod_dict = {}
+  with open(sdk_podfile) as f:
+    for line in f:
+      tokens = re.split(' |,', line)
```
granluo, line 47: Maybe we can set the pattern `' |,'` as a global variable which could be used in app_podfile as well since they are both Podfiles and could use the same pattern to split tokens.

DellaBitta (approved):
LGTM

granluo (dismissed):

scripts/gha/integration_testing/update_podfile.py
```diff
@@ -0,0 +1,66 @@
...
+
+  # update podVersion in app's podfile
+  with fileinput.input(app_podfile, inplace=True) as f:
+    for line in f:
+      tokens = re.split(' |,', line)
+      tokens = list(filter(None, tokens))
+      if tokens[0] == 'pod' and tokens[1] in pod_dict:
+        print(line.replace(tokens[2], pod_dict[tokens[1]]).rstrip())
+      else:
+        print(line.rstrip())
```
anonymous-akorn, line 60: I was confused as to what was going on here, but after reading up on fileinput, it looks like this will create a backup of the file, hijack standard output, and then write standard output to the original file location? That's clever, but horrifying. Among other things, it means that in this block, print does something very unusual, and adding a print statement for debugging purposes will corrupt the podfile.

The standard approach would be to read the file, modify the contents in-memory, then overwrite the file. I'd strongly recommend doing that instead. Even the fileinput module itself recommends using 'open' instead when dealing with single files:

"This module implements a helper class and functions to quickly write a loop over standard input or a list of files. If you just want to read or write one file see open()."

anonymous-akorn (changes requested):

anonymous-akorn (approved):

PR #170, opened 2020-10-19
fix storage integration testapp for iOS 

sunmou99:
I believe this commit causing the iOS test fail: https://github.com/firebase/firebase-cpp-sdk/commit/3ece8b2f28ade78f6b8991833d01544ac31d4866# @DellaBitta 

And I was wondering if should we sync the changes into google3.

### DIFF
--- a/storage/integration_test/src/integration_test.cc
+++ b/storage/integration_test/src/integration_test.cc
@@ -607,11 +607,14 @@ TEST_F(FirebaseStorageTest, TestLargeFilePauseResumeAndDownloadCancel) {
     // operation.
     ASSERT_TRUE(controller.is_valid());
 
-    ProcessEvents(500);
-    // After waiting a moment for the operation to start, pause the operation
-    // and verify it was successfully paused. Note that pause might not take
-    // effect immediately, so we give it a few moments to pause before
-    // failing.
+    while(controller.bytes_transferred() == 0) 
+    {
+      ProcessEvents(500);
+    }
+    
+    // After waiting a moment for the operation to start (above), pause the
+    // operation and verify it was successfully paused when the future
+    // completes.
     LogDebug("Pausing upload.");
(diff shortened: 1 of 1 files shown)


DellaBitta:
What aspect of the test is broken by the prior change?  Also there were multiple locations where I replaced a simple timed wait with a tight loop waiting for data to be sent.  Does this change need to be made at all of those other locations, too?

sunmou99:
> What aspect of the test is broken by the prior change? 

It breaks FirebaseStorageTest.TestLargeFilePauseResumeAndDownloadCancel in [iOS](https://github.com/firebase/firebase-cpp-sdk/runs/1278097351?check_suite_focus=true). 

> Also there were multiple locations where I replaced a simple timed wait with a tight loop waiting for data to be sent. Does this change need to be made at all of those other locations, too?
I don't think so. Some locations doesn't affect Mobile Platform, and others doesn't break the mobile tests. I believe this is the only place need to change. iOS tests [passed](https://github.com/firebase/firebase-cpp-sdk/runs/1278389281?check_suite_focus=true) after the change.

anonymous-akorn:
Additional context in b/171415295

Is it possible this is also causing the issues for Android?

sunmou99:
> Is it possible this is also causing the issues for Android?

I don't think so. https://github.com/firebase/firebase-cpp-sdk/commit/3ece8b2f28ade78f6b8991833d01544ac31d4866 has nothing to do with 'FirebaseStorageTest.TestDeleteFile'. Also, the Android (Mac) tests [passed](https://github.com/firebase/firebase-cpp-sdk/runs/1278097336?check_suite_focus=true) on dev branch.

DellaBitta:
I verified the test on Android as well. Great!

DellaBitta (approved):

anonymous-akorn (approved):

PR #174, opened 2020-10-28
Fix NDK env var discrepancy in integration tests

anonymous-akorn:
NDK_ROOT was being set, but not ANDROID_NDK_HOME, the latter pointing to the VM's default installation of NDK21 or so. This was causing a build failure for the integration tests.


jonsimantov (dismissed):

anonymous-akorn:
Looks like this doesn't quite work properly on Windows.

install_prereqs.sh downloads the NDK to a UNIX path "/tmp/...", which translates to something very different on Windows (C:\Users\<user>\AppData\...). But when I set an env var in the workflow file to that, it does not get translated to such a path, so the NDK isn't found on Windows. Going to test a different way to set the env vars.

Didn't run into the issue previously since it was apparently using the default value for ANDROID_NDK_HOME, which pointed to a valid path for the NDK already on the VM.

jonsimantov:
> Looks like this doesn't quite work properly on Windows.
> 
> install_prereqs.sh downloads the NDK to a UNIX path "/tmp/...", which translates to something very different on Windows (C:\Users<user>\AppData...). But when I set an env var in the workflow file to that, it does not get translated to such a path, so the NDK isn't found on Windows. Going to test a different way to set the env vars.
> 
> Didn't run into the issue previously since it was apparently using the default value for ANDROID_NDK_HOME, which pointed to a valid path for the NDK already on the VM.

There are a number of Windows Android issues, I believe @vimanyu was considering disabling the Windows Android workflows and making it a separate task to get it working again.

jonsimantov:
Also, with the change I merged into dev, it should actually build now if you stop setting ANDROID_NDK_HOME (it shouldn't try to build those alternate ABIs with the default NDK) which might be sufficient for now.

PR #179, opened 2020-11-06
Feature/aks prebuilt integration tests

anonymous-akorn:
Add integration tests to the packaging workflow.

This runs all integration tests on Desktop, Android and Windows, against either the SDK build by the packaging workflow, an already released SDK, or an SDK from a previous packaging run. 

By default the integration tests do not run (controlled by an input). This may be changed once the tests are green. This PR also includes a change to use the default NDK on the VMs (NDK 21), as using NDK 16 was running into problems.


jonsimantov:
> This PR also includes a change to use the default NDK on the VMs (NDK 21), as using NDK 16 was running into problems.

Does this mean the integration tests are built using NDK 21, but the SDK is still built using NDK 16?

anonymous-akorn:
> Does this mean the integration tests are built using NDK 21, but the SDK is still built using NDK 16?

Assuming the released SDK is built with NDK 16 (which I believe is the case), yes.

.github/workflows/cpp-packaging.yml
```diff
@@ -601,45 +604,75 @@ jobs:
...
-        if: runner.os == 'macOS'
-        run: sudo xcode-select -s /Applications/Xcode_${{ env.xcodeVersion }}.app/Contents/Developer
-
-      - name: download SDK zip
-        uses: actions/download-artifact@v2
-        with:
-          name: firebase_cpp_sdk.zip
-          path: .
-
-      - name: download hash
```
DellaBitta, line 620: Was the hash check step purposefully removed?

scripts/gha/build_testapps.py
```diff
@@ -362,31 +362,45 @@ def _validate_android_environment_variables():
...
+    logging.warning("No NDK env var set. Set one of %s", ", ".join(ndk_vars))
 
 
-# If sdk_dir contains no framework, consider it is Github repo, then 
+# If sdk_dir contains no framework, consider it is Github repo, then
 # generate makefiles for ios frameworks
 def _generate_makefiles_from_repo(sdk_dir):
   ios_framework_builder = os.path.join(
-    sdk_dir, "build_scripts", "ios", "build.sh")
+      sdk_dir, "build_scripts", "ios", "build.sh")
```
DellaBitta, line 403: Nit: are these indentation changes correct?   Most other instances seem to be 2 spaces.

DellaBitta (changes requested):

.github/workflows/cpp-packaging.yml
anonymous-akorn: Yes, the hash command is not present on all operating systems.

scripts/gha/build_testapps.py
anonymous-akorn: Yes, normal indents are 2 spaces, hanging indents are 4 spaces.

DellaBitta (commented):
Addressed my concerns.  Thank you.

DellaBitta (approved):

scripts/gha/build_testapps.py
```diff
@@ -400,15 +414,15 @@ def _generate_makefiles_from_repo(sdk_dir):
 # build required ios frameworks based on makefiles
 def _build_ios_framework_from_repo(sdk_dir, api_config):
```
vimanyu, line 415: docstring... Looks like api_config needs to have a `frameworks` property.

scripts/gha/build_testapps.py
```diff
@@ -400,15 +414,15 @@ def _generate_makefiles_from_repo(sdk_dir):
...
   ios_framework_builder = os.path.join(
-    sdk_dir, "build_scripts", "ios", "build.sh")
-  
+      sdk_dir, "build_scripts", "ios", "build.sh")
+
   # build only required targets to save time
   target = set()
   for framework in api_config.frameworks:
     target.add(os.path.splitext(framework)[0])
```
vimanyu, line 422: Will be nice to see an example string in comments of what is expected to happen. I had to follow the calls to see an example of what values could possibly come as framework.

vimanyu (commented):
Looks good overall. Super minor comments.

.github/workflows/cpp-packaging.yml
```diff
@@ -601,45 +604,75 @@ jobs:
...
-        with:
-          path: src
-          ref: ${{ github.event.inputs.commitIdToPackage }}
+        fi
+    - name: Setup python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.7'
+    - name: Add msbuild to PATH (windows)
+      if: startsWith(matrix.os, 'windows')
```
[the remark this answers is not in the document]
anonymous-akorn: Good call: runner.os is a far better source of truth for this.

scripts/gha/build_testapps.py
```diff
@@ -362,31 +362,45 @@ def _validate_android_environment_variables():
...
+  # for these, in this order, and set the others to the first found.
+  # If none are set, we check the default location for the ndk.
+  ndk_path = None
+  ndk_vars = [_NDK_ROOT, _ANDROID_NDK_HOME]
+  for env_var in ndk_vars:
+    val = os.environ.get(env_var)
+    if val:
+      ndk_path = val
+      break
+  if not ndk_path:
```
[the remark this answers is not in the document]
anonymous-akorn: The pattern would definitely fit here, but I think it's a bit too idiomatic: I think there's a good chance someone reading the code would go "what the heck does this dangling 'else' do, surely it's a syntax error?"

scripts/gha/build_testapps.py
anonymous-akorn: Added one-liner docstrings.

scripts/gha/build_testapps.py
anonymous-akorn: Added a comment "firebase_analytics.framework -> firebase_analytics", let me know if that's not what you had in mind.

scripts/gha/build_testapps.py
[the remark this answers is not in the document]
vimanyu: Yeah, thats the tricky part about idiomatic constructs like these.

vimanyu (approved):
LGTM

PR #183, opened 2020-11-11
Add hash for generated C++ SDK package so we can verify the file later

jonsimantov:
This adds a file hash for the firebase_cpp_sdk.zip package, which will allow us to verify the generated package later (i.e. when it is put up for download):
- Log a SHA256 hash of the generated firebase_cpp_sdk.zip package, and add the hash as a downloadable artifact as well.
- Add a file to the generated package with a hash for every file contained in the package. In the integration test step, code has been added to verify every file's hash if this file exists and error out if any don't match.

Also, if the workflow is run with certain inputs to customize its behavior, this is now logged:
- If a run is made against a specific commit ID, log the commit ID.
- If a run is just to test an existing packaged SDK (public or from a previous run), log its source.
- If a run is set to preserve intermediate artifacts, log it.

*Important:* This PR changes the final C++ SDK package artifact from **firebase_cpp_sdk** to **firebase_cpp_sdk.zip**, so that the hash can be generated. Any subsequent step that uses that SDK package artifact will have to change the artifact name it downloads to firebase_cpp_sdk.zip and unzip it first before use.

Also note that because GitHub Actions does not yet support "::notice", we are logging these as "::warning" messages for now. This will be changed in the future when possible.

### DIFF
--- a/.github/workflows/cpp-packaging.yml
+++ b/.github/workflows/cpp-packaging.yml
@@ -414,6 +414,7 @@ jobs:
   download_sdk_package:
     name: download-sdk-package
     runs-on: ubuntu-latest
+    needs: [log_inputs]
     if: ${{ github.event.inputs.downloadPublicVersion != '' || github.event.inputs.downloadPreviousRun != '' }}
     steps:
       - name: fetch artifact from previous run
@@ -461,7 +462,7 @@ jobs:
     name: final-merge-packages
     runs-on: ubuntu-latest
     if: ${{ github.event.inputs.downloadPublicVersion == '' && github.event.inputs.downloadPreviousRun == '' }}
-    needs: [build_and_package_ios, build_and_package_android, package_desktop]
+    needs: [build_and_package_ios, build_and_package_android, package_desktop, log_inputs]
     steps:
       - name: fetch SDK
         uses: actions/checkout@v2.3.1
--- a/.github/workflows/cpp-packaging.yml
+++ b/.github/workflows/cpp-packaging.yml
@@ -29,9 +29,9 @@ jobs:
       - name: log run inputs
         run: |
           if [[ -n "${{ github.event.inputs.downloadPublicVersion }}" ]]; then
-            echo "::warning ::Using public SDK package at https://dl.google.com/firebase/sdk/cpp/firebase_cpp_sdk_${{ github.event.inputs.downloadPublicVersion }}.zip"
+            echo "::warning ::Downloading public SDK package from https://dl.google.com/firebase/sdk/cpp/firebase_cpp_sdk_${{ github.event.inputs.downloadPublicVersion }}.zip"
           elif [[ -n "${{ github.event.inputs.downloadPreviousRun }}" ]]; then
-            echo "::warning ::Using SDK package from previous run at https://github.com/firebase/firebase-cpp-sdk/actions/runs/${{ github.event.inputs.downloadPreviousRun }}"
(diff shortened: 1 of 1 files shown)


jonsimantov:
A couple sample runs to see how the logs come out:
https://github.com/firebase/firebase-cpp-sdk/actions/runs/358363311
https://github.com/firebase/firebase-cpp-sdk/actions/runs/358362746

.github/workflows/cpp-packaging.yml
```diff
@@ -417,19 +441,28 @@ jobs:
             sleep 300
           done
           set -e
-          cd firebase-cpp-sdk-final
-          unzip ../firebase_cpp_sdk.zip
-      - name: upload artifacts
+
+      - name: compute SDK hash
+        run: |
```
vimanyu, line 446: Should we force the shell to "bash" here too since we are redirecting to a file since PowerShell redirection's default encoding is UTF-16 not UTF-8?

.github/workflows/cpp-packaging.yml
```diff
@@ -380,7 +405,7 @@ jobs:
         shell: bash
         run: |
           find firebase-cpp-sdk-*-package -type f
-      - name: upload artifacts
+      - name: upload SDK zip
```
vimanyu, line 408: nit: Minor note on formatting. There is an empty line between steps in some places and not in others.
I personally feel its easier to read with an empty line in between but dont mind either format.

vimanyu (commented):
Looks great. Added some minor comments.

.github/workflows/cpp-packaging.yml
jonsimantov: It shouldn't matter as this step only runs on Ubuntu, but I forced it nonetheless.

.github/workflows/cpp-packaging.yml
jonsimantov: Done, I fixed all the steps throughout the file to have blank lines.

vimanyu (approved):
LGTM!

vimanyu (approved):

.github/workflows/cpp-packaging.yml
```diff
@@ -498,14 +576,33 @@ jobs:
...
+      - name: download hash
+        uses: actions/download-artifact@v2
+        with:
+          name: firebase_cpp_sdk_hash.txt
+          path: .
+
       - name: List binary SDK files.
         run: |
+          # Verify zipfile hash first.
+          ${{ env.hashCommand }} -c --quiet firebase_cpp_sdk_hash.txt
```
anonymous-akorn, line 594: FYI Github does not install sha256sum on MacOS, since MacOS ships with shasum. So I won't be able to include these hash-related commands in the test workflows (which run on all three OSes) as is. Though since they only log a warning, I'm assuming they're not essential.

anonymous-akorn (approved):

.github/workflows/cpp-packaging.yml
jonsimantov: Yeah, we can get rid of that check from the general testing workflow, and move it to a standalone hash verification step that just runs on ubuntu, separately from everything else.

If the sha256sum command fails it will actually error (due to the command returning an error), though of the file_hashes.txt file is missing then *that* just a warning (since that file was not shipped in our most recent public release).

jonsimantov: Oh also, I believe shasum can be used to verify hashes created by sha256sum so I can experiment with that if you think we should keep this check. (It still won't do what we want on Windows, I think, though.)

PR #184, opened 2020-11-12
Run complete C++ SDK packaging step whenever a push to "dev" occurs.

jonsimantov:
Whenever a feature branch is merged into "dev", this will kick off a full C++ packaging workflow on that push. (All of the inputs are blank strings in this case, so ensure that blank inputs are treated as default.)

This ensures that we always have an up-to-date package for the latest 'dev' branch, and is an alternative to setting up a cron workflow (as cron will only work once we move back to the 'master' branch).


.github/workflows/cpp-packaging.yml
```diff
@@ -536,7 +540,7 @@ jobs:
     name: cleanup-artifacts
     runs-on: ubuntu-latest
     needs: [merge_packages]
-    if: ${{ github.event.inputs.preserveIntermediateArtifacts == 0 && github.event.inputs.downloadPublicVersion == '' && github.event.inputs.downloadPreviousRun == '' }}
+    if: ${{ (github.event.inputs.preserveIntermediateArtifacts == 0 || github.event.inputs.preserveIntermediateArtifacts == '') && github.event.inputs.downloadPublicVersion == '' && github.event.inputs.downloadPreviousRun == '' }}
```
DellaBitta, line 543: nit:  the test on line 47 uses "0" but here it's just a test against 0.  Can this be consistent?

DellaBitta (approved):
LGTM with a nit.

.github/workflows/cpp-packaging.yml
jonsimantov: Updated the log step around line 47 to use this exact conditional (well, inverted).

DellaBitta (approved):

PR #187, opened 2020-11-13
Fix final package to be usable on Mac

jonsimantov:
Sets Xcode version to 11.7.

Also fixes include path in the final package, which is mistakenly include/include/firebase. This PR corrects it to include/firebase.


vimanyu (dismissed):
LGTM!

jonsimantov:
> LGTM!

Modified this PR to also set the Xcode version since I couldn't use the built SDK with Xcode 11.7 (which is our minimum requirement), PTAL.

vimanyu (approved):

PR #188, opened 2020-11-16
Feature/python tool build apps with firebase

vimanyu:
Standalone python tool with no external dependencies to build apps against the prebuilt firebase libraries or C++ source. 
Simplifies building external apps against the firebase SDK. Reduces the complexity and number of steps required to a single python command. Supports both building with prebuilt libraries and source SDK. 

Tested this with quickstart-cpp examples. 

Example usage:
Let's say we want to build the quickstart cpp example for firebase database.
As specified above, there are 2 options - build against the firebase source or
prebuilt firebase libraries.
```
# Build against the firebase cpp sdk source
python3 scripts/build_desktop_app_with_firebase.py --app_dir ~/quickstart-cpp/database/testapp
                                                   --sdk_dir . --build_dir build_source
(or)
# Build against the prebuilt released firebase libraries
python3 scripts/build_desktop_app_with_firebase.py --app_dir ~/quickstart-cpp/database/testapp
                                                   --sdk_dir ~/prebuilt/firebase_cpp_sdk_6.15.1/
                                                   --build_dir build_prebuilt
```


scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+
+def get_vcpkg_triplet(arch, msvc_runtime_library='static'):
+  """ Get vcpkg target triplet (platform definition).
+
+  Args:
+    arch (str): Architecture (eg: 'x86', 'x64').
+    msvc_runtime_library (str): Runtime library for MSVC (eg: 'static', 'dynamic').
+
+  Raises:
+    ValueError: If current OS is not win,mac or linux.
```
DellaBitta, line 88: nit:  space between comma and mac.

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+
+def build_source_vcpkg_dependencies(sdk_source_dir, arch, msvc_runtime_library):
+  """Build C++ dependencies for firebase source SDK using vcpkg.
+
+  Args:
+   sdk_source_dir (str): Path to Firebase C++ source directory.
+   arch (str): Platform Architecture (example: 'x64').
+   msvc_runtime_library (str): Runtime library for MSVC (eg: 'static', 'dynamic').
+  """
+  # TODO: Remove this once dev branch of firebase-cpp-sdk repo has been merged
```
DellaBitta, line 120: Recommend creating a ticket on the backlog  and referencing it here.

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+  subprocess.run([python_exe, "scripts/gha/install_prereqs_desktop.py"],
+                 cwd=sdk_source_dir, check=True)
+  subprocess.run([python_exe, "scripts/gha/build_desktop.py", "--arch", arch,
+                              "--msvc_runtime_library", msvc_runtime_library,
+                              "--vcpkg_step_only"], cwd=sdk_source_dir, check=True)
+
+def build_app_with_source(app_dir, sdk_source_dir, build_dir, arch,
+                          msvc_runtime_library='static', config=None,
+                          target_format=None):
+  """Build desktop app directly against the firebase C++ SDK source.
```
DellaBitta, line 136: Nit: capitalize Firebase here.

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+
+  Since this invovles a cmake configure, it is advised to run this on a clean
+  build directory.
```
DellaBitta, line 139: Should a note about this be made at the top of the file / usage printouts?

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+  subprocess.run([python_exe, "scripts/gha/build_desktop.py", "--arch", arch,
+                              "--msvc_runtime_library", msvc_runtime_library,
+                              "--vcpkg_step_only"], cwd=sdk_source_dir, check=True)
+
+def build_app_with_source(app_dir, sdk_source_dir, build_dir, arch,
+                          msvc_runtime_library='static', config=None,
+                          target_format=None):
+  """Build desktop app directly against the firebase C++ SDK source.
+
+  Since this invovles a cmake configure, it is advised to run this on a clean
```
DellaBitta, line 138: Nit: "involves"

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+   config (str): Release/Debug config.
+          If its not specified, cmake's default is used (most likely Debug).
+   target_format (str): If specified, build for this targetformat ('frameworks' or 'libraries').
+  """
+  # Cmake configure
+  cmd = ['cmake', '-S', '.', '-B', build_dir]
+  cmd.append('-DFIREBASE_CPP_SDK_DIR={0}'.format(sdk_source_dir))
+
+  # If generator is not specifed, default for platform is used by cmake, else
+  # use the specified value
```
DellaBitta, line 156: Nit:  [The Google style guide ](https://google.github.io/styleguide/pyguide.html#386-punctuation-spelling-and-grammar)says that all comments should include punctuation.

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+    # If this option is not specified, the default value is /MT (static).
+    if msvc_runtime_library == "static":
+      cmd.append('-DMSVC_RUNTIME_LIBRARY_STATIC=ON')
+
+  if (target_format):
+    cmd.append('-DFIREBASE_XCODE_TARGET_FORMAT={0}'.format(target_format))
+  print("Running {0}".format(' '.join(cmd)))
+  subprocess.run(cmd, cwd=app_dir, check=True)
+
+  #CMake build
```
DellaBitta, line 193: Nit: A space between # and CMake

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+  cmd = ['cmake', '--build', build_dir, '-j', str(os.cpu_count()), '--config', config]
+  print("Running {0}".format(' '.join(cmd)))
+  subprocess.run(cmd, cwd=app_dir, check=True)
+
+def build_app_with_prebuilt(app_dir, sdk_prebuilt_dir, build_dir, arch,
+                            msvc_runtime_library='static', config=None):
+  """Build desktop app directly against the prebuilt firebase C++ libraries.
+
+  Since this invovles a cmake configure, it is advised to run this on a clean
```
DellaBitta, line 202: nit: invovles

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+  build directory.
```
DellaBitta, line 203: Again, should we note this at the top of the file?

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+
+  Args:
+   app_dir (str): Path to directory containing application's CMakeLists.txt.
+   sdk_prebuilt_dir (str): Path to prebuilt firebase C++ libraries.
+   build_dir (str): Output build directory.
+   arch (str): Platform Architecture (example: 'x64').
+   msvc_runtime_library (str): Runtime library for MSVC (eg: 'static', 'dynamic').
+   config (str): Release/Debug config.
```
DellaBitta, line 211: Might be worth specifying examples here.  Like this should be either the string "Release" or "Debug".

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+           "Please check the argument to '--sdk_dir'")
+    sys.exit(1)
+
+  if not is_path_valid_for_cmake(args.app_dir):
+    print ("App path provided is not valid. Could not find a CMakeLists.txt at the root level.\n"
+           "Please check the argument to '--app_dir'")
+    sys.exit(1)
+
+  if is_sdk_path_source(args.sdk_dir):
+    print("SDK path provided is Firebase C++ source directory. Building...")
```
DellaBitta, line 251: Nit: "is a Firebase C++ source directory"

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+    build_app_with_source(args.app_dir, args.sdk_dir, args.build_dir, args.arch,
+                          args.msvc_runtime_library, args.config, args.target_format)
+  else:
+    validate_prebuilt_args(args.arch, args.config)
+    print("SDK path provided is Firebase C++ prebuilt libraries. Building...")
+    build_app_with_prebuilt(args.app_dir, args.sdk_dir, args.build_dir, args.arch,
+                            args.msvc_runtime_library, args.config)
+
+  print ("Build successful!\n"
+         "Please find your executables in build directory: {0}".format
```
DellaBitta, line 262: the build directory (?)

scripts/gha/build_desktop.py
```diff
@@ -182,7 +182,7 @@ def cmake_configure(build_dir, arch, msvc_runtime_library='static',
 
     # Use our special cmake option for /MD (dynamic).
     # If this option is not specified, the default value is /MT (static).
-    if msvc_runtime_library == "dynamic":
+    if msvc_runtime_library == "static":
```
DellaBitta, line 185: oh my!  👍

DellaBitta (changes requested):
Some nits on formatting and documentation.  Since Python isn't my forte, I hope someone else will also review this.

Really looking forward to getting this into the repo so that I can use it!

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+    if arch == 'x86' or config == 'Debug':
+      raise ValueError("Prebuilt mac firebase libraries are built for x64 and Release mode only. "
+                       "Please fix the command line arguments and try again")
+
+  if platform.system() == 'Linux':
+    if config == 'Debug':
+      raise ValueError("Prebuilt linux firebase libraries are built with Release mode only. "
+                       "Please fix the --config command line argument and try again.")
+
+def get_vcpkg_triplet(arch, msvc_runtime_library='static'):
```
anonymous-akorn, line 80: This logic (and some others) already exist in the SDK repo, don't they? Should we find a way to avoid duplication?

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
...
+   arch (str): Platform Architecture (example: 'x64').
+   msvc_runtime_library (str): Runtime library for MSVC (eg: 'static', 'dynamic').
+  """
+  # TODO: Remove this once dev branch of firebase-cpp-sdk repo has been merged
+  # onto main branch. This is required because vcpkg lives only in dev branch currently.
+  subprocess.run(['git', 'checkout', 'dev'],
+                 cwd=sdk_source_dir, check=True)
+  subprocess.run(['git', 'pull'], cwd=sdk_source_dir, check=True)
+
+  python_exe = 'python3' if distutils.spawn.find_executable('python3') else 'python'
```
anonymous-akorn, line 126: I think it would be preferable to use the same executable that was used to run this script, which can be accessed with sys.executable

scripts/build_desktop_app_with_firebase.py
```diff
@@ -0,0 +1,281 @@
+#!/usr/bin/env python
```
anonymous-akorn, line 1: glint offers about a hundred complaints about formatting on this file. Mostly lines too long and inconsistent string delimiters (single vs double quotes). Copying the file into cider is probably the easiest way to see them.

anonymous-akorn (commented):
I wonder if it would be better to modify the integration test builder (build_testapps.py) to be able to handle these projects. AFAIK their build process is almost identical - the integration tests depend on gtest, and are in a different location, but I think that's about it. Or alternatively to think about how to avoid having separate projects altogether (like Unity, which extends the sample project with integration tests), though that's perhaps overly ambitious at this point. 

Though build_testapps.py does not support nearly as many options at the moment (x64 vs x86, release vs debug, etc.)

scripts/build_desktop_app_with_firebase.py
vimanyu: Done

scripts/build_desktop_app_with_firebase.py
vimanyu: Great idea. Done!

scripts/build_desktop_app_with_firebase.py
vimanyu: Done

scripts/build_desktop_app_with_firebase.py
vimanyu: Yup. Makes sense. Easier for users to read this with all the examples at the top. Done.

scripts/build_desktop_app_with_firebase.py
DellaBitta: TIL'ed...

scripts/build_desktop_app_with_firebase.py
vimanyu: Fixed.

scripts/build_desktop_app_with_firebase.py
vimanyu: Done

scripts/build_desktop_app_with_firebase.py
vimanyu: Done

scripts/build_desktop_app_with_firebase.py
vimanyu: Done.

scripts/build_desktop_app_with_firebase.py
vimanyu: Done

scripts/gha/build_desktop.py
vimanyu: haha! Also, removed the confusing comment on the top.

scripts/build_desktop_app_with_firebase.py
vimanyu: Done.

scripts/build_desktop_app_with_firebase.py
vimanyu: Good to know. Thanks for pointing this out. Done.

scripts/build_desktop_app_with_firebase.py
vimanyu: Done.

scripts/build_desktop_app_with_firebase.py
vimanyu: Good point. We should at some point consolidate these scripts but I feel the main goal here was to provide a super simple self sufficient single script for our users to understand how to build things with vcpkg. If the script does too many other things, the logic might not be that straightforward to follow. 
I wouldn't want to add integration test specific command line arguments (eg: api) to this script (just to keep this one generic that users can place anywhere) but we could call this script from integration scripts as that seems to be a more special case.

scripts/build_desktop_app_with_firebase.py
vimanyu: Nice! Much cleaner.

scripts/build_desktop_app_with_firebase.py
vimanyu: Thanks. Super super helpful! Will be doing this for any python development in the future. I wonder if vscode has something similar to this.

scripts/build_desktop_app_with_firebase.py
vimanyu: Marking this as resolved for the time being but we should definitely make a note of this.

scripts/build_desktop_app_with_firebase.py
anonymous-akorn: https://g3doc.corp.google.com/devtools/gpylint/g3doc/editor_integration.md?cl=head#vscode

Appears to be the case.

anonymous-akorn (approved):
Looks good to me, though we should be very conscious of the implied maintenance burden of this kind of thing, especially where there is duplicated logic.

DellaBitta (approved):

PR #193, opened 2020-11-25
Deduplicate vcpkg logic in testapp builder

anonymous-akorn:
The testapp builder duplicated some vcpkg logic: this removes that duplication, instead relying on build_desktop.py's "vcpkg_step_only" mode, and some logic from utils.py

This fixes a break caused by a change in the structure of the vcpkg response files which was not reflected in this duplicated logic, and will prevent similar breakages in the future.


vimanyu (dismissed):
LGTM

vimanyu (approved):

PR #194, opened 2020-11-25
Fix gCloud steps in integration tests on Windows

anonymous-akorn:
The setting of these env vars was recently updated to use the new syntax. This does not work as-is on powershell due to encoding, as documented here:
https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#environment-files

As a result, the gCloud tools were not working on Windows as an env var needed to work around a known gCloud bug was not being set.

To fix this, we set the env var using bash.


vimanyu (approved):
LGTM!

PR #200, opened 2020-12-01
Added more test phone numbers.

DellaBitta:
The use of a single phone number caused issues when multiple clients were running in the integration test CI framework. causing the PhoneAuth attempts to globber one another.  

Added 9 more phone numbers, and made the choice of the phone number to be random based on the time.


anonymous-akorn:
What exactly was the underlying issue? Was it that multiple concurrent instances of the integration test were making a request with this phone number and this caused problems?

DellaBitta:
> What exactly was the underlying issue? Was it that multiple concurrent instances of the integration test were making a request with this phone number and this caused problems?

Yes, exactly that.

anonymous-akorn (approved):

vimanyu (approved):

PR #201, opened 2020-12-02
Remove several Windows test jobs from pull requests

anonymous-akorn:
We're experiencing queuing on GHA from trying to use too many machine resources. A disproportionate chunk of this is coming from Windows and its excessive build permutations.

Android on Windows has been broken for a while - temporarily removed until fixed.

For desktop, remove 6 of the permutations, leaving the two required ones: Release-x64-static and Debug-x86-dynamic.

### DIFF
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -22,16 +22,19 @@ jobs:
         architecture: ["x64", "x86"]
         python_version: [3.7]
         include:
+        # Add these Windows jobs.
         - os: windows-latest
           build_type: "Release"
           architecture: "x64"
           msvc_runtime: "static"
-          vcpkg_triplet: "windows-static"
+          vcpkg_triplet_suffix: "windows-static"
         - os: windows-latest
           build_type: "Debug"
           architecture: "x86"
           msvc_runtime: "dynamic"
-          vcpkg_triplet: "windows-static-md"
+          vcpkg_triplet_suffix: "windows-static-md"
+
+        # Specify additional parameters for these jobs.
         - os: ubuntu-latest
           msvc_runtime: "static"
           vcpkg_triplet_suffix: "linux"
@@ -39,6 +42,7 @@ jobs:
           msvc_runtime: "static"
           vcpkg_triplet_suffix: "osx"
 
+        # Don't build for x86 on MacOS.
         exclude:
         - os: macos-latest
           architecture: "x86"
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -17,32 +17,30 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest]
         build_type: ["Release", "Debug"]
         architecture: ["x64", "x86"]
-        msvc_runtime: ["static"]
         python_version: [3.7]
         include:
         - os: windows-latest
-          vcpkg_triplet_suffix: "windows-static"
+          build_type: "Release"
+          architecture: "x64"
+          msvc_runtime: "static"
+          vcpkg_triplet: "windows-static"
+        - os: windows-latest
+          build_type: "Debug"
+          architecture: "x86"
+          msvc_runtime: "dynamic"
+          vcpkg_triplet: "windows-static-md"
         - os: ubuntu-latest
+          msvc_runtime: "static"
           vcpkg_triplet_suffix: "linux"
         - os: macos-latest
+          msvc_runtime: "static"
           vcpkg_triplet_suffix: "osx"
 
         exclude:
         - os: macos-latest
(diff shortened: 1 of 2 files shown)


vimanyu:
This makes sense but I was wondering if there is any merit in also preserving the exhaustive list for "non-PR" cases? With some tricks, we could maintain the same list of configurations in desktop.yml and modify the matrix dependent on if its a PR event or not. This way, we trim the configurations for PR (or any other events) and also have an exhaustive list of configurations just incase someone wants to run them all.

Pros:
- no need to maintain a separate workflow for exhaustive list which could be useful to catch certain edge cases (for example, unit tests erroring out in Debug and not Release because of ASSERT macros).
- no need to modify the configurations in feature development branch (honestly not that big a deal because I do modify the workflow anyway to add a "push" trigger for my branch testing but this means adding more things).

Cons:
- makes the workflow more complicated and harder to follow

It probably makes more sense to keep it simple for now maintaining a trimmed list and let developers add things if they need to.

anonymous-akorn:
> This makes sense but I was wondering if there is any merit in also preserving the exhaustive list for "non-PR" cases? With some tricks, we could maintain the same list of configurations in desktop.yml and modify the matrix dependent on if its a PR event or not. This way, we trim the configurations for PR (or any other events) and also have an exhaustive list of configurations just incase someone wants to run them all.
> 
> Pros:
> 
> * no need to maintain a separate workflow for exhaustive list which could be useful to catch certain edge cases (for example, unit tests erroring out in Debug and not Release because of ASSERT macros).
> * no need to modify the configurations in feature development branch (honestly not that big a deal because I do modify the workflow anyway to add a "push" trigger for my branch testing but this means adding more things).
> 
> Cons:
> 
> * makes the workflow more complicated and harder to follow
> * no clean way of doing dynamic matrix in github actions
> 
> It probably makes more sense to keep it simple for now maintaining a trimmed list and let developers add things if they need to.

I do think we should avoid duplicating the workflow just to have different build parameters. The matrix can be populated dynamically from arbitrary JSON, which means one way to switch between matrices with complex differences is to store distinct matrices as JSON and have a simple switch in a job that runs before the rest. I did something not too dissimilar, albeit in a smaller way, in the integration test workflow to modify input parameters before storing them as matrix parameters.

On the flip side, I wonder how much value the other permutations really give us here. The two cases in this PR hit every build parameter, just not in every permutation. Do we know of concrete cases where a change could work in these two cases, but not in one of the other 6 permutations? 

Regardless, my opinion is that we should submit this change as-is to alleviate the queueing. If we want to restore the ability to hit the other permutations in a non-PR capacity, that should be addressed in a separate PR as the change will be non-trivial.

vimanyu:
I agree. Lets push this out for now and we can add the dropped permutations later if we need to.

.github/workflows/desktop.yml
```diff
@@ -17,30 +17,31 @@ jobs:
...
         architecture: ["x64", "x86"]
-        msvc_runtime: ["static", "dynamic"]
         python_version: [3.7]
         include:
         - os: windows-latest
-          vcpkg_triplet_suffix: "windows-static"
+          build_type: "Release"
+          architecture: "x64"
+          msvc_runtime: "static"
+          vcpkg_triplet: "windows-static"
```
vimanyu, line 29: I think you meant vcpkg_triplet_suffix here.

.github/workflows/desktop.yml
```diff
@@ -17,30 +17,31 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [windows-latest, ubuntu-latest, macos-latest]
+        os: [ubuntu-latest, macos-latest]
```
vimanyu, line 20: Removing windows-latest from here makes the workflow slightly less readable as we have to pay attention to the special include cases.

vimanyu (changes requested):

.github/workflows/desktop.yml
anonymous-akorn: Either special include cases, or special exclude cases. This seemed like the lesser of two evils: we'd have to include 3 exclude statements for windows, each of which hits two permutations. This way we also get to remove two additional exclude statements for dynamic msvc runtime with macos and linux. 

I was having difficulties figuring out which jobs this config queued, or how many.

.github/workflows/desktop.yml
anonymous-akorn: Whoops, good catch.

.github/workflows/desktop.yml
anonymous-akorn: I added some brief comments to the include/exclude statements to make it easier their purposes, especially to disambiguate between includes for the sake of adding jobs versus includes for the sake of adding variables to existing jobs.

vimanyu: Perfect. That should help. Thanks!

vimanyu (approved):

PR #205, opened 2020-12-07
Use an older version of binutils on Mac

jonsimantov:
Using the latest binutils causes blastdoor to create invalid Mac libraries that crash the linker. Until this is fixed in binutils, use an older version that is known to work.

Because this older version does not support thread-local storage on Mac, disable thread-local storage in our usage of websockets via a define at build time. This does not cause a problem because we only use websockets via the Firebase scheduler, which runs all websockets requests in a single worker thread.

### DIFF
--- a/.github/workflows/cpp-packaging.yml
+++ b/.github/workflows/cpp-packaging.yml
@@ -21,10 +21,6 @@ on:
 
 env:
   # Packaging prerequisites
-  # Binutils 2.35.1 released Sep 19, 2020
-  binutilsVerLinux: "2.35.1"
-  # On Mac, use an older Binutils to avoid corrupted libraries.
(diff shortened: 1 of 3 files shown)


.github/workflows/cpp-packaging.yml
```diff
@@ -67,8 +65,12 @@ jobs:
         include:
         - os: ubuntu-latest
           tools_platform: linux
+          # Binutils 2.35.1 released Sep 19, 2020
+          binutils_version: "2.35.1"
```
jonsimantov, line 69: Note: I've learned you can't use env substitution in matrix definitions, so I had to move the version numbers directly into here.

a-maurice (approved):

a-maurice (approved):

PR #207, opened 2020-12-08
Allow all dependencies (curl, websockets, grpc) to use boringssl instead of openssl when making packaged C++ SDK.

jonsimantov:
This PR gives you the option to switch all desktop platforms over to using boringssl rather than openssl, which does not like being embedded in our SDKs. If you are building a binary SDK, you may want to enable this option. If you are linking directly to our SDK from source via cmake, you probably don't need it (but you should ensure that OpenSSL is installed on your system.

It will be enabled if you turn on FIREBASE_USE_BORINGSSL at cmake configure time, e.g.:
**cmake _[source directory]_ -DFIREBASE_USE_BORINGSSL=ON**

With this option enabled, it will build boringssl during the configure step by shelling out to another instance of cmake. Much work is done in build_external_dependencies to ensure that this (and future) sub-builds correctly match the main project's selection of:
- CPU architecture
- build type (debug/release)
- Windows MSVC runtime
- Linux legacy/c++11 ABI

A few notes on switching to boringSSL:
- We manually set all of the OPENSSL_* flags that CMake's find_package(OpenSSL) sets, so that any subprojects using openssl will get the include files and libraries of boringssl instead.
- We have updated to the latest libcurl which supports boringssl properly.
- OpenSSL has been removed from the vcpkg files.
- Golang has been added as a prerequisite as it's required for boringssl's build. Github runners already have it installed, but I've added it to the prereqs script anyway.

Additional changes in this PR:
- The Linux build has been switched to use the legacy ABI rather than the C++11 ABI. An upcoming PR will enable both ABIs.
- Binutils on Mac is restored to the latest version, as openssl was the cause of the library corruption we were seeing.
- Some missing Firestore core headers are now added to the final C++ SDK package.
- Integration test github action now builds and runs both openssl and boringssl versions.

You'll also see a few workaround at the end of the download_external_sources function, this is to work around a few issues in subprojects that expect openssl but are now getting boringssl. There are PRs to fix these in the upstream sources, but until we update to their latest version these will need to be there.


jonsimantov:
Tests are now passing and this should be ready for review.

CMakeLists.txt
```diff
@@ -115,6 +115,13 @@ else()
   set(DESKTOP OFF)
 endif()
 
+if(DESKTOP AND NOT MSVC AND NOT APPLE)
+  # Linux-specific option.
+  add_definitions(-D_GLIBCXX_USE_CXX11_ABI=0)
```
vimanyu, line 120: Possible to collect this setting in a variable and use that across the board?
Will keep it more maintainable for future edits.

CMakeLists.txt
```diff
@@ -174,6 +181,72 @@ else()
   set(FIRESTORE_SOURCE_DIR ${FIREBASE_BINARY_DIR}/external/src/firestore)
 endif()
 
+if(DESKTOP)
+  # Use the static versions of the OpenSSL libraries.
+  set(OPENSSL_USE_STATIC_LIBS TRUE)
+  if (MSVC)
+    # Get the correct version of the OpenSSL libraries based on building for MT.
+    if ("${CMAKE_CXX_FLAGS_DEBUG}" MATCHES "/MT" OR
```
vimanyu, line 189: Any downside of being more specific for debug flag ("/MTd")?

CMakeLists.txt
```diff
@@ -174,6 +181,72 @@ else()
...
+
+  # Use BoringSSL instead of OpenSSL.
+  set(BORINGSSL_ROOT_DIR ${CMAKE_BINARY_DIR}/external/src/boringssl/src CACHE STRING "" FORCE)
+  set(BORINGSSL_BINARY_DIR ${CMAKE_BINARY_DIR}/external/src/boringssl-build CACHE STRING "" FORCE)
+  set(OPENSSL_ROOT_DIR ${BORINGSSL_ROOT_DIR} CACHE STRING "" FORCE)
+
+  # The call below to build_external_dependencies will make sure that these
+  # libraries exist before the libraries are imported via add_library.
+  if (MSVC)
+    set(OPENSSL_SSL_LIBRARY ${BORINGSSL_BINARY_DIR}/ssl/${CMAKE_BUILD_TYPE}/ssl.lib CACHE FILEPATH "" FORCE)
```
vimanyu, line 205: Have we tested what happens if CMAKE_BUILD_TYPE is not specified?
Like, if we run 
`cmake ..` and not `cmake .. -DCMAKE_BUILD_TYPE=(Release|Debug)`

vimanyu: If cmake doesn't set this variable, maybe we should set it to default value ourselves to ensure it always has a value.

CMakeLists.txt
```diff
@@ -174,6 +181,72 @@ else()
...
+    set(OPENSSL_CRYPTO_LIBRARY ${BORINGSSL_BINARY_DIR}/crypto/${CMAKE_BUILD_TYPE}/crypto.lib CACHE FILEPATH "" FORCE)
+  else()
+    set(OPENSSL_SSL_LIBRARY ${BORINGSSL_BINARY_DIR}/ssl/libssl.a CACHE FILEPATH "" FORCE)
+    set(OPENSSL_CRYPTO_LIBRARY ${BORINGSSL_BINARY_DIR}/crypto/libcrypto.a CACHE FILEPATH "" FORCE)
+  endif()
+endif()
+
+if(NOT FIREBASE_CPP_USE_PRIOR_GRADLE_BUILD)
+  message(STATUS "Building external project dependencies...")
```
vimanyu, line 214: nit: Something that wasn't immediately clear was why we have this condition for building external dependencies. Also, isn't boringssl required only for desktop builds? Maybe a one line comment explaining the case might help.

CMakeLists.txt
```diff
@@ -223,37 +296,27 @@ endif()
...
-    else()
-      set(OPENSSL_MSVC_STATIC_RT FALSE)
-    endif()
-  endif()
-
   # Build curl as a static library
   set(CURL_STATICLIB ON CACHE BOOL "")
   if (WIN32)
-    set(CMAKE_USE_WINSSL ON CACHE BOOL "")
+    set(CMAKE_USE_SCHANNEL ON CACHE BOOL "")
```
vimanyu, line 302: nit: will be great to put a one line comment explaining what this option does,
"enable Windows native SSL/TLS"

CMakeLists.txt
```diff
@@ -304,7 +367,8 @@ if(DESKTOP)
     )
     target_link_libraries(libuWS
       PRIVATE
-        ${OPENSSL_LIBRARIES}
+        OpenSSL::SSL
```
vimanyu, line 370: Since ssl->crypto dependency is captured already in ssl, do we need to specify these 2 libraries separately here?

cmake/external/boringssl.cmake
```diff
@@ -0,0 +1,39 @@
+# Copyright 2018 Google LLC
```
vimanyu, line 1: nit: 2020

cmake/external/boringssl.cmake
```diff
@@ -0,0 +1,39 @@
...
+  DOWNLOAD_NAME boringssl-${commit}.tar.gz
+  URL https://github.com/google/boringssl/archive/${commit}.tar.gz
+
+  PREFIX ${PROJECT_BINARY_DIR}
+
+  CONFIGURE_COMMAND ""
+  BUILD_COMMAND     ""
+  INSTALL_COMMAND   ""
+  TEST_COMMAND      ""
+)
```
vimanyu, line 39: nit:newline at the end

cmake/external_rules.cmake
```diff
@@ -90,6 +92,153 @@ function(download_external_sources)
     OUTPUT_FILE ${PROJECT_BINARY_DIR}/external/output_cmake_build.txt
     WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/external
   )
+
+  if(NOT ANDROID AND NOT IOS)
```
vimanyu, line 96: Is this equivalent to easier to read`if(DESKTOP)`?

cmake/external_rules.cmake
```diff
@@ -90,6 +92,153 @@ function(download_external_sources)
...
+  if(CMAKE_HOST_SYSTEM_NAME STREQUAL "Windows")
+    # Windows doesn't have an 'env' command
+    set(ENV_COMMAND "")
+  else()
+    set(firebase_command_line_path "$ENV{PATH}")
+    set(firebase_command_line_home "$ENV{HOME}")
+    set(ENV_COMMAND env -i PATH=${firebase_command_line_path} HOME=${firebase_command_line_home} )
+  endif()
+
+  message("CMake generator: ${CMAKE_GENERATOR}")
```
vimanyu, line 130: Recommend using STATUS to ensure these go to stdout. Messages without a mode go to stderr.

cmake/external_rules.cmake
```diff
@@ -90,6 +92,153 @@ function(download_external_sources)
...
+      WORKING_DIRECTORY ${PROJECT_BINARY_DIR}/external/src/boringssl-build
+      RESULT_VARIABLE boringssl_configure_status
+    )
+    if (boringssl_configure_status AND NOT boringssl_configure_status EQUAL 0)
+      message(FATAL_ERROR "BoringSSL configure failed: ${boringssl_configure_status}")
+    endif()
+  
+    # Run builds in parallel if we know how
+    if(CMAKE_GENERATOR STREQUAL "Unix Makefiles")
+      set(cmake_build_args -j)
```
vimanyu, line 225: I think this flag works with other generators too. 
https://cmake.org/cmake/help/latest/release/3.12.html#command-line

database/CMakeLists.txt
```diff
@@ -182,9 +182,10 @@ else()
   set(additional_link_LIB
       flatbuffers
       firebase_rest_lib
-      ${OPENSSL_CRYPTO_LIBRARY}
       leveldb
-      libuWS)
+      libuWS
+      OpenSSL::SSL
```
vimanyu, line 187: Same as before. Do we need to state this dependency explicitly?

vimanyu (changes requested):
Looks good for most part. Most comments are nits.

cmake/external_rules.cmake
jonsimantov: I'm just copying what we do in the download function above.

CMakeLists.txt
jonsimantov: Assuming you mean the c++11 option, I have this in my follow up PR to support both, is that ok?

database/CMakeLists.txt
jonsimantov: Without this, the tests were failing to link on Linux.

jonsimantov:
I think may actually gate this boringssl behavior behind a build flag, so it can still use your system's openssl if you are linking from source. WDYT?

CMakeLists.txt
jonsimantov: On Windows it seems to always be set. I'll set it to a default value instead - my research indicates that if it's not set on Windows, it's treated as debug.

CMakeLists.txt
jonsimantov: This was leftover from when this call was right after the download, but I moved the external build into the if(DESKTOP) instead, since we only need it on desktop anyway.

cmake/external_rules.cmake
jonsimantov: I don't know if DESKTOP is necessarily set here since it's in a function, but I'll use the external_platform value set above.

jonsimantov: Oh whoops, that's in a different function. Anyway I'll put a comment in, but because we aren't guaranteed to have DESKTOP set (since it's set in a different cmake file) I'd rather just check this here, as it's safer.

jonsimantov:
Changes made and comments left, PTAL.

jonsimantov:
I went ahead and added a FIREBASE_USE_BORINGSSL option, so that if you link against C++ SDK source it will still use your system's OpenSSL by default. Have tested on Mac and Linux with both configurations.

CMakeLists.txt
vimanyu: Sounds good. Yeah, I see the hardcoded string being duplicated and was wondering if we can use a variable instead. 
Will wait on the next PR.

CMakeLists.txt
vimanyu: Hmm... I think I did a quick test on linux and noticed it wasn't set. I thought cmake lets the generator use whatever their default is. But we are safe for now as we provide this value explicitly in our build steps but something to watch out for. Marking it resolved for now.

cmake/external_rules.cmake
vimanyu: I wonder why we restrict it to only one generator. Anthony might have more input on this but its not a big deal for now.

vimanyu (approved):
LGTM!
Since it is a build system change, we should probably wait for @a-maurice to take a look.

jonsimantov:
> Seems to be causing a couple issues that I will look into before this moves forward.

This is now fixed, you can link the SDK from source either with openssl (default) or specifying FIREBASE_USING_BORINGSSL.

PTAL

vimanyu:
Now that there is option to also use openssl, do we want to re-add openssl to vcpkg's list of packages?
Or we could assume users are trying to use a system level openssl package? 

vcpkg will only make sense if we don't want to use a system specific package (vcpkg packages are localized in a directory and only used for that project) and if we care about the versions of openssl library.

jonsimantov:
I think we should keep it out of vcpkg - again, I do think it might make sense to move off of vcpkg and build all dependencies in the binary directory in the future. Also putting it into vcpkg causes cmake to sometimes think it wants to use that library.

vimanyu:
Cool, sounds good. I agree, it would be awesome to have a self contained build in the future and avoid dealing with external packages.
Just to summarize, if users build our SDK from source by themselves and if they do not set `FIREBASE_USE_BORINGSSL=ON` (default value is OFF), they will have to ensure OpenSSL availability on their system but if they use the `build_desktop.py` script, then they dont have to deal with openssl as the command sets this option before running cmake. 
Approving the PR.

vimanyu (dismissed):
LGTM

app/CMakeLists.txt
```diff
@@ -325,11 +325,13 @@ target_compile_definitions(firebase_app
   PRIVATE
     -DINTERNAL_EXPERIMENTAL=1
 )
+
 # firebase_app has a dependency on flatbuffers, which needs to be included.
 target_link_libraries(firebase_app
   PRIVATE
     flatbuffers
     ${LIBSECRET_LIBRARIES}
+    ${ADDITIONAL_LIBS}
```
a-maurice, line 334: I don't think ADDITIONAL_LIBS is defined in this context, at least I'm not sure what this is doing?

a-maurice (changes requested):

app/CMakeLists.txt
jonsimantov: Ah, this was leftover from when I was trying to fix a few things, will remove.

a-maurice (dismissed):

a-maurice (approved):

PR #208, opened 2020-12-09
Feature/linux x86 dont install if packages exist

vimanyu:
For linux x86 builds, we have to install some required i386 packages before proceeding with the build.
This is an optimization to only run these install commands if those packages dont exist already).
Avoids having the user to type in their password everytime as these commands require admin rights.

### DIFF
--- a/scripts/gha/build_desktop.py
+++ b/scripts/gha/build_desktop.py
@@ -62,21 +62,24 @@ def install_x86_support_libraries():
     packages = ('libglib2.0-dev:i386', 'libsecret-1-dev:i386')
 
     # First check if these packages exist on the machine already
-    import apt
-    cache = apt.Cache()
-    package_installed = []
     for package in packages:
-      if package in cache.keys():
-        package_installed.append(cache[package].is_installed)
-      else:
-        package_installed.append(False)
+      devnull = open(os.devnull, "w")
+      retval = subprocess.call(["dpkg", "-s", package], stdout=devnull, stderr=subprocess.STDOUT)
+      devnull.close()
+      # if package is not installed, dpkg returns a value other than 0 and in this case,
+      # we get out of this loop and install those packages.
+      if retval != 0:
+        break
+    else:
+      # If the for loop iterated without breaking, all required packages are already installed.
+      # Nothing to do. Return from the function.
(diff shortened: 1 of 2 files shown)


vimanyu:
Oh no. Looks like debian python apt package does not exist on github runners. Looking at other ways.

scripts/gha/build_desktop.py
```diff
@@ -55,14 +55,31 @@ def append_line_to_file(path, line):
...
+      # If the for loop iterated without breaking, all required packages are already installed.
+      # Nothing to do. Return from the function.
+      return
+
     utils.run_command(['apt', 'install', 'gcc-multilib', 'g++-multilib'], as_root=True)
     utils.run_command(['dpkg', '--add-architecture', 'i386'], as_root=True)
     utils.run_command(['apt', 'update'], as_root=True)
-    utils.run_command(['apt', 'install', 'libglib2.0-dev:i386'], as_root=True)
-    utils.run_command(['apt', 'install', 'libsecret-1-dev:i386'], as_root=True)
+    for package in packages:
```
jonsimantov, line 81: Bit of an optimization, both dpkg and apt install can take multiple package arguments, so you can get rid of BOTH for loops and just do subprocess.call(['dpkg', '-s'] + packages, ...) and utils.run_command(['apt', 'install'] + packages).

The dpkg will return 1 if any of the packages you specified are not installed, so that should work fine for your purposes.

jonsimantov: Also it's probably worth querying gcc-multilib in the list too, no?

vimanyu: Great tip! Didn't know there was such a convenient way to find a missing package with one command. 
Made the change and am also checking for *-multilib packages now.

jonsimantov (approved):
LGTM as long as it's okay that gcc-multilib and g++multilib are now installed before the i386 architecture is added.

vimanyu:
I think you meant that we are installing them after i386 architecture is added, which is different from what we did earlier.
I feel it should be okay as these packages dont have an architecture specific version and the tests passed on the runners. My guess is its a safe change but will keep an eye out for this.

PR #211, opened 2020-12-14
Change integration test log level to "debug"

anonymous-akorn:
Currently the logging behaviour for "debug" level logs is set to show them at the end of the integration test, if there were any failures. If the integration test prematurely ends (crash, failed assertion, or timeout), this results in them not showing up at all. In practice, this often leads to a lack of relevant logs to understand why a test didn't complete.

To fix this, the overall log level is changed from info to debug. This ensures debug logs are logged immediately, so a prematurely ended test won't prevent them from appearing.

This includes a fix to the desktop_tester script to capture logs in the event of a process timeout: previously, result.stdout was used, which doesn't get populated if the process timed out (resulting in missing logs). Instead, we access output from the error. Experimentation revealed that this output will sometimes be text, and sometimes binary: it's unclear what's responsible for this difference (possibly the encoding settings of the environment running the script). Thus we try to decode the binary, and store it as-is if it's already text.

### DIFF
--- a/scripts/gha/desktop_tester.py
+++ b/scripts/gha/desktop_tester.py
@@ -94,18 +94,26 @@ class Test(object):
   # them as fields so they can be accessed from the main thread.
   def run(self):
     """Executes this testapp."""
-    result = subprocess.run(
-        args=[self.testapp_path],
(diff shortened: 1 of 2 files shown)


vimanyu (approved):
LGTM

PR #212, opened 2020-12-14
Minor fixes to testing in the packaging workflow

anonymous-akorn:
Desktop integration testing was separated from the build script, but not applied to the testing in the packaging workflow (or possibly lost during merge conflict resolution). This applies it, consistent with the integration_tests.yml workflow.

Also updates the setup-msbuild action to 1.0.2 as 1.0.1 no longer works (this is consistent with other usages of the action in these workflows).

The change to integration_tests.yml removes redundant references to github.workspace (no behaviour change).


vimanyu (approved):
LGTM!

PR #213, opened 2020-12-15
Support both Linux ABIs, legacy and C++11

jonsimantov:
This change keeps the legacy ABI but also adds an option to build using the C++11 ABI if FIREBASE_LINUX_USE_CXX11_ABI is set during the cmake configure step.

In packaging, the location of the Linux libraries is now libs/linux/$arch/$abi, where $arch is x86_64 or i386 (as before), and $abi is legacy or cxx11. For example, libs/linux/x86_64/cxx11/libfirebase_app.a. The binary package's CMakeLists.txt has been updated to choose the correct one based on whether the compiler option is enabled (it checks the actual CFLAGS for the option, not the Firebase-specific flag, so it should handle the selection automatically even if the developer's CMake project doesn't provide the Firebase-specific flag).

Also any sub-builds (e.g. boringssl) will also be configured with the requested ABI option.


jonsimantov:
Hmm, I'm seeing a cmake error in the x86 legacy build, any ideas what might be causing this?
CMake Error at /usr/local/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:218 (message):
  Could NOT find Threads (missing: Threads_FOUND)

jonsimantov:
> Hmm, I'm seeing a cmake error in the x86 legacy build, any ideas what might be causing this?
> CMake Error at /usr/local/share/cmake-3.19/Modules/FindPackageHandleStandardArgs.cmake:218 (message):
> Could NOT find Threads (missing: Threads_FOUND)

Fixed, it was that same cflags quoting error I've seen before. Quotes were missing in the legacy option.

a-maurice (approved):

jonsimantov:
Windows build is stuck for unrelated reasons, I'm going to go ahead and merge this anyway.

PR #216, opened 2020-12-16
Work around flaky integration test builds due to gtest

anonymous-akorn:
Currently, we rely on some cmake logic to setup gtest, but this is flaky when building for Android. To work around this, this PR changes build_testapps.py to run the gtest download script before running the integration test setup script: the latter will copy over gtest alongside the test framework files into the project. This will result in the problematic gtest cmake logic being bypassed.

With this in place, the previous workaround of running the android build twice in the presence of a specific failure is no longer necessary. It has been removed.


scripts/gha/build_testapps.py
```diff
@@ -247,7 +247,7 @@ def _build(
   os.chdir(project_dir)
 
   _run_setup_script(root_dir, project_dir)
-
+
```
DellaBitta, line 250: This seems like it can be removed.

DellaBitta (approved):
Approved with comment.

PR #217, opened 2020-12-16
Fix the packaged build to have the correct generated headers, and also include instance ID impl.

jonsimantov:
Fixes building and running integration tests against the packaged SDK binary.


anonymous-akorn (dismissed):

jonsimantov:
Error looks like a transient http error.

PR #218, opened 2020-12-16
Change desktop packaging script to run merge_libraries in parallel.

jonsimantov:
By far the slowest part of C++ desktop packaging is running merge_libraries, especially on Windows. This PR allows the merge_libraries step -- which uses Linux runners to merge Windows libraries - to run in parallel as long as the "parallel" command is installed on the machine.

To enable this change, a bit of code is added to merge_libraries.py to lock and unlock the cache file (which is used across merge_libraries instances) using flock(). It requests a shared lock while reading, and an exclusive lock while writing. The flock() call will block until the process can obtain the lock.

This change saves nearly an hour in packaging time for the longest packaging platform, Windows Debug X64 Static (and Dynamic).


.github/workflows/cpp-packaging.yml
```diff
@@ -433,7 +433,7 @@ jobs:
           for pkg in artifacts/firebase-cpp-sdk-${{ matrix.sdk_platform }}${{ matrix.suffix }}*-build/*.tgz; do
             # determine the build variant based on the artifact filename
             variant=$(sdk-src/build_scripts/desktop/get_variant.sh "${pkg}")
-            sdk-src/build_scripts/desktop/package.sh -b ${pkg} -o firebase-cpp-sdk-${{ matrix.sdk_platform }}${{ matrix.suffix }}-package -p ${{ matrix.sdk_platform }} -t bin -d ${variant} -P python3
+            sdk-src/build_scripts/desktop/package.sh -b ${pkg} -o firebase-cpp-sdk-${{ matrix.sdk_platform }}${{ matrix.suffix }}-package -p ${{ matrix.sdk_platform }} -t bin -d ${variant} -P python3 -j
```
vimanyu, line 436: Does this work on machines that might not have "python3" command. I think I have seen cases where if python3 is the only version on box, then we just have "python" command rather than "python3". But if your actions are running on github runners already, I am guessing this is safe.

.github/workflows/cpp-packaging.yml
```diff
@@ -433,7 +433,7 @@ jobs:
           for pkg in artifacts/firebase-cpp-sdk-${{ matrix.sdk_platform }}${{ matrix.suffix }}*-build/*.tgz; do
             # determine the build variant based on the artifact filename
```
vimanyu, line 434: Just a general comment on this step and not this particular line. 
Is this step supposed to run on all runners (including Windows)?

vimanyu (approved):
Minor comments (questions) but LGTM otherwise.

.github/workflows/cpp-packaging.yml
jonsimantov: It runs on Linux runners for packaging the Linux and Windows SDKs.
It runs on MacOS runners for packaging the Mac SDK (as it needs to use mac-specific tools like lipo and libtool).

.github/workflows/cpp-packaging.yml
jonsimantov: Yeah it runs on the runners just fine already. Is there a python environment variable or similar that I should use here instead?

jonsimantov:
Added an optimization to run the longest job first (after the initial cache population), basically to give Firestore the maximum amount of time to do its packaging while the other jobs run.

build_scripts/desktop/package.sh
```diff
@@ -357,23 +357,34 @@ for product in ${product_list[*]}; do
...
 if [[ ${run_in_parallel} -ne 0 ]]; then
+  # Analytics is the smallest SDK, so it should be the shortest job.
+  shortest=analytics
   echo "There are ${#product_list[@]} jobs to run."
-  echo "Running first job to populate cache, then remaining jobs in parallel..."
-  "${merge_libraries_tmp}/merge_${product_list[0]}.sh"
+  echo "Running shortest job to populate cache, then remaining jobs in parallel..."
+  "${merge_libraries_tmp}/merge_${shortest}.sh"
+  # Zero out the job that we already did.
+  echo "#!/bin/bash" > "${merge_libraries_tmp}/merge_${shortest}.sh"
```
jonsimantov, line 372: Come to think of it I could also just rm the script, lmk if you have a preference.

.github/workflows/cpp-packaging.yml
vimanyu: Unfortunately I dont know of any. 
In other python scripts, we had this condition to check for "python3" and if it fails, look for "python".
Another way could be to run "python --version" and query the version but if its running fine for now on the runners, I dont think its a big deal.

build_scripts/desktop/package.sh
vimanyu: I think it is fine to leave it as it is.

vimanyu (approved):
LGTM

PR #230, opened 2021-01-12
Automatically expand summary step in failed jobs

anonymous-akorn:
A small quality of life improvement. When opening the page to a failed job, Github Actions will automatically expand the last failing step. We can exploit this to automatically expand the summary step by forcing it to fail, which is the preferred starting point for figuring out what failed. Since we don't want to fail otherwise successful jobs, we only execute the summary step if a prior step failed. The summary is only needed when something fails.

This eliminates some navigation, as well as the knowledge burden of knowing to navigate to the summary step to get an overview of what went wrong.


DellaBitta (dismissed):
I recall in the past some steps were failing silently.. would running cat on the summary always and then adding an if failure() clause to run exit 1 be useful?

anonymous-akorn:
> I recall in the past some steps were failing silently.. would running cat on the summary always and then adding an if failure() clause to run exit 1 be useful?

Yeah that's better, I overlooked the possibility of checking the status within the step's logic.

DellaBitta (approved):

PR #231, opened 2021-01-13
Run packaging integration tests by default

jonsimantov:
Set the C++ Packaging workflow to run integration tests by default. Log a message if tests are skipped.

(We also check for a blank value because when running on merge or cron, all input params are blank.)

### DIFF
--- a/.github/workflows/cpp-packaging.yml
+++ b/.github/workflows/cpp-packaging.yml
@@ -15,8 +15,8 @@ on:
         description: 'public version # to test against'
       downloadPreviousRun:
         description: 'previous run # to test against'
-      runIntegrationTests:
-        description: 'run integration tests?'
+      skipIntegrationTests:
+        description: 'skip integration tests?'
         default: 0
 
 env:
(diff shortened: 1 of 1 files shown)


anonymous-akorn (dismissed):

.github/workflows/cpp-packaging.yml
```diff
@@ -34,6 +34,11 @@ jobs:
     name: log-inputs
     runs-on: ubuntu-latest
     steps:
+      - name: log if skipping integration tests
+        if: |
+          github.event.inputs.skipIntegrationTests != 0 && github.event.inputs.skipIntegrationTests != ''
```
anonymous-akorn: Should this be || ?

jonsimantov: It was logically correct, but I made it the same as the check below (with a !) so it's clear that it's testing the same condition.

anonymous-akorn: Or rather, should it be "0 or empty" instead of "not 0 and not empty"?

anonymous-akorn: er, my mistake, I'm confusing myself

this is correct

anonymous-akorn (approved):

PR #248, opened 2021-01-22
Explicitly list the Flatbuffer source files

a-maurice:
FlatBuffers_Library_SRCS is not always set, so just list the source files directly.


jonsimantov (approved):
Looks good as long as you can confirm (maybe with a debug printout) that the flatbuffers source dir is set correctly.

PR #255, opened 2021-01-27
Bugfix/integration tests mac openssl error

vimanyu:
- Adds support for using vcpkg openssl in `build_desktop.py`.
- Added openssl installation for mac and linux as a step if integration tests depend on openssl instead of boringssl.


.github/workflows/integration_tests.yml
```diff
@@ -135,6 +135,21 @@ jobs:
...
           choco install openssl -r
 
+      - name: Install OpenSSL (MacOS)
+        if: matrix.target_platform == 'Desktop' &&
+            matrix.ssl_variant == 'openssl' &&
+            startsWith(matrix.os, 'macos')
+        run: |
+          brew install openssl
+
+      - name: Install OpenSSL (MacOS)
```
DellaBitta, line 145: replace MacOS with Linux

.github/workflows/integration_tests.yml
```diff
@@ -135,6 +135,21 @@ jobs:
...
+        if: matrix.target_platform == 'Desktop' &&
+            matrix.ssl_variant == 'openssl' &&
+            startsWith(matrix.os, 'ubuntu')
+        run: |
+          sudo apt install openssl
+
+
```
DellaBitta, line 152: remove extra line.

.github/workflows/integration_tests.yml
DellaBitta: Do we need this step?  Aren't our linux builds succesfull without it?

DellaBitta (changes requested):
Out of curiosity, what is this exactly?

"Submodule vcpkg updated 3754 files"

Does github do a deep comparison of submodule version change implications?

.github/workflows/integration_tests.yml
vimanyu: Good catch. Replaced MacOS with Linux.
I left the linux step for consistency because I dont know how is it finding openssl otherwise. The github runners (ubuntu 18) does not have openssl pre-installed.

scripts/gha/build_desktop.py
```diff
@@ -264,6 +277,7 @@ def parse_cmdline_args():
   parser.add_argument('--config', default='Release', help='Release/Debug config')
   parser.add_argument('--target', nargs='+', help='A list of CMake build targets (eg: firebase_app firebase_auth)')
   parser.add_argument('--target_format', default=None, help='(Mac only) whether to output frameworks (default) or libraries.')
+  parser.add_argument('--use_openssl', action='store_true', default=None, help='Use openssl for build instead of boringssl')
```
jonsimantov, line 280: What do you thin about just adding an openssl flag to the prereqs script also? (And we can move the Windows openssl setup into there too.)

vimanyu: Thats a good point but as per our original design, we wanted to keep all architecture specific steps under build_desktop. And if we are using vcpkg, we have to specify the exact architecture to build. Something to discuss in our "various ways to build" meeting but if we want to continue using vcpkg, it makes sense to not have openssl under the prereqs script. For now, I will leave it here but it should be trivial to move this to prereqs if we decide to go that way after the meeting.

DellaBitta (approved):

PR #256, opened 2021-01-27
Change Podfile version for integration tests to 7.0.0.

jonsimantov:
Update the integration test Podfile versions to match the versions in ios_pod/Podfile.

### DIFF
--- a/admob/integration_test/Podfile
+++ b/admob/integration_test/Podfile
@@ -1,10 +1,10 @@
 
 source 'https://github.com/CocoaPods/Specs.git'
-platform :ios, '8.0'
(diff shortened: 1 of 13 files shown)


paulb777:
Any reason for 7.0.0 versus the current 7.5.0 or letting it float to the most recent minor version?

a-maurice (approved):

a-maurice:
> Any reason for 7.0.0 versus the current 7.5.0 or letting it float to the most recent minor version?

Updating them to 7.5.0 introduces some additional risk, as we would also need to update the Android versions, etc, but after this release is out, we will be following it with a release with the newer versions.

PR #274, opened 2021-02-09
Improve how we deal with artifacts

jonsimantov:
Several enhancements to reduce transient failures in the packaging step, and just make the workflow cleaner in general:
1. Delete build artifacts sooner, after they are no longer needed, to reduce the number of artifacts final-merge-packages has to download.
2. Update to the latest upload-artifact and download-artifact versions, which add better HTTP retry logic.
3. Switch to a newer version of the delete-artifact action that allows wildcards.


vimanyu (dismissed):
LGTM!

jonsimantov:
Realized I could improve this even more, PTAL. Will wait for packaging run to finish before merging.

vimanyu (dismissed):

vimanyu (approved):

Issue #271, opened 2021-02-05
[Win64] Prebuilt 7.0.1 linkage errors.


Pandoa:
### [REQUIRED] Please fill in the following fields:

  * Pre-built SDK from the [website](https://firebase.google.com/download/cpp)
  * Firebase C++ SDK version: 7.0.1
  * Problematic Firebase Component: App, Remote Config
  * Other Firebase Components in use: Auth, Firestore
  * Platform you are using the C++ SDK on: Windows x64-Release /MD vs2019
  * Platform you are targeting: desktop

### [REQUIRED] Please describe the issue here:

Impossible to build an app linked against the SDK with a linkage error if the following libraries are linked: 
1. firebase_app.lib
2. firebase_remote_config.lib
3. firebase_firestore.lib
4. firebase_auth.lib

Exact linkage error:
```
1>firebase_app.lib(0865aae1211b26b3ccf22fd1b579e000_protobuf_nanopb_static.dir_Release_pb_decode.obj) : error LNK2005: f_b_pb_decode already defined in firebase_remote_config.lib(c80cae8cb6fc354c32078cec2fc51bbc_firebase_remote_config.dir_Release_pb_decode.obj)
1>firebase_app.lib(0865aae1211b26b3ccf22fd1b579e000_protobuf_nanopb_static.dir_Release_pb_decode.obj) : error LNK2005: f_b_pb_decode_noinit already defined in firebase_remote_config.lib(c80cae8cb6fc354c32078cec2fc51bbc_firebase_remote_config.dir_Release_pb_decode.obj)
1>firebase_app.lib(0865aae1211b26b3ccf22fd1b579e000_protobuf_nanopb_static.dir_Release_pb_decode.obj) : error LNK2005: f_b_pb_decode_delimited already defined in firebase_remote_config.lib(c80cae8cb6fc354c32078cec2fc51bbc_firebase_remote_config.dir_Release_pb_decode.obj)
...
fatal error LNK1169: one or more multiply defined symbols found.
```

#### Steps to reproduce:

1. Download the Firebase C++ SDK version 7.0.1.
2. Create a new blank Visual Studio project.
3. Compile the Relevant Code provided below.

Have you been able to reproduce this issue with just the [Firebase C++ quickstarts](https://github.com/firebase/quickstart-cpp) ? Yes
Issue repro rate: `100%`

Project to reproduce: 
[...]

PR #275, opened 2021-02-09
Remove extra inclusion of pb_common.c and pb_decode.c, to avoid multiple symbol definitions.

jonsimantov:
Fix for #271.

### DIFF
--- a/remote_config/CMakeLists.txt
+++ b/remote_config/CMakeLists.txt
@@ -66,6 +66,11 @@ if (NOT PROTOBUF_FOUND)
 else()
   # Generates config.pb.c and config.pb.h
   NANOPB_GENERATE_CPP(PROTO_SRCS PROTO_HDRS src_protos/config.proto)
+  # Remove pb_common.c and pb_decode.c from PROTO_SRCS, as they are
+  # provided by Firebase App. Including them here causes symbols to
+  # have multiple definitions.
+  list(FILTER PROTO_SRCS EXCLUDE REGEX "/pb_.*\\.c")
+
   set(desktop_SRCS
       ${PROTO_SRCS}
       ${PROTO_HDRS}
--- a/remote_config/CMakeLists.txt
+++ b/remote_config/CMakeLists.txt
@@ -116,6 +116,7 @@ else()
   set(additional_link_LIB
       firebase_rest_lib
       firebase_instance_id_desktop_impl
+      protobuf-nanopb-static
       flatbuffers)
 endif()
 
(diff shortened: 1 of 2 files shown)


remote_config/CMakeLists.txt
```diff
@@ -66,6 +66,11 @@ if (NOT PROTOBUF_FOUND)
 else()
   # Generates config.pb.c and config.pb.h
   NANOPB_GENERATE_CPP(PROTO_SRCS PROTO_HDRS src_protos/config.proto)
+  # Remove pb_common.c and pb_decode.c from PROTO_SRCS, as they are
+  # provided by Firebase App. Including them here causes symbols to
+  # have multiple definitions.
+  list(FILTER PROTO_SRCS EXCLUDE REGEX "/pb_.*\\.c")
```
a-maurice, line 72: I think this would only fix it if Firestore is included in the build, since that is the other thing adding proto.

I think you could put a
if(FIREBASE_INCLUDE_FIRESTORE)
...
endif()

jonsimantov: > I think this would only fix it if Firestore is included in the build, since that is the other thing adding proto.

I actually don't think I need to do this. Here are the situations:

### Binary SDK
#### Currently
Nanopb is vendored into Firebase App via packaging. It's referenced by Firestore. It's [currently] included in Remote Config as well, not via packaging but via building and linking them in PROTO_SRCS here. This only causes the multiple definitions error when Firestore is linked because the act of having Firestore reference Nanopb externally causes those symbols to have to be resolved by the linker, which then sees both versions.

(Neither Firestore nor its dependencies use this NANOPB_GENERATE_CPP rule, so the pb_decode and pb_common files are not actually linked into Firestore.)

#### With this change
Nanopb is vendored into Firebase App via packaging. It's referenced by Firestore and Remote Config. No multiple definitions.

### Source SDK
#### Currently
Nanopb is included in Remote Config via PROTO_SRCS, as above. It's set as a linker library in Firestore. These could possibly conflict depending on the order in which you link Remote Config and Firestore, and which compiler you are actually using (MSVC vs GCC vs clang).

#### With this change
Nanopb is set as a linker library in both Firestore and Remote Config. If you include both libraries in your app, they will both depend on protobuf-nanopb-static and that file will only be linked once. We download nanopb as part of our external dependencies, and find_package(Nanopb) is included in our root CMakeLists.txt, so that library target should be available even without Firestore enabled.

a-maurice (approved):

PR #287, opened 2021-02-26
build ios xcframework

sunmou99:
Build and providing iOS XCFramework instead of Framework.
Testing against XCFramework.
XCode version is updated to 12.0.


build_scripts/ios/build.sh
```diff
@@ -35,10 +40,10 @@ cmakeBuild=true
 IFS=',' # split options on ',' characters
 while getopts ":b:s:a:t:g:ch" opt; do
     case $opt in
-	h)
-	    usage
-	    exit 0
-	    ;;
+        h)
```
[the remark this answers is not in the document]
sunmou99: No tabs. My Editor inserts spaces pressing tab.

build_scripts/ios/build.sh
```diff
@@ -101,91 +117,130 @@ buildpath=$(mkdir -p ${buildpath} && cd ${buildpath} && pwd)    #full path
...
+        for arch in ${architectures[@]}; do 
+            if [[ "${platform}" == "device" && " ${DEVICE_ARCHITECTURES[@]} " =~ " ${arch} " ]]; then
+                toolchain="cmake/toolchains/ios.cmake"
+            elif [[ "${platform}" == "simulator" && " ${SIMULATOR_ARCHITECTURES[@]} " =~ " ${arch} " ]]; then
+                toolchain="cmake/toolchains/ios_simulator.cmake"
+            else
+                continue
+            fi
+
+            echo "generate Makefiles start"
```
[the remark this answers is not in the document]
sunmou99: Yes.

.github/workflows/cpp-packaging.yml
```diff
@@ -30,9 +30,9 @@ env:
   demumbleVer: "1.1.0"
   # Use SHA256 for hashing files.
   hashCommand: "sha256sum"
-  # Xcode version 11.7 is the version we build the SDK with.
+  # Xcode version 12 is the version we build the SDK with.
   # Our MacOS runners will use the version in /Applications/Xcode_${xcodeVersion}.app
-  xcodeVersion: "11.7"
```
jonsimantov, line 35: This changes our Xcode minimum requirement from 11.7 to 12, so we should discuss it before making this change.

Does xcode 11 not support xcframeworks at all?

We should check what the minimum Xcode version is for the Firebase iOS SDK as well.

sunmou99: Xcode 11 support xcframeworks. But only 12+ support simulator-arm64.

sunmou99: We need to build xcframeworks with xcode 12+. But it is possible to build our integration_test.app with Xcode 11.7 (I tried locally).
Also checked the minimum Xcode version for the Firebase iOS SDK. It's Xcode 12.0 (or later).
https://github.com/firebase/firebase-ios-sdk#development

jonsimantov: OK, we should bump our minimum listed requirement to 12 then, and ensure that all of our workflows use the same version.

jonsimantov:
Looks good but just edit the Xcode version in release_build_files/readme.md, and add a note for an upcoming "7.2.0" release saying that the minimum Xcode version is changed. (You can model this after the note from the previous time we changed Xcode to 11.7.)

jonsimantov:
Also could you run the packaging workflow one last time with this change to make sure the tests all still work?

sunmou99:
> Looks good but just edit the Xcode version in release_build_files/readme.md, and add a note for an upcoming "7.2.0" release saying that the minimum Xcode version is changed. (You can model this after the note from the previous time we changed Xcode to 11.7.)

Note Added. Also update the Framework to XCFramework.



> Also could you run the packaging workflow one last time with this change (merging from main first) to make sure the tests all still work?

integration _test workflow passed: https://github.com/firebase/firebase-cpp-sdk/actions/runs/618737579

jonsimantov (approved):

PR #289, opened 2021-02-27
Feature/android resources crash old devices

vimanyu:
Fixes an issue on old Android devices (< Android 6.0) where we observed apps crashing on startup. This was happening because the various gradle "_resources" (app_resources, google_api_resources, database_resources etc) sub-projects' output dex files were being cached under the same file name in the C++ layer. 
This was causing a race condition amongst the threads handling each resources and the stack traces were inconsistent between runs. 
Separating each resource project's output into a non-conflicting file name fixes the problem.

### DIFF
--- a/admob/CMakeLists.txt
+++ b/admob/CMakeLists.txt
@@ -28,9 +28,9 @@ set(common_SRCS
 
 # Define the resource build needed for Android
 firebase_cpp_gradle(":admob:admob_resources:generateDexJarRelease"
-    "${CMAKE_CURRENT_LIST_DIR}/admob_resources/build/dexed.jar")
+    "${CMAKE_CURRENT_LIST_DIR}/admob_resources/build/admob_resourcs_lib.jar")
 binary_to_array("admob_resources"
-    "${CMAKE_CURRENT_LIST_DIR}/admob_resources/build/dexed.jar"
+    "${CMAKE_CURRENT_LIST_DIR}/admob_resources/build/admob_resources_lib.jar"
     "firebase_admob"
     "${FIREBASE_GEN_FILE_DIR}/admob")
 
--- a/admob/CMakeLists.txt
+++ b/admob/CMakeLists.txt
@@ -28,7 +28,7 @@ set(common_SRCS
 
 # Define the resource build needed for Android
 firebase_cpp_gradle(":admob:admob_resources:generateDexJarRelease"
-    "${CMAKE_CURRENT_LIST_DIR}/admob_resources/build/admob_resourcs_lib.jar")
+    "${CMAKE_CURRENT_LIST_DIR}/admob_resources/build/admob_resources_lib.jar")
 binary_to_array("admob_resources"
     "${CMAKE_CURRENT_LIST_DIR}/admob_resources/build/admob_resources_lib.jar"
     "firebase_admob"
--- a/android_build_files/extract_and_dex.gradle
+++ b/android_build_files/extract_and_dex.gradle
@@ -43,7 +43,7 @@ def defineExtractionTasks(String resourceName, String buildType) {
       include "classes.jar"
     }
 
-    String dexedJar = "$buildDir/dexed.jar"
+    String dexedJar = "$buildDir/${resourceName}_lib.jar"
     String outPro = "$buildDir/${resourceName}.pro"
     Task dexJar = tasks.create(
         name: "generateDexJar$buildType",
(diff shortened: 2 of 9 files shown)


cynthiajoan (commented):
Are these all the cmakelist files that we need to change? How about the rest productions like remote config?

vimanyu:
Good point. Remote config and other products do not suffer from this directly because they do not have a separate "_resources" sub project building anything. I think Remote Config was broken because app/auth was broken.

cynthiajoan (approved):

DellaBitta (dismissed):

admob/CMakeLists.txt
```diff
@@ -28,9 +28,9 @@ set(common_SRCS
 
 # Define the resource build needed for Android
 firebase_cpp_gradle(":admob:admob_resources:generateDexJarRelease"
-    "${CMAKE_CURRENT_LIST_DIR}/admob_resources/build/dexed.jar")
+    "${CMAKE_CURRENT_LIST_DIR}/admob_resources/build/admob_resources_lib.jar")
```
jonsimantov, line 31: I will say, I still think this is a bit dangerous - it requires that every CMake file included in the project specify a different basename, which could cause an issue later on.

WDYT about modifying cmake/binary_to_array.cmake to use the specified namespace as a prefix for the filename, just in case?

jonsimantov (changes requested):

admob/CMakeLists.txt
jonsimantov: Or actually just the specified NAME - is this just a bug that binary_to_array.py is not correctly using the --filename parameter, and is instead using the basename of the input file?

jonsimantov: (I just want to be sure this is fixed in the correct place - whether it's here, in binary_to_array.cmake, binary_to_array.py, or in the extract embedded files code on device.)

vimanyu: I think this is a hidden implementation detail for just the resources sub projects and should be fine because we are explicitly stating these build rules for each of the products anyway. But yes, we can modify the _filename in our python script too and let me give that a try. The only thing there is that currently there is a one to one mapping between the name of the file in intermediate files and the final output of python script - the python file extracts binary content from a file "dexed.jar" and that is indeed the _filename. We will be changing it to "read dexed.jar but consider it as <product>dexed.jar".

jonsimantov: It looks like if the binary_to_array.cmake doesn't specify --filename=, then the python script uses the basename of the input file as the filename. This seems dangerous (even though it never caused an issue in the previous build), for the reason fixed by this bug. I think it would be good to add a required filename parameter to the binary_to_array cmake function and have it specify the filename explicitly, in addition to the change you've made to the jar file name.

vimanyu: Yup. Noticed that and spawned integration tests with that change half an hour ago. Waiting for them to finish and I am almost confident that will work. Good thing is that if someone wants to use the python script directly, then we dont modify the default filename. But the cmake file can dictate the filename. I will close this PR and create a new one if this works.

a-maurice: So, after finally understanding why this fix works, I think just renaming the outputs as you have done should be fine.  Getting information to the "extract_and_dex" step is really painful, so I don't want to try to future proof it around possible name collisions when we could just avoid those collisions.

jonsimantov: Discussed with @a-maurice, the fix is good as-is. 

But could you add a comment in binary_to_array cmake function in cmake/binary_to_array.cmake that all embedded resources must have unique base filenames because the directory names are removed? This way if we come across an issue in the future due to a resource file having the same name, we'll have a place to notice the caveat.

vimanyu: So this is the alternate way to fix it.
https://github.com/firebase/firebase-cpp-sdk/compare/feature/android-resources-crash-old-devices2?expand=1

We don't change any other CMake files except the one file and the python file already accepts a flag for a custom filename. Just a matter of choosing which fix makes most sense.

jonsimantov: This fix is fine, just add the comment I mentioned [here](https://github.com/firebase/firebase-cpp-sdk/pull/289#discussion_r584967357) and a note in the release readme.md that this is fixed, and we're good.

jonsimantov (approved):

PR #290, opened 2021-03-01
Remove the extra delay in storage integration test on mobile.

jonsimantov:
This delay broke the Storage integration test on mobile, though it helps with flakes on desktop.


DellaBitta (approved):
Is there a ticket associated with this that we can reference?

DellaBitta:
Also it would be nice if the description detailed why this change was made.  You know, for the archives.

jonsimantov:
> Is there a ticket associated with this that we can reference?

No ticket, should I add one retroactively?

PR #291, opened 2021-03-01
[RemoteConfig]Update desktop rest unit test

cynthiajoan:
[RemoteConfig] Update rest unit test and remove nanopb dependency


remote_config/CMakeLists.txt
```diff
@@ -181,4 +170,4 @@ if(FIREBASE_CPP_BUILD_TESTS)
 endif()
 
 cpp_pack_library(firebase_remote_config "")
-cpp_pack_public_headers()
+cpp_pack_public_headers()
```
DellaBitta, line 173: Could you add a newline here?

remote_config/src/desktop/remote_config_request.h
```diff
@@ -21,12 +21,20 @@
 #include "remote_config/request_generated.h"
 #include "remote_config/request_resource.h"
 
+#ifdef FIREBASE_TESTING
+#include "testing/base/public/gunit.h"
+#endif  // FIREBASE_TESTING
+
```
DellaBitta, line 27: Intrigued. I don't know if I've seent his before.  Is this a standard practice?

remote_config/src/desktop/remote_config_request.h
```diff
@@ -82,4 +94,4 @@ class RemoteConfigRequest
 }  // namespace remote_config
 }  // namespace firebase
 
-#endif  // FIREBASE_REMOTE_CONFIG_CLIENT_CPP_SRC_DESKTOP_REMOTE_CONFIG_REQUEST_H_
+#endif  // FIREBASE_REMOTE_CONFIG_CLIENT_CPP_SRC_DESKTOP_REMOTE_CONFIG_REQUEST_H_
```
DellaBitta, line 97: Add a newline.

remote_config/src/desktop/remote_config_response.cc
```diff
@@ -101,4 +104,4 @@ void RemoteConfigResponse::MarkCompleted() {
 
 }  // namespace internal
 }  // namespace remote_config
-}  // namespace firebase
+}  // namespace firebase
```
DellaBitta, line 107: Newline

remote_config/src/desktop/remote_config_response.cc
```diff
@@ -90,6 +90,9 @@ Variant RemoteConfigResponse::GetEntries() { return entries_; }
 // Mark the response completed for both header and body.
 void RemoteConfigResponse::MarkCompleted() {
   ResponseJson::MarkCompleted();
+  if (strlen(GetBody()) == 0) {
```
DellaBitta, line 93: Since no other lines of code use GetBody() in this function, but instead call into flatbuffers, I think it's worth writting a comment above this line to describe why this is a shortcut out of the function.

remote_config/src/desktop/request.fbs
```diff
@@ -34,4 +35,4 @@ table Request {
   analyticsUserProperties:string;
 }
 
-root_type Request;
+root_type Request;
```
DellaBitta, line 38: Newline.

remote_config/src/desktop/rest.cc
```diff
@@ -152,72 +149,33 @@ void RemoteConfigREST::SetupRestRequest(
       app_instance_id_token_);  // TODO(cynthiajiang) change to installations
 
   rc_request_.SetPlatformVersion("2");
+  std::string locale = firebase::internal::GetLocale();
+  if (locale.length() != 0) {
+    rc_request_.SetLanguageCode(locale);
+    rc_request_.SetCountryCode(
+        locale.substr(0, 2));  // get country code from locale
```
DellaBitta, line 156: How sure are we that language codes returned here are 2 characters long.  Standard language codes may be longer.  [For instance](https://www.google.com/search?q=locale+country+codes&oq=locale+country+code).

remote_config/src/desktop/rest.cc
```diff
@@ -269,4 +227,4 @@ uint64_t RemoteConfigREST::MillisecondsSinceEpoch() {
 
 }  // namespace internal
 }  // namespace remote_config
-}  // namespace firebase
+}  // namespace firebase
```
DellaBitta, line 230: Newline

remote_config/src/desktop/rest.h
```diff
@@ -162,4 +143,4 @@ class RemoteConfigREST {
 }  // namespace remote_config
 }  // namespace firebase
 
-#endif  // FIREBASE_REMOTE_CONFIG_CLIENT_CPP_SRC_DESKTOP_REST_H_
+#endif  // FIREBASE_REMOTE_CONFIG_CLIENT_CPP_SRC_DESKTOP_REST_H_
```
DellaBitta, line 146: Newline.  Maybe your editor has a setting?

remote_config/src/desktop/rest_fake.cc
```diff
@@ -70,4 +65,4 @@ uint64_t RemoteConfigREST::MillisecondsSinceEpoch() { return 0; }
 
 }  // namespace internal
 }  // namespace remote_config
-}  // namespace firebase
+}  // namespace firebase
```
DellaBitta, line 68: Newline

DellaBitta (changes requested):

remote_config/src/desktop/remote_config_response.cc
jonsimantov: It's also faster not to strlen() the whole body in case it's quite large. Just check if GetBody()[0] == '\0'.

remote_config/src/desktop/remote_config_request.h
cynthiajoan: So first this code change is not correct, I was using the reference from internal pattern so I changed that. Second I don't know whether it's a standard practice, but we do have existing case in rest.h, I guess we can accept this?

remote_config/src/desktop/remote_config_response.cc
cynthiajoan: DONE

remote_config/src/desktop/rest.cc
cynthiajoan: Yeah the locale code could be longer, but I thought the country code would always be the first two letters

remote_config/CMakeLists.txt
cynthiajoan: DONE

remote_config/src/desktop/remote_config_request.h
cynthiajoan: DONE

remote_config/src/desktop/rest.cc
cynthiajoan: Need to find out a way to grab country code, the current approach is to get the two digit language from locale

DellaBitta (approved):

PR #297, opened 2021-03-03
Add "expanded matrix" option to workflows

DellaBitta:
Support for an expanded build and test matrix in all workflows. The expanded_matrix will grow with more host and target device configurations and will generally be meant for cron jobs and/or heavier testing around a release.

Notes: 
- A new python script contains json definitions for a normal matrix and an expanded matrix. It can be extended to alter the build matrix for the following workflows: android, desktop, integration tests, ios.  
- Created a new prepare_matrix job in each workflow which detects how the job was executed and any manual override parameters that were defined.  This, in turn, invokes the python script to export a standard, extended, or manually overridden matrix.
- Added manual triggers to the following workflows: android, desktop, ios.
- Added nightly trigger to integration tests.
- Altered how some jobs are named.    **Note:** this will make it appear that some of our required workflows haven't been executed, however they've simply been renamed.
   - Fixed Android naming convention which was attempting to name mangle using values that didn't exist. 
   - Worfklows use new dynamic matrix parameters such as xcode version to mangle their name.
- To "prove" that the worfklows can be dynamicallyc onfigured with an extended matrix, Created an extended matrix definition for desktop and iOS builds to build with both xcode 12 and 12.4, and to build for all windows x86/x64 static/dynamic targets.
- Removed some imports from the desktop worfklow. These imports would have added jobs regardless of how we dynamcially configure the matrix, even if the targets weren't desired.  Instead, these jobs are part of the matrix definitions in the script.
- Added excludes to filter matrix values that don't pertain to all build types.  For instance, removed multiple xcode versions from windows and linux builds.

TODO: In a subsequent change we may export the matrix as one json string for each worfklow. For now it remains exported to distinct fields.


anonymous-akorn:
As a general thought, I wonder how much can be moved from workflow files into a centralized location. As I understand it, a workflow matrix can be populated entirely via a single 'fromJson' invocation. That JSON could come from a static file, or a script, possibly with inputs to handle all the different cases. Among other things, that might provide a way to have a single source of truth for all parameters, in addition to greatly reducing the amount of configuration.

.github/workflows/cpp-packaging.yml
```diff
@@ -74,10 +97,11 @@ jobs:
   build_tools:
     name: build-tools-${{ matrix.tools_platform }}
     runs-on: ${{ matrix.os }}
+    needs: prepare_matrix
     if: ${{ github.event.inputs.downloadPublicVersion == '' && github.event.inputs.downloadPreviousRun == '' }}
     strategy:
       matrix:
-        os: [ubuntu-latest, macos-latest]
+        os: ${{ fromJson(needs.prepare_matrix.outputs.matrix_build_tools_os) }}
```
jonsimantov, line 104: Is this just from the json for completeness? Since it has nothing to do with testing?

.github/workflows/cpp-packaging.yml
```diff
@@ -35,12 +39,31 @@ env:
...
+  prepare_matrix:
+    runs-on: ubuntu-latest
+    env:
+      MATRIX_BUILD_TOOLS_OS: 'ubuntu-latest,macos-latest'
+      EXPANDED_MATRIX_BUILD_TOOLS_OS: 'ubuntu-latest,macos-latest'
+    outputs:
+      matrix_build_tools_os: ${{ steps.export-result.outputs.matrix_build_tools_os }}
+    steps:
+      - name: Use expanded matrix
+        if: github.event.inputs.expanded_matrix == '1'
```
jonsimantov, line 51: Also use expanded matrix if this is a build triggered by cron?

.github/workflows/cpp-packaging.yml
DellaBitta: I was unsure what part of this script should be made dynamic based on the expanded matrix.  I'm going to revert this part of the change for now.

.github/workflows/cpp-packaging.yml
DellaBitta: Done

.github/workflows/cpp-packaging.yml
```diff
@@ -35,9 +39,26 @@ env:
   xcodeVersion: "12"
 
 jobs:
+  prepare_matrix:
+    runs-on: ubuntu-latest
+    env:
+      USE_EXPANDED_MATRIX: 0
```
vimanyu, line 45: Am probably missing something here but is there a reason why this is different from the `android.yml` workflow? We didn't define a separate job evn variable there.

.github/workflows/desktop.yml
```diff
@@ -3,54 +3,105 @@ name: Desktop Builds
...
+        if: github.event.inputs.expanded_matrix == '1'
+        run: |
+          echo "EXPANDED_MATRIX_PARAM="-e=1" >> $GITHUB_ENV
+      - id: export-result
+        run: |
+          echo "::set-output name=matrix_os::$( python scripts/gha/print_matrix_configuration.py -w desktop -k os ${EXPANDED_MATRIX_PARAM})"
+          echo "::set-output name=matrix_build_type::$( python scripts/gha/print_matrix_configuration.py -w desktop -k build_type ${EXPANDED_MATRIX_PARAM})"
+          echo "::set-output name=matrix_architecture::$( python scripts/gha/print_matrix_configuration.py -w desktop -k architecture ${EXPANDED_MATRIX_PARAM})"
+          echo "::set-output name=msvc_runtime::$( python scripts/gha/print_matrix_configuration.py -w desktop -k msvc_runtime ${EXPANDED_MATRIX_PARAM})"
+          echo "::set-output name=vcpkg_triplet_suffix::$( python scripts/gha/print_matrix_configuration.py -w desktop -k vcpkg_triplet_suffix ${EXPANDED_MATRIX_PARAM})"
```
vimanyu, line 41: I think vcpkg_triplet_suffix shouldn't be in the matrix parameters. This can be easily derived from other values in the matrix and avoiding this in the matrix will trim down the amount of "excludes" in the workflow.

vimanyu: It could just be an environment variable that we define and pass to the steps.

.github/workflows/integration_tests.yml
```diff
@@ -31,26 +33,45 @@ on:
       ios_version:
         description: 'iOS device version'
         default: '11.4'
+      use_expanded_matrix:
+        description: 'Enabling (1) will run an extended test matrix instead of your config defined above'
```
vimanyu, line 37: nit: for consistency, keep the same description across all workflows

vimanyu (changes requested):
Thanks for doing this. Looks good for most part except few minor suggestions.
I feel trimming down the "includes" and "excludes" will make the workflows easier to follow/debug.

.github/workflows/cpp-packaging.yml
DellaBitta: The other workflows drive the expanded build matrix through the python script.  In those yml files we optionally define  ${EXPANDED_MATRIX_PARAM} as is "-e=1" for expanded runs.  Its not defaulted on purpose becuase we use it's lack of exisitence to withdraw the parameter from the python command line arguments.  For instance:

`python scripts/gha/print_matrix_configuration.py -w android -k os ${EXPANDED_MATRIX_PARAM}`

if ${EXPANDED_MATRIX_PARAM} isn' defined then it resolves to `python scripts/gha/print_matrix_configuration.py -w android -k os`.  But if it is defined to `-e=1` then the python execution looks like `python scripts/gha/print_matrix_configuration.py -w android -k os -e=1`.


However, for this workflow, we don't have an expanded build matrix so we don't yet invoke the python script.  In the future I believe that the intention is to have an expanded test matrix though, and this work simply adds an exported bool so that the other jobs may be rewritten to make their own choices on what to do for an expanded test matrix.

I signal this with either a value of `1` or a `0`.  I didn't want the two optoins to be  `1` or `non-value`, so this this env directive defaults the value to `01.  If I didn't do this then the line below `echo "::set-output name=use_expanded_matrix::${USE_EXPANDED_MATRIX}"` would have resolved to `echo "::set-output name=use_expanded_matrix::"` instead of `echo "::set-output name=use_expanded_matrix::0"`, which I think is cleaner.

.github/workflows/integration_tests.yml
DellaBitta: The description here is different to avoid confusion .  This particular workflow has a lot of parameters, almost all of which will be ignored if an expanded matrix is selected.  I didn't just want to say "Use and Expanded Matrix?" here and have the user wonder why they defined to only run a linux datbase api build but suddently the expanded matrix meant that we ignored those configuration changes.  I'll try to tweek it so that at last the begining conforms with the other scripts.

.github/workflows/desktop.yml
DellaBitta: In the previous version of this workflow th VCKPKG suffix was being added to the matrix with include statements below. Unfortunatley the use of include restricts us from dynamically excluding some builds through the configuration above since the include would simply re-add the build to the job matrix again.

that is to say that the previous implementation meant that we had to run with all windows, linux and mac, always.

I'll see if I can change this in a timely manner, but since the original script also had vcpkg in the matrix this might be another story/ticket to resolve later.

DellaBitta: Fixed!

.github/workflows/integration_tests.yml
DellaBitta: Updated, but if you still think that all descriptoins should be the same then please let me know!

vimanyu (approved):
LGTM

.github/workflows/cpp-packaging.yml
```diff
@@ -35,9 +39,26 @@ env:
   xcodeVersion: "12"
 
 jobs:
+  prepare_matrix:
```
jonsimantov, line 42: So, I believe we should actually *use* the prepared matrix on the "tests" job near the bottom of this file. I'd rather not leave this for a subsequent PR.

PR #300, opened 2021-03-04
Fixed race condition when removing listeners.

alexames:
Fixed a crash that can occur as a result of a race condition when
rapidly removing and deleting a listener.

When a listener is added or removed, it doesn't get added or removed
immediately. Instead, a transaction is scheduled to run at some point in
the future. If the developer adds and removes a listener in quick
succession and then deletes it, those scheduled callbacks might not run
until after the listener has already been deleted, resulting in calling
a member function of a now garbage pointer and crashing.

The solution is to immediately mark a listener as having been removed
from the main thread (as well as scheduling it to be removed from the
background thread, as it was already doing). Any new event that come in
on a listener that has been marked removed will be ignored. The
developer is free to delete their Listener after that point without fear
that an event fire from the background thread after it has been deleted.

### DIFF
--- a/release_build_files/readme.md
+++ b/release_build_files/readme.md
@@ -576,6 +576,9 @@ code.
     -   General (iOS): iOS SDKs are now built using Xcode 12.
     -   General (iOS): iOS SDKs are now providing XCFrameworks instead of 
         Frameworks.
+    -   Database: Fixed a potential crash that can occur as a result of a race
+        condidtion when adding, removing and deleting `ValueListener`s or
+        `ChildListener`s rapidly.
 
 ### 7.1.1
 -   Changes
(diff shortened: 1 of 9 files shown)


release_build_files/readme.md
```diff
@@ -576,6 +576,10 @@ code.
     -   General (Android): Use non-conflicting file names for embedded resources
         in Android builds. This fixes segfault crashes on old Android devices
         (Android 5 and below).
+    -   Database: Fixed a potential crash that can occur as a result of a race
```
DellaBitta, line 579: I beleive that 7.1.1 is already out with different release notes. Should this be in the next releasE?

DellaBitta (changes requested):

DellaBitta (approved):

PR #309, opened 2021-03-11
Fix is_command_installed(...) check for openssl

dconeybe:


dconeybe:
@jonsimantov is working on a better fix in another PR. I'll delete this one.

PR #311, opened 2021-03-11
[RemoteConfig] Fix the default cache value

cynthiajoan:
The default cache should be in milliseconds


remote_config/src/include/firebase/remote_config.h
```diff
@@ -115,8 +115,8 @@ enum ConfigSetting {
 };
 
 /// @brief The default cache expiration used by Fetch(), equal to 12 hours,
-/// in seconds.
-static const uint64_t kDefaultCacheExpiration = 60 * 60 * 12;
+/// in milliseconds.
+static const uint64_t kDefaultCacheExpiration = 60 * 60 * 12 * 1000;
```
vimanyu, line 119: nit: For clarity and consistency, should we name this "kDefaultCacheExpirationInMilliseconds" just like "kDefaultTimeoutInMilliseconds" ?

vimanyu (commented):
minor nit but looks good otherwise.

remote_config/src/include/firebase/remote_config.h
cynthiajoan: We could before, but I'm hesitating to change an api name now.

vimanyu: Good point. I guess we will leave it as it is for now then.

vimanyu (approved):
LGTM

PR #316, opened 2021-03-12
Firestore: Remove firestore_cc_library

alexames:
All exception-enabled builds should be done with --define=firestore_exceptions=1

PiperOrigin-RevId: 356770616


jonsimantov (commented):
I'm pretty sure that this file should just be completely removed from the repo.

alexames:
> I'm pretty sure that this file should just be completely removed from the repo.

Done.

var-const (approved):
I'm not sure how it looks from history perspective (this is a change that didn't exist in google3), but otherwise LGTM.

alexames:
Originally this PR was an export of a google3 change, but it was a change to a script that did blaze builds, which we don't need anymore, so at Jon's suggestion I just removed it.

PR #318, opened 2021-03-12
Firestore: Translate Java exceptions to C++ exceptions of the right type

alexames:
See go/firestore-exceptions for details.

PiperOrigin-RevId: 357237099

### DIFF
--- a/firestore/src/android/exception_android.cc
+++ b/firestore/src/android/exception_android.cc
@@ -1,5 +1,8 @@
 #include "firestore/src/android/exception_android.h"
 
+#include <stdexcept>
+
+#include "firestore/src/common/macros.h"
 #include "firestore/src/jni/env.h"
 #include "firestore/src/jni/loader.h"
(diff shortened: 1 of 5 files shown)


firestore/src/android/exception_android.cc
```diff
@@ -1,5 +1,8 @@
 #include "firestore/src/android/exception_android.h"
 
+#include <stdexcept>
```
jonsimantov, line 3: Does this build on stlport? (You can run CPP packaging on this PR to check.)

DellaBitta: Yep, I [ran a packaging workflow which failed](https://github.com/firebase/firebase-cpp-sdk/actions/runs/649538742) in the iOS tests, but so does the nightly build, unfortunately.  Looks like we have some mac ios scenarios to deflake.

var-const (approved):

PR #323, opened 2021-03-15
Allow integration_tests workflow to run against a packaged SDK.

jonsimantov:
This change updates the integration_tests workflow so that it can download a packaged SDK from GHA and build the integration tests against it, instead of against the SDK source.

With that change, we no longer need to duplicate the integration test code in the C++ packaging workflow.

Instead, the C++ packaging workflow directly invokes the integration_tests workflow at the end via the GitHub REST API.

(Note that because you cannot invoke another workflow with the same GITHUB_TOKEN that your workflow is already using due to a limitation in GitHub actions, the packaging workflow fetches a new GitHub App token to use for the request.)

### DIFF
--- a/.github/workflows/cpp-packaging.yml
+++ b/.github/workflows/cpp-packaging.yml
@@ -666,4 +666,8 @@ jobs:
         if [[ -z ${USE_EXPANDED_MATRIX} ]]; then
           USE_EXPANDED_MATRIX=0
         fi
-        python scripts/gha/trigger_workflow.py -t ${{ steps.generate-token.outputs.token }} -w integration_tests.yml -p test_packaged_sdk ${{ github.run_id }} -p use_expanded_matrix ${USE_EXPANDED_MATRIX} -v
+        verbose_flag=
+        if [[ -n "${{ github.event.inputs.verboseBuild }}" && "${{ github.event.inputs.verboseBuild }}" -ne 0 ]]; then
+          verbose_flag=-v
+        fi
+        python scripts/gha/trigger_workflow.py -t ${{ steps.generate-token.outputs.token }} -w integration_tests.yml -p test_packaged_sdk ${{ github.run_id }} -p use_expanded_matrix ${USE_EXPANDED_MATRIX} ${verbose_flag}
--- a/.github/workflows/cpp-packaging.yml
+++ b/.github/workflows/cpp-packaging.yml
@@ -655,15 +655,16 @@ jobs:
       run: |
         echo "USE_EXPANDED_MATRIX=1" >> $GITHUB_ENV
     - name: Generate token for GitHub API
+      uses: machine-learning-apps/actions-app-token@0.21
       uses: tibdex/github-app-token@v1
-      id: generate-token
+      id: get_token
       with:
-        app_id: ${{ secrets.WORKFLOW_TRIGGER_APP_ID }}
-        private_key: ${{ secrets.WORKFLOW_TRIGGER_APP_PRIVATE_KEY }}
+        APP_PEM: ${{ secrets.WORKFLOW_TRIGGER_APP_PRIVATE_KEY }}
+        APP_ID: ${{ secrets.WORKFLOW_TRIGGER_APP_ID }}
     - name: Use GitHub API to start workflow
       shell: bash
       run: |
         if [[ -z ${USE_EXPANDED_MATRIX} ]]; then
           USE_EXPANDED_MATRIX=0
         fi
-        python scripts/gha/trigger_workflow.py -t ${{ steps.generate-token.outputs.token }} -w integration_tests.yml -p test_packaged_sdk ${{ github.run_id }} -p use_expanded_matrix ${USE_EXPANDED_MATRIX} -v
+        python scripts/gha/trigger_workflow.py -t ${{ steps.get_token.outputs.app_token }} -w integration_tests.yml -p test_packaged_sdk ${{ github.run_id }} -p use_expanded_matrix ${USE_EXPANDED_MATRIX} -v
--- a/.github/workflows/cpp-packaging.yml
+++ b/.github/workflows/cpp-packaging.yml
@@ -677,4 +677,4 @@ jobs:
           verbose_flag=-v
         fi
         set -e
-        python scripts/gha/trigger_workflow.py -t ${{ steps.generate-token.outputs.token }} -w integration_tests.yml -p test_packaged_sdk ${{ github.run_id }} -p use_expanded_matrix ${USE_EXPANDED_MATRIX} -A ${verbose_flag}
+        python scripts/gha/trigger_workflow.py -t ${{ steps.generate-token.outputs.token }} -w integration_tests.yml -p test_packaged_sdk ${{ github.run_id }} -p use_expanded_matrix ${USE_EXPANDED_MATRIX} -s 10 -A ${verbose_flag}
--- a/.github/workflows/cpp-packaging.yml
+++ b/.github/workflows/cpp-packaging.yml
@@ -660,4 +660,4 @@ jobs:
         if [[ -z ${USE_EXPANDED_MATRIX} ]]; then
           USE_EXPANDED_MATRIX=0
         fi
-        python scripts/gha/trigger_workflow.py -t ${{ secrets.GITHUB_TOKEN }} -w integration_tests.yml -p test_packaged_sdk ${{ github.run_id }} -p use_expanded_matrix ${USE_EXPANDED_MATRIX}
+        python scripts/gha/trigger_workflow.py -t ${{ secrets.GITHUB_TOKEN }} -w integration_tests.yml -p test_packaged_sdk ${{ github.run_id }} -p use_expanded_matrix ${USE_EXPANDED_MATRIX} -v
--- a/.github/workflows/cpp-packaging.yml
+++ b/.github/workflows/cpp-packaging.yml
@@ -669,12 +669,13 @@ jobs:
     - name: Use GitHub API to start workflow
       shell: bash
       run: |
-        if [[ -z ${USE_EXPANDED_MATRIX} ]]; then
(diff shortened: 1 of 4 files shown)


database/integration_test/src/integration_test.cc
```diff
@@ -800,6 +800,8 @@ class LoggingValueListener : public firebase::database::ValueListener {
 };
 
 TEST_F(FirebaseDatabaseTest, TestAddAndRemoveListenerRace) {
+  SKIP_TEST_ON_MOBILE;
```
jonsimantov, line 803: Disabling this test, which crashes, to ensure that tests can be green.

DellaBitta:
The expanded matrix was meant to also expand the build configurations we use, including building across multiple XCode targets, which we do in some other workflows.  I thought this was important for our packaging workflows, too, which I thought was going to complicate the script.  But now I'm realizing that we only ship one version of the packaged SDK, so why would our CI build multiple Xcode versions, etc.  Is that your philosophy, too?

.github/workflows/cpp-packaging.yml
```diff
@@ -8,8 +8,6 @@ on:
     - cron: "0 9 * * *"  # 9am UTC = 1am PST / 2am PDT
   workflow_dispatch:
     inputs:
-      commitIdToPackage:
-        description: 'commit ID to package'
```
DellaBitta, line 12: Why was this taken out?  It seems useful. 

Is the intention that we would create a branch from a particular commit instead, and just run the workflow on that branch?

.github/workflows/cpp-packaging.yml
```diff
@@ -39,26 +37,9 @@ env:
   xcodeVersion: "12"
```
DellaBitta, line 37: You can remove this env now.

.github/workflows/cpp-packaging.yml
```diff
@@ -39,26 +37,9 @@ env:
...
-    steps:
-      - name: Use expanded matrix
-        if: github.event_name == 'schedule' || github.event.inputs.use_expanded_matrix == '1'
-        run: |
-          echo "USE_EXPANDED_MATRIX=1" >> $GITHUB_ENV
-      - id: export-result
-        run: |
-          echo "USE_EXPANDED_MATRIX: ${USE_EXPANDED_MATRIX}"
-          echo "::set-output name=use_expanded_matrix::${USE_EXPANDED_MATRIX}"
-
```
DellaBitta, line 57: Exporting the configuration into a job output us to normalize the inputs across all worfklow triggers.  Additionally duplicate checks like this:

`if github.event.inputs.verboseBuild != 0 && github.event.inputs.verboseBuild != "" `

become simply 

`if ${{ needs.prepare_matrix.outputs.verboseBuild }}`

across all of our workflows. 

The two prepare_matrix and log_inputs could be combined.

.github/workflows/cpp-packaging.yml
```diff
@@ -92,10 +66,13 @@ jobs:
           github.event.inputs.downloadPublicVersion == '' && github.event.inputs.downloadPreviousRun == ''
         run: echo "::warning ::Verbose build enabled."
 
+      - name: log if expanded matrix enabled
+        if: github.event_name == 'schedule' || github.event.inputs.use_expanded_matrix == '1'
```
DellaBitta, line 70: And one-off workflow checks, like this schedule check, become simply:

`if ${{ needs.prepare_matrix.outputs.use_expanded_matrix }}`

which I feel is easier to maintain.

.github/workflows/cpp-packaging.yml
```diff
@@ -668,88 +637,44 @@ jobs:
...
-    - name: Summarize build and test results
-      if: ${{ !cancelled() }}
+        echo "USE_EXPANDED_MATRIX=1" >> $GITHUB_ENV
+    - name: Generate token for GitHub API
+      # This step is necessary because the existing GITHUB_TOKEN cannot be used inside one workflow to trigger another.
+      # 
+      # Instead, generate a new token here, using our GitHub App's private key and App ID (saved as Secrets).
+      # 
+      # This method is preferred over the "personal access token" solution, as the GitHub App's scope is limited to just
+      # the firebase-cpp-sdk repository.
```
DellaBitta, line 663: I'm super duper glad this works. It reduces so much maintenance.

.github/workflows/integration_tests.yml
```diff
@@ -60,18 +62,27 @@ jobs:
         if: github.event_name == 'schedule' || github.event.inputs.use_expanded_matrix == '1'
         run: |
           echo "EXPANDED_MATRIX_PARAM=-e=1" >> $GITHUB_ENV
+          echo "::warning ::Expanded test matrix enabled."
```
DellaBitta, line 65: Tip: I used to do this, too, but the fact that the `Use expanded matrix` step is grayed out during the GHA jobs steps, or not grayed out depending on the `if` clause above, also tells the story of the status of the expanded matrix.

scripts/gha/trigger_workflow.py
```diff
@@ -0,0 +1,138 @@
...
+    # the contents are not valid JSON.
+    error_message = re.search(r'"message": "([^"]+)"', run_output).group(1)
+    print('%sFailed to trigger workflow %s: %s' % (
+      '::error ::' if args.in_github_action else '', args.workflow, error_message))
+    return(-1)
+
+  print('Success!')
+  time.sleep(5)  # Give a few seconds for the job to become queued.
+  # Unfortunately, the GitHub REST API doesn't return the new workflow's run ID.
+  # Query the list of workflows to find the one we just added.
```
DellaBitta, line 87: :(

scripts/gha/trigger_workflow.py
```diff
@@ -0,0 +1,138 @@
...
+      print(run_output)
+    # Super quick and dirty way to get the message text since the appended status code means that
+    # the contents are not valid JSON.
+    error_message = re.search(r'"message": "([^"]+)"', run_output).group(1)
+    print('%sFailed to trigger workflow %s: %s' % (
+      '::error ::' if args.in_github_action else '', args.workflow, error_message))
+    return(-1)
+
+  print('Success!')
+  time.sleep(5)  # Give a few seconds for the job to become queued.
```
DellaBitta, line 85: Given how slow GHA have been over the last few weeks, we might want to make this sleep a bit longer.

DellaBitta (commented):
Some suggestions.  Let me know what you think.

.github/workflows/cpp-packaging.yml
jonsimantov: Yes, because you can only trigger a workflow on a branch or tag, not on any old commit hash, I wanted to ensure that we are definitely triggering the integration_tests workflow on the same git revision as this script is running on. This removes the ambiguity. If you need to package any old commit, make a branch or tag for it first.

.github/workflows/cpp-packaging.yml
jonsimantov: It's still needed, because it controls what xcode version the SDK is built with.

jonsimantov:
> The expanded matrix was meant to also expand the build configurations we use, including building across multiple XCode targets, which we do in some other workflows. I thought this was important for our packaging workflows, too, and in my head this would have complicated the script. But now I'm realizing that we only ship one version of the packaged SDK, so why would our CI build multiple Xcode versions of it, etc. Is this your philosophy, too?

That was my thought. We build and ship the binary SDK in a single set of configurations. I am happy to keep the SDK packaging matrix in this workflow, and the test matrix in the script. The packaging matrix should never change in the normal course of development.

.github/workflows/cpp-packaging.yml
jonsimantov: I agree, but I'd like to do this cleanup in a followup.

.github/workflows/cpp-packaging.yml
jonsimantov: I agree, but I'd like to do this cleanup in a followup.

.github/workflows/cpp-packaging.yml
jonsimantov: It's a neat workaround! I'm glad we don't have to use anyone's personal token.

.github/workflows/integration_tests.yml
jonsimantov: I removed it from this workflow. I would like to log a few things but we'll take some time to do it thoughtfully.

scripts/gha/trigger_workflow.py
jonsimantov: Added a customizable sleep time, at least. Or maybe I should add a retry?

scripts/gha/trigger_workflow.py
jonsimantov: Added a customizable sleep time, at least. Or maybe I should add a retry?

scripts/gha/trigger_workflow.py
jonsimantov: Yeah, I'm so mad they just return an empty response. 😛

.github/workflows/cpp-packaging.yml
DellaBitta: The code  `sudo xcode-select -s /Applications/Xcode_${{ env.xcodeVersion }}.app/Contents/Developer` has been removed below.  Is there something else in here that uses it?  Some silent GHA observance of the variable?

.github/workflows/cpp-packaging.yml
DellaBitta: ok.

scripts/gha/trigger_workflow.py
DellaBitta: It's fine for now.  We can always tune it later.

DellaBitta (approved):
Please double check on that xcodeVersion, but otherwise a-ok.

PR #327, opened 2021-03-16
Deflake: Cache NDK Downloads

DellaBitta:
There have been many workflow runs which have failed to download the NDK. I'mt not sure if this is due to its size or if the GHA Vms have a bad connection, etc, but more often than not when theres a failed download, it's the NDK.

Adding steps to cache the NDK to the Integration, Packaging and Android workflows. The NDK cache is based on the host OS and is based on the path of the android/install_prereqs.py script.

Includes some indentation changes to the Android Prereqs script, but no logical changes.

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -76,7 +76,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: /tmp/android-ndk-r16b
-          key: android-ndk-${{ matrix.os }}
+          key: android-ndk-${{ matrix.os }}-r16b
 
       - name: Install prerequisites
         shell: bash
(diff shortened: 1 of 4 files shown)


.github/workflows/android.yml
```diff
@@ -70,6 +70,13 @@ jobs:
       - name: Add msbuild to PATH
         if: startsWith(matrix.os, 'windows')
         uses: microsoft/setup-msbuild@v1.0.2
+      
+      - name: Cache NDK
+        id: cache_ndk
+        uses: actions/cache@v2
+        with:
+          path: /tmp/android-ndk-r16b
+          key: android-ndk-${{ matrix.os }}
```
vimanyu, line 79: Do you think we should encode the version number in the key too? If we ever expand our matrix to test/build on different NDKs, it might be helpful.

vimanyu (approved):
Minor comment about the key name but LGTM otherwise!

.github/workflows/android.yml
DellaBitta: Done!

vimanyu (approved):

PR #339, opened 2021-03-19
Trigger integration tests by setting a "tests-requested" label.

jonsimantov:
This PR adds some label-based triggers for integration tests, which allow the integration tests to be requested on a PR by setting the "tests-requested" label.

There are two modes for requesting tests, "quick" and "full":

"quick" tries to be clever and figure out a subset of APIs (and even platforms) to test. This heuristic is really simple right now (it just looks at directory names and, for platform, filenames) and if there is *any* ambiguity it will just run the normal suite of tests.

"full" just runs all the tests. Currently this is set to use an expanded test matrix, though this could be changed in the future.

The test workflow handles updating the labels and posting/updating a comment with the status of the integration tests as they run. All of the jobs that mess with labels and comments are skipped if the workflow wasn't run from the PR label.

(See #337—which won't be merged—for an example of "quick" mode in action, autodetecting to just run a subset of tests.)

### DIFF
--- a/.github/workflows/check-labels.yml
+++ b/.github/workflows/check-labels.yml
@@ -14,9 +14,9 @@ jobs:
     # This check fails if integration tests are queued, in progress, or failed.
     runs-on: ubuntu-latest
     steps:
-    - uses: danielchabr/pr-labels-checker@v3.0
+    - uses: docker://agilepathway/pull-request-label-checker:latest
       # Uncomment the below line if you want this check to only be on the main branch.
       # if: github.base_ref == 'main'
       with:
-        hasNone: "${{ env.statusLabelInProgress }},${{ env.statusLabelFailed }},${{ env.triggerLabelFull }},${{ env.triggerLabelQuick }}"
-        githubToken: ${{ github.token }}
+        none_of: "${{ env.statusLabelInProgress }},${{ env.statusLabelFailed }},${{ env.triggerLabelFull }},${{ env.triggerLabelQuick }}"
+        repo_token: ${{ github.token }}
--- a/.github/workflows/check-labels.yml
+++ b/.github/workflows/check-labels.yml
@@ -12,7 +12,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: danielchabr/pr-labels-checker@v3.0
-      if: github.base_ref == 'main'
+      # Uncomment the below line if you want this check to only be on the main branch.
+      # if: github.base_ref == 'main'
       with:
         hasNone: "${{ env.statusLabelInProgress }},${{ env.statusLabelFailed }}"
         githubToken: ${{ github.token }}
--- a/.github/workflows/check-labels.yml
+++ b/.github/workflows/check-labels.yml
@@ -1,7 +1,7 @@
 name: Check PR Labels
 on:
   pull_request:
-    types: [opened, reopened, synchronize]
+    types: [opened, reopened, synchronize, labeled, unlabeled]
 
 jobs:
   check_labels:
--- a/.github/workflows/check-labels.yml
+++ b/.github/workflows/check-labels.yml
@@ -12,7 +12,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: danielchabr/pr-labels-checker@v3.0
-      if: github.base_ref == "main"
+      if: github.base_ref == 'main'
       with:
         hasNone: "${{ env.statusLabelInProgress }},${{ env.statusLabelFailed }}"
         githubToken: ${{ github.token }}
--- a/.github/workflows/check-labels.yml
+++ b/.github/workflows/check-labels.yml
@@ -8,7 +8,7 @@ jobs:
   check_labels:
     runs-on: ubuntu-latest
     steps:
-    - uses: danielchabr/pr-labels-checker@v3
+    - uses: danielchabr/pr-labels-checker@v3.0
       if: github.base_ref == 'main'
       with:
         hasNone: 'tests: failed, tests: in-progress'
--- a/.github/workflows/check-labels.yml
+++ b/.github/workflows/check-labels.yml
@@ -8,7 +8,7 @@ env:
   statusLabelFailed: "tests: failed"
 
 jobs:
-  check_labels:
+  check_integration_test_success:
     runs-on: ubuntu-latest
     steps:
     - uses: danielchabr/pr-labels-checker@v3.0
--- a/.github/workflows/check-labels.yml
+++ b/.github/workflows/check-labels.yml
@@ -11,6 +11,7 @@ env:
 
 jobs:
   check_integration_test_labels:
+    # This check fails if integration tests are queued, in progress, or failed.
     runs-on: ubuntu-latest
     steps:
     - uses: danielchabr/pr-labels-checker@v3.0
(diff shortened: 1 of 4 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -57,17 +152,27 @@ jobs:
...
         run: |
           echo "EXPANDED_MATRIX_PARAM=-e=1" >> $GITHUB_ENV
-          
+
       - id: export-result
         # e.g. 'ubuntu-latest,macos-latest' -> '["ubuntu-latest","macos-latest"]'
         run: |
-          echo "::set-output name=matrix_platform::$( python scripts/gha/print_matrix_configuration.py -w integration_tests ${EXPANDED_MATRIX_PARAM} -k platform -o "${{github.event.inputs.platforms}}" )"
-          echo "::set-output name=matrix_os::$( python scripts/gha/print_matrix_configuration.py -w integration_tests ${EXPANDED_MATRIX_PARAM} -k os -o "${{github.event.inputs.operating_systems}}" )"
+          if [[ "${{ needs.check_trigger.outputs.requested_tests }}" == "auto" ]]; then
```
DellaBitta, line 165: Suggestion:

These two distinct blocks of three echo below are nearly identical except for the parameter ` --auto_diff ${{github.event.pull_request.base.sha}}`.

I recommend adding a new named step above this one. Triggers on if:  needs.check_trigger.outputs.requested_tests  == "auto".  

If it's true, `run:  echo "AUTO_DIFF_PARAM=--auto_diff ${{github.event.pull_request.base.sha}}" >> $GITHUB_ENV`

Then below have only one block of three echos 

```
echo "::set-output name=apis::$( python scripts/gha/print_matrix_configuration.py -c -w integration_tests -k apis -o "${{github.event.inputs.apis}}" ${AUTO_DIFF_PARAM})"

echo "::set-output name=matrix_platform::$( python scripts/gha/print_matrix_configuration.py -w integration_tests ${EXPANDED_MATRIX_PARAM} -k platform -o "${{github.event.inputs.platforms}}" ${AUTO_DIFF_PARAM})"

echo "::set-output name=matrix_os::$( python scripts/gha/print_matrix_configuration.py -w integration_tests ${EXPANDED_MATRIX_PARAM} -k os -o "${{github.event.inputs.operating_systems}}" ${AUTO_DIFF_PARAM})"
```

.github/workflows/integration_tests.yml
```diff
@@ -248,3 +383,54 @@ jobs:
           if [[ "${{ job.status }}" != "success" ]]; then
             exit 1
           fi
+
+  add_success_label:
```
DellaBitta, line 387: It seems odd that the check failure above is part of the tests job, but the two label operations below are their own job.  Could the blow jobs be added as steps to the above job, just like the add failure label and comment steps?  They all could share the same get time, a step which then wouldn't need any conditionals.

scripts/gha/print_matrix_configuration.py
```diff
@@ -181,6 +185,82 @@ def print_value(value):
...
+
+  Remove entries from the list based on what we observe in the
+  source tree, relative to the given base branch."""
+  file_list = set(subprocess.check_output(['git', 'diff', '--name-only', auto_diff]).decode('utf-8').rstrip('\n').split('\n'))
+  if parm_key == 'apis':
+    custom_triggers = {
+      # Ones set to None are ignored.
+      "external": None,
+      "release_build_files": None,
+      # Ones set to a list of APIs will trigger those APIs.
```
DellaBitta, line 199: not quite sure what this comment means.

.github/workflows/integration_tests.yml
```diff
@@ -57,17 +152,27 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
+          fetch-depth: 0
```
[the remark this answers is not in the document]
DellaBitta: Oh does this unlock the autodetect via diff?

DellaBitta (commented):
Questions, suggestions, clarifications.

.github/workflows/integration_tests.yml
jonsimantov: The failure is part of the tests job because we mark failure if even a single matrix entry fails. Whereas we mark success only once the entire matrix is complete, if there were no failures..

.github/workflows/integration_tests.yml
jonsimantov: Good idea, will do.

.github/workflows/integration_tests.yml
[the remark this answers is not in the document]
jonsimantov: Actually it might not be required, previously I was using the base branch and not the base sha... let me see if I can remove it.

.github/workflows/integration_tests.yml
```diff
@@ -39,12 +42,104 @@ on:
         required: true
       test_packaged_sdk:
         description: 'Optional: Packaging run # to build against?'
+env:
+  triggerLabelPrefix: "tests-requested: "
+  triggerLabelFull: "tests-requested: full"
```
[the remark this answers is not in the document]
jonsimantov: Let's try it!

.github/workflows/check-labels.yml
```diff
@@ -0,0 +1,22 @@
...
+  statusLabelFailed: "tests: failed"
+
+jobs:
+  check_integration_test_labels:
+    # This check fails if integration tests are queued, in progress, or failed.
+    runs-on: ubuntu-latest
+    steps:
+    - uses: docker://agilepathway/pull-request-label-checker:latest
+      # Uncomment the below line if you want this check to only be on the main branch.
+      # if: github.base_ref == 'main'
```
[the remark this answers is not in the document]
jonsimantov: SG

.github/workflows/integration_tests.yml
jonsimantov: I originally had them combined but it actually didn't quite work - using success() / failure() to check the status of the entire workflow only only works in an "if:" at the job level, if you call them in an "if:" at the steps level, it only checks if the *current* job succeeded or failed. So the triggers need to be jobs-based.

scripts/gha/print_matrix_configuration.py
jonsimantov: Clarifying.

.github/workflows/integration_tests.yml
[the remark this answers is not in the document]
jonsimantov: actually, on second thought, I'd like not to do this - I want to make sure there is no chance of a text mismatch (extra spaces, bad punctuation, etc.) so i'd like all the labels to be explicitly specified.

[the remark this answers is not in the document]
DellaBitta: Ok.

.github/workflows/integration_tests.yml
DellaBitta: Hm, interesting.

.github/workflows/integration_tests.yml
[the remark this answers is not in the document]
jonsimantov: Looks like the fetch-depth option is required - otherwise we can't get the base branch sha.

DellaBitta (approved):
Guess you need the 0 fetch after all, eh? :)

:+1:

PR #344, opened 2021-03-22
Bugfix Database Large String Crash on Set on Desktop

DellaBitta:
A patch to the uWebsockets implementation to avoid a crash when SSL_write returns SSL_ERROR_WANT_WRITE.  The previous code would delegate this error state to a non existent async handler.  This new implementation re-attempts the write operation instead.

Includes integration test changes to upload and download a 1mb string.   I hand-tested larger than this (10mb) but chose a reduced size to run regularly in our CI due to RTDB usage limits combined with our expanded test matrix potentially running this numerous times in short order.

This implementation uses a python script to patch uWebSockets/src/Socket.h during the cmake configuration process. I tried a few other alternatives before resolving to this version, and I'm open for comments on alternatives:

- hosted our own version of the entire Socket.h file and copied our version over the downloaded version using cmake file commands. I thought this clunky - hosting an entire sourcefile from another repo seemed awkward, and our small patch wasn't self-evident unless someone read the whole file.
- Created a custom proprietary patch file with the startline, no. of lines to replace, and our custom code to embed in Socket.h. I thought this, too, wasn't a great option.  This changed was intended to be a quick fix and so the patch file didnt have a robust or defined  format. Additionally. I didn't like disassociating the logic from the patch source. Finally, the use of a file suggested that the script could be executed for other patches, but really the. script isn't built to be flexible.
- Investigated [PyPatch](https://pypi.org/project/patch/) which applies git patch files,  an industry standard format. However, I didn't want to burden our users with another python dependency for such a small change.


scripts/patch_websockets.py
```diff
@@ -0,0 +1,91 @@
...
+  ignore_lines_count = 0
+  with open(FLAGS.file,'w') as sockets_file:
+    index = 0
+    for line in lines:
+      if(ignore_lines_count != 0):
+        ignore_lines_count -= 1
+        index+=1
+        continue
+
+      if( index == REPLACE_BLOCK_START_LINE ):
```
vimanyu, line 78: It might be worth doing a simple check for the file being patched. If versions of websockets change, this will break. Couple of options could be to not use line numbers for block replacements or just verify the block you are targetting has the expected values.

vimanyu (changes requested):

scripts/patch_websockets.py
chkuang-g: Yes, I think this script should throw exception if the Socket.h is changed.
And in cmake file, the build should fail when this python script throw exception.

scripts/patch_websockets.py
```diff
@@ -0,0 +1,91 @@
...
+from absl import app
+from absl import flags
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_string("file", None, "Sockets.h file to patch.")
+
+# The line of Socket.h that we intend to start overwriting, and the 
+# number of lines to overwite.
+REPLACE_BLOCK_START_LINE = 308
```
chkuang-g, line 27: I'm a bit confused which part of the code are you trying to replace in Sockets.h.
https://github.com/uNetworking/uWebSockets/blob/4d94401b9c98346f9afd838556fdc7dce30561eb/src/Socket.h#L308

Line 308-325 does not look right to be replaced.

chkuang-g: I think you are actually replacing 309-326. Worth changing the description and adding the url to uWebSocket code for clarification.

scripts/patch_websockets.py
chkuang-g: To simplify the check, I think you can basically check if the commit in cmake/external/uWebSockets.cmake is 4d94401b9c98346f9afd838556fdc7dce30561eb

Every time when we bump up uWebSockets version, we need to check if this patch is still valid.

scripts/patch_websockets.py
```diff
@@ -0,0 +1,91 @@
...
+  "              bool continue_loop = true;\n",
+  "              do {\n",
+  "                sent = SSL_write(ssl, message->data, message->length);\n",
+  "                static int bytes_written = 0;\n",
+  "                bytes_written += sent;\n",
+  "                if (sent == (ssize_t) message->length) {\n",
+  "                  wasTransferred = false;\n",
+  "                  return true;\n",
+  "                }\n",
+  "                else if (sent < 0) {\n",
```
chkuang-g, line 43: nit: else should be behind {

scripts/patch_websockets.py
```diff
@@ -0,0 +1,91 @@
...
+  "                  case SSL_ERROR_WANT_READ:\n",
+  "                    continue_loop = false;\n",
+  "                    break;\n",
+  "                  case SSL_ERROR_WANT_WRITE:\n",
+  "                    break;\n",
+  "                  default:\n",
+  "                    return false;\n",
+  "                }\n",
+  "              }\n",
+  "            }  while (continue_loop);\n"
```
chkuang-g, line 54: Is it possible that SSL_ERROR_WANT_WRITE occurs forever and this loop never ends?
Perhaps we can set a retry limit like 100 times?

scripts/patch_websockets.py
```diff
@@ -0,0 +1,91 @@
...
+REPLACE_BLOCK_OVERWRITE_LENGTH = 17
+
+# The code which fixes the crash which will replace the existing
+# error handling implementation with one that retries the writes
+# when the SSL encoding buffer is full.
+REPLACE_BLOCK_CODE_LINES = [
+  "              bool continue_loop = true;\n",
+  "              do {\n",
+  "                sent = SSL_write(ssl, message->data, message->length);\n",
+  "                static int bytes_written = 0;\n",
```
chkuang-g, line 37: This variable does not seems to be used later in the code.

scripts/patch_websockets.py
```diff
@@ -0,0 +1,91 @@
...
+  "                  wasTransferred = false;\n",
+  "                  return true;\n",
+  "                }\n",
+  "                else if (sent < 0) {\n",
+  "                  switch (SSL_get_error(ssl, sent)) {\n",
+  "                  case SSL_ERROR_WANT_READ:\n",
+  "                    continue_loop = false;\n",
+  "                    break;\n",
+  "                  case SSL_ERROR_WANT_WRITE:\n",
+  "                    break;\n",
```
chkuang-g, line 49: It looks like you removed the entire changePoll() logic.
https://github.com/uNetworking/uWebSockets/blob/4d94401b9c98346f9afd838556fdc7dce30561eb/src/Socket.h#L318

Do you know why that is causing the crash? 
Also, should we replace the part without SSL?
https://github.com/uNetworking/uWebSockets/blob/4d94401b9c98346f9afd838556fdc7dce30561eb/src/Socket.h#L341

scripts/patch_websockets.py
DellaBitta: I'll definitely add the check.   Smart move.  

Just a note: there are few verisons to bump to.  The library was rewritten a few versions after the one wer're using.  I think we should have a longer horizon task on our backlog to replace this uWebSockets implementation with either their latest and greatest, or a different library, since support for the version that we're using was sunset nearly 3 years ago.

scripts/patch_websockets.py
DellaBitta: I tried doing just that. First off, I found it fuzzy to estimate a cap on the number of iterations of the loop, since the number of iterations will depend on indeterminate factors like encoding time, network usage, etc.  The loop count varied wildly. Addtionally returning false from this function to signify an error causes the uWebsockets implementation to hang, so even if we decided to opt out of the write then their library doesn't handle it well.

At any rate, an idefinite loop  shouldn't occur.  This is a buffer consumption issue where the underlying SSL process funnels the data out of the full SSL buffer and into the socket.  Eventually. it will empty unless there's an issue with the socket itself (timeout, broken pipe, etc) and  and the SSL_write operation will return an error other than WANT_WRITE.  Of course, the uWebsockets implementation then hangs indefinitely again, which is the same thing as an infinite loop anyway.

Generally speaking this version of uWebSockets doesn't seem very robust. It obviously. doesn't handle errors well. I think of this PR as a simpy a temproary patch of buggy code to unblock our users.  Really we should replace the entire implementation, but that's a signifcant amount of work that. should be scheduled.

scripts/patch_websockets.py
DellaBitta: Oh right, thanks. Yes I was using this to debug stuff.  I'l remove it.

scripts/patch_websockets.py
DellaBitta: The changePoll logic was errant. It attempted to delegate to pointer to an Aysnc object that didn't exist, and therefore this is what was actually causing the crash.

I also don't believe the intention of their implemetnation was correct.  The socket shouldn't be set into a read state when SSL_WANT_WRITE is returned.  I've written several MQTT clients which wrapped SLL and network stakcs  I've never seen anything switching to read when the write buffer was full.

Instead code should either loop back and write again. On embeded devices we would throw in a delay and attempt on the next tick, or call select on the underlying socket descriptor to check when its ready to write to again.  This was only an inference that the SSL wrapper had flushed all of its buffered data. to the native socket, though.   

However, in this code we no longer have the socket descriptor to run a select on, and even if we did, we would need to write more code to do this correctly across the the different host opeating systems.  And since this code is  run by us on desktop platforms we can leverage the fact that the SLL implementation  runs in its own process which the OS will timeslice to-- looping on the write should be fine since it loops for only a few moments.   If we were porting this to other devices with slower CPUs or OSes that have scheduling problems then i would suggest extending their implementation to use select, but again the fix would be a lot larger and we'd be heavily meddling with the code of an abandoned implementation, which seems frought with danger in itself.

Also I don't believe we ever use this code in the non ssl mode, so I didn't patch that.

scripts/patch_websockets.py
DellaBitta: Done

scripts/patch_websockets.py
DellaBitta: Done.

scripts/patch_websockets.py
```diff
@@ -0,0 +1,101 @@
...
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_string("file", None, "Sockets.h file to patch.")
+flags.DEFINE_string("cmakefile", None, "cmake file which downloaded uWebSockets")
+
+# The line of Socket.h that we intend to start overwriting, and the 
+# number of lines to overwite. 
+# See: https://github.com/uNetworking/uWebSockets/blob/e402f022a43d283c4fb7c809332951c3108d6906/src/Socket.h#L309
+REPLACE_BLOCK_START_LINE = 309
```
jonsimantov, line 29: Maybe add a sanity check to ensure that these are the correct lines?

jonsimantov:
Why not use the standard POSIX patch utility? (If the Windows runners have it, that is.)

scripts/patch_websockets.py
```diff
@@ -0,0 +1,101 @@
...
+REPLACE_BLOCK_OVERWRITE_LENGTH = 17
+
+# The code which fixes the crash which will replace the existing
+# error handling implementation with one that retries the writes
+# when the SSL encoding buffer is full.
+REPLACE_BLOCK_CODE_LINES = [
+  "              bool continue_loop = true;\n",
```
chkuang-g, line 36: It makes it clear that this part of the code is overridden by patch_websockets.py, perhaps to add comments like the following in Sockets.h: 
// BEG Patched by scripts/patch_websockets.py
// END Patched by scripts/patch_websockets.py

DellaBitta:
I'm not aware of a native patch utility that comes with Windows, and I have both only our GHA but our individual developer out-of-the-box local windows build experience in mind.  I'd rather not add another tool that people would have to download an install just for this, but if you think that it's worth it then let me know.

scripts/patch_websockets.py
```diff
@@ -0,0 +1,101 @@
...
+  ignore_lines_count = 0
+  with open(FLAGS.file,'w') as sockets_file:
+    index = 0
+    for line in lines:
+      if(ignore_lines_count != 0):
+        ignore_lines_count -= 1
+        index+=1
+        continue
+
+      if( index+1 == REPLACE_BLOCK_START_LINE ):
```
chkuang-g, line 87: nit: I think you can initiate index from 1 so that you do not need to do `index+1` here.
That is:
index = 1
...
...
    if (index == REPLACE_BLOCK_START_LINE) :
...

scripts/patch_websockets.py
DellaBitta: Done.

scripts/patch_websockets.py
DellaBitta: Fixed this.

scripts/patch_websockets.py
DellaBitta: Done.

scripts/patch_websockets.py
DellaBitta: Check added.  Great idea.

chkuang-g (dismissed):

database/integration_test/src/integration_test.cc
```diff
@@ -59,6 +59,7 @@ using testing::PrintToString;
 using testing::UnorderedElementsAre;
 
 const char kIntegrationTestRootPath[] = "integration_test_data";
+const uint32_t kLargeWriteStringLength = 1024 * 1024; // 1 Megabyte.
```
jonsimantov, line 62: nit: size_t is your friend!

scripts/patch_websockets.py
jonsimantov: Probably also worth adding handling that doesn't return an error if the file is already patched, but just accepts the file.

cmake/external_rules.cmake
```diff
@@ -117,6 +117,10 @@ function(download_external_sources)
       file(INSTALL "${PROJECT_SOURCE_DIR}/cmake/external/leveldb.cmake"
            DESTINATION "${PROJECT_BINARY_DIR}/external/src/firestore/cmake/external")
     endif()
+
```
jonsimantov, line 120: Since we are now patching a couple of different things (this, boringssl, leveldb's cmake file), if there's ever even *one* more I'd sure love to see this changed to a generic patcher script, maybe with a json config file or something. :)

jonsimantov (dismissed):

vimanyu (dismissed):

jonsimantov (dismissed):

jonsimantov:
Feel free to add the readme item here or in a subsequent PR.

cmake/external_rules.cmake
```diff
@@ -117,6 +117,16 @@ function(download_external_sources)
...
+    execute_process(
+      COMMAND "python" "${PROJECT_SOURCE_DIR}/scripts/patch_websockets.py" 
+              "-file" "${PROJECT_BINARY_DIR}/external/src/uWebSockets/src/Socket.h"
+              "-cmakefile" "${PROJECT_SOURCE_DIR}/cmake/external/uWebSockets.cmake" 
+      RESULT_VARIABLE STATUS)
+      if( STATUS AND NOT STATUS EQUAL 0 )
+        message(FATAL_ERROR "FAILED to patch uWebsockets/src/Socket.h")
+        message(FATAL_ERROR "see cmake/eterna_rules.cmake.")
```
jonsimantov, line 128: nit: external.

(though "eterna rules" sounds neat!)

jonsimantov (approved):

PR #346, opened 2021-03-25
Feature/fix android windows builds

vimanyu:
Fixes a series of issues that were causing Windows Android builds to error out.
1. flatbuffers was not getting downloaded as an external library and gradle/cmake build was failing to find cmake function `build_flatbuffers`
2. Android Gradle CMake was invoking system level CMake as a child process to download and build external libraries. In addition to using the same cmake as the parent process, some CMake options had to be passed through to the child process to make it work correctly. For example, Android Gradle CMake prefers to use the generator "Android Gradle - Ninja" and ninja as the make program.
3. strings.exe is a special utility tool required on Android/Windows. Our android build script was downloading strings.exe to the repo root but since the PATH was not modified, this downloaded tool wasn't found. Additionally, there is another conflicting strings.exe on Github runners which meant that we have to prepend our repo root directory to the PATH variable on Github runners.
4. strings.exe prompts for accepting license on first invocation. Found an undocumented flag `-accepteula` that accepts the license which is run in a dummy github workflow step. Just running with this fails exits the command with error code but we ignore that error.


.github/workflows/android.yml
```diff
@@ -83,6 +83,27 @@ jobs:
...
+        if: startsWith(matrix.os, 'windows')
+        continue-on-error: true
+        shell: bash
+        run: |
+          # This invocation is just to accept the license. Otherwise the command
+          # fails during build process. Also, running a command with just this
+          # flag makes the process exit with code 127 and stops the workflow.
+          # We ignore the error and continue with the rest of the steps.
+          strings -accepteula
+
```
DellaBitta, line 106: Nit: Our workflow logs are getting pretty verbose.  I think it would be better if these were in the same step. WDYT?

cmake/external_rules.cmake
```diff
@@ -88,7 +91,7 @@ function(download_external_sources)
   endif()
 
   execute_process(
-    COMMAND ${ENV_COMMAND} cmake --build . -- ${cmake_build_args}
+    COMMAND ${ENV_COMMAND} ${CMAKE_COMMAND} --build . -- ${cmake_build_args}
```
DellaBitta, line 94: Should there really be a space between `--` and `${cmake_build_args}` ?

scripts/gha/print_matrix_configuration.py
```diff
@@ -85,12 +85,12 @@
 
   "android": {
     "matrix": {
-      "os": ["ubuntu-latest", "macos-latest"],
+      "os": ["ubuntu-latest", "macos-latest", "windows-latest"],
```
DellaBitta, line 88: :+1:

DellaBitta (approved):
Approved, but also please check my comments.

.github/workflows/android.yml
jonsimantov: ProTip: in bash, you can ignore the error by just running as: strings -accepteula || true

cmake/external_rules.cmake
jonsimantov: I'm just not clear on why we use the "--" at all - typically that's how you separate sets of arguments, for example the arguments afterwards are sent to a subprocess, or are considered a list of filenames even if they start with a "-". But why is that necessary here?

(For example, if you have a filename called "-q", you can't normally run "grep hello -q" because grep will interpret "-q" as an argument. So you must do "grep hello -- -q" which tells grep that everything following the "--" is a filename. But I'm still not sure why that's required for this cmake call.)

DellaBitta: The more you know.

.github/workflows/android.yml
vimanyu: Unfortunately they can't be in the same step. Github's PATH update takes effect in the following steps/actions. There is no other Windows specific  tasks in this workflow that we can place this or the update in.

cmake/external_rules.cmake
vimanyu: Good catch. I didn't spot this and am not sure why this exists. I substituted hard coded cmake to using a variable.

jonsimantov: What's with the "--" though?

jonsimantov (approved):

cmake/external_rules.cmake
vimanyu: The only change that I made is "cmake" converted to "${CMAKE_COMMAND}". Will investigate why the "-- ${cmake_build_args}" exists in another PR as I don't want to introduce new changes for this Windows/Android task.

vimanyu:
Integration tests builds are also fixed now. 
https://github.com/firebase/firebase-cpp-sdk/runs/2221512221?check_suite_focus=true

The only error was with the Storage test which is unrelated to issues with building on Windows for Android. 
```
[  FAILED  ] FirebaseStorageTest.TestLargeFilePauseResumeAndDownloadCancel
```

PR #357, opened 2021-04-06
Add 3rd Party Notices File for Packaged Releases

DellaBitta:
Performed a survey of external dependencies, those that we we use as submodules + those downloaded during our build process, and gathered their copyright and distribution notices.  Placed them in a single file in the build release file directory for distribution in our C++ SDK build archive.

Note: I removed the googletest notice as I don't believe we ship binaries that include the test sources.

### DIFF
--- a/release_build_files/NOTICES
+++ b/release_build_files/NOTICES
@@ -1,0 +1,1093 @@
+================ 
+Firebase C++ SDK
+================
+                                 Apache License
+                           Version 2.0, January 2004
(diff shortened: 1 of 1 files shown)


DellaBitta:
I'm running [a Packaging action](https://github.com/firebase/firebase-cpp-sdk/actions/runs/723020172) to ensure that the notices file is added to the .zip file:

cynthiajoan (approved):

PR #358, opened 2021-04-06
Change GKLocalPlayer to be runtime referenced

cynthiajoan:
To fix https://github.com/firebase/quickstart-unity/issues/954

### DIFF
--- a/auth/src/ios/credential_ios.mm
+++ b/auth/src/ios/credential_ios.mm
@@ -169,8 +169,8 @@ Future<Credential> GameCenterAuthProvider::GetCredential() {
   /**
    Linking GameKit.framework without using it on macOS results in App Store rejection.
    Thus we don't link GameKit.framework to our SDK directly. `optionalLocalPlayer` is used for
-   checking whether the APP that consuming our SDK has linked GameKit.framework. If not, a
-   `GameKitNotLinkedError` will be raised.
+   checking whether the APP that consuming our SDK has linked GameKit.framework. If not, will
+   complete with kAuthErrorInvalidCredential error.
    **/
   GKLocalPlayer *_Nullable optionalLocalPlayer = [[NSClassFromString(@"GKLocalPlayer") alloc] init];
 
@@ -207,8 +207,8 @@ bool GameCenterAuthProvider::IsPlayerAuthenticated() {
   /**
    Linking GameKit.framework without using it on macOS results in App Store rejection.
    Thus we don't link GameKit.framework to our SDK directly. `optionalLocalPlayer` is used for
-   checking whether the APP that consuming our SDK has linked GameKit.framework. If not, a
-   `GameKitNotLinkedError` will be raised.
+   checking whether the APP that consuming our SDK has linked GameKit.framework. If not, 
+   early out.
    **/
   GKLocalPlayer *_Nullable optionalLocalPlayer = [[NSClassFromString(@"GKLocalPlayer") alloc] init];
   // If the GameKit Framework isn't linked - early out.
--- a/auth/src/ios/credential_ios.mm
+++ b/auth/src/ios/credential_ios.mm
@@ -175,7 +175,7 @@ Future<Credential> GameCenterAuthProvider::GetCredential() {
   GKLocalPlayer *_Nullable optionalLocalPlayer = [[NSClassFromString(@"GKLocalPlayer") alloc] init];
 
   // Early-out if GameKit is not linked
-  if (optionalLocalPlayer == nullptr) {
+  if (!optionalLocalPlayer) {
     future_api->Complete(handle, kAuthErrorInvalidCredential,
                        "GameCenter authentication is unavailable - missing GameKit capability.");
     return MakeFuture(future_api, handle);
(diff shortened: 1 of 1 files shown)


auth/src/ios/credential_ios.mm
```diff
@@ -197,11 +204,18 @@ @implementation PhoneListenerDataObjC
...
+   early out.
+   **/
+  GKLocalPlayer *_Nullable optionalLocalPlayer = [[NSClassFromString(@"GKLocalPlayer") alloc] init];
   // If the GameKit Framework isn't linked - early out.
-  if ([GKLocalPlayer class] == nullptr) {
+  if (optionalLocalPlayer == nullptr) {
     return false;
   }
-  GKLocalPlayer *localPlayer = [GKLocalPlayer localPlayer];
+  __weak GKLocalPlayer *localPlayer = [[optionalLocalPlayer class] localPlayer];
```
chkuang-g, line 218: What you need to do here is basically
1. Get the class property [localPlayer](https://developer.apple.com/documentation/gamekit/gklocalplayer/1515401-localplayer?language=objc) from GKLocalPlayer, which is an instance of GKLocalPlayer
2. Get [isAuthenticated](https://developer.apple.com/documentation/gamekit/gklocalplayer/1515402-isauthenticated) from the instance of GKLocalPlayer

I think this can be replaced with [Objective-C Runtime](https://developer.apple.com/documentation/objectivec/objective-c_runtime). You may need to dig a bit deeper to figure out which functions to use.

I wonder why we have this public interface, while there is no such API for iOS SDK
https://firebase.google.com/docs/reference/ios/firebaseauth/api/reference/Classes/FIRGameCenterAuthProvider

I suggest you to contact iCore team if you want to get a quick answer for this.

auth/src/ios/credential_ios.mm
```diff
@@ -166,9 +166,16 @@ @implementation PhoneListenerDataObjC
   FIREBASE_ASSERT(future_api != nullptr);
 
   const auto handle = future_api->SafeAlloc<Credential>(kCredentialFn_GameCenterGetCredential);
+  /**
+   Linking GameKit.framework without using it on macOS results in App Store rejection.
+   Thus we don't link GameKit.framework to our SDK directly. `optionalLocalPlayer` is used for
+   checking whether the APP that consuming our SDK has linked GameKit.framework. If not, will
+   complete with kAuthErrorInvalidCredential error.
+   **/
+  GKLocalPlayer *_Nullable optionalLocalPlayer = [[NSClassFromString(@"GKLocalPlayer") alloc] init];
```
chkuang-g, line 175: By referencing `GKLocalPlayer`, the linker will try to link to Gamekit.framework.  We should avoid using this class entirely. (As well as `#import <GameKit/GameKit.h>` at the top of this file)

I think we can just use [NSClassFromString](https://developer.apple.com/documentation/foundation/1395135-nsclassfromstring). Ex.
```
if (NSClassFromString(@"GKLocalPlayer" == nil) {
  return false
}
```

Similar to the change below

auth/src/ios/credential_ios.mm
chkuang-g: 1. We do not need to remove this API now.
2. Seems like you are following the pattern in this https://github.com/firebase/firebase-ios-sdk/pull/2355. In that case, you change should be fine.

auth/src/ios/credential_ios.mm
chkuang-g: Similarly, since you are following https://github.com/firebase/firebase-ios-sdk/pull/2355/files, this should be fine

auth/src/ios/credential_ios.mm
```diff
@@ -166,9 +166,16 @@ @implementation PhoneListenerDataObjC
...
 
   // Early-out if GameKit is not linked
-  if ([GKLocalPlayer class] == nullptr) {
+  if (optionalLocalPlayer == nullptr) {
```
chkuang-g, line 178: nit: though it might not matter too much, but let's not mix with `nil` and `nullptr` here.

Use the following instead:
```
if (!optionalLocalPlayer) {
```

auth/src/ios/credential_ios.mm
```diff
@@ -197,11 +204,18 @@ @implementation PhoneListenerDataObjC
...
+  /**
+   Linking GameKit.framework without using it on macOS results in App Store rejection.
+   Thus we don't link GameKit.framework to our SDK directly. `optionalLocalPlayer` is used for
+   checking whether the APP that consuming our SDK has linked GameKit.framework. If not, 
+   early out.
+   **/
+  GKLocalPlayer *_Nullable optionalLocalPlayer = [[NSClassFromString(@"GKLocalPlayer") alloc] init];
   // If the GameKit Framework isn't linked - early out.
-  if ([GKLocalPlayer class] == nullptr) {
+  if (optionalLocalPlayer == nullptr) {
```
chkuang-g, line 215: Similar change to above

auth/src/ios/credential_ios.mm
cynthiajoan: Done

auth/src/ios/credential_ios.mm
cynthiajoan: Done

chkuang-g (approved):

bikram-cda:
How do I implement this fix in Unity SDK?

PR #359, opened 2021-04-07
Added a new test suite for SyncPoints

alexames:
The new test suite is driven by the same master set of test cases that
drive SyncPoint tests on the iOS and Android implementation.

All tests are run twice, once at the root, and once again at a deeper
level, to ensure that the depth of the changes made does not matter.

Some tests currently fail. It's not clear if these are legitimate bugs
in the test suite, or if there are some bugs in the DB implementation.
For the time being, these tests are skipped using `GTEST_SKIP();` and
are annotated with whether they result in an assert being triggered or
just a failure due to unmet expectations. Fixes for these failing tests
will be made in follow-up PR's.

### DIFF
--- a/database/src/desktop/persistence/persistence_manager.cc
+++ b/database/src/desktop/persistence/persistence_manager.cc
@@ -206,10 +206,8 @@ void PersistenceManager::DoPruneCheckAfterServerUpdate() {
           tracked_query_manager_->PruneOldQueries(*cache_policy_);
       PruneForestRef prune_forest_ref(&prune_forest);
       if (prune_forest_ref.PrunesAnything()) {
-        std::cout << __LINE__ << std::endl;
         storage_engine_->PruneCache(Path(), prune_forest_ref);
       } else {
-        std::cout << __LINE__ << std::endl;
         can_prune = false;
       }
       cache_size = storage_engine_->ServerCacheEstimatedSizeInBytes();
--- a/database/src/desktop/query_desktop.cc
+++ b/database/src/desktop/query_desktop.cc
@@ -55,7 +55,9 @@ static bool ValidateQueryEndpoints(const QueryParams& params, Logger* logger) {
       std::string start_name = GetStartName(params);
       if ((start_name != QueryParamsComparator::kMinKey) ||
           !(start_node.is_string())) {
-        logger->LogWarning(message);
+        if (logger) {
+          logger->LogWarning(message);
+        }
         return false;
       }
     }
@@ -64,7 +66,9 @@ static bool ValidateQueryEndpoints(const QueryParams& params, Logger* logger) {
       std::string end_name = GetEndName(params);
       if ((end_name != QueryParamsComparator::kMaxKey) ||
           !(end_node.is_string())) {
-        logger->LogWarning(message);
+        if (logger) {
+          logger->LogWarning(message);
+        }
         return false;
       }
     }
@@ -72,10 +76,12 @@ static bool ValidateQueryEndpoints(const QueryParams& params, Logger* logger) {
     if (params.order_by == QueryParams::kOrderByPriority) {
       if ((HasStart(params) && !IsValidPriority(GetStartValue(params))) ||
           (HasEnd(params) && !IsValidPriority(GetEndValue(params)))) {
-        logger->LogWarning(
-            "When using orderByPriority(), values provided to "
-            "StartAt(), EndAt(), or EqualTo() must be valid "
-            "priorities.");
+        if (logger) {
+          logger->LogWarning(
+              "When using orderByPriority(), values provided to "
+              "StartAt(), EndAt(), or EqualTo() must be valid "
+              "priorities.");
+        }
         return false;
       }
     }
@@ -86,12 +92,16 @@ static bool ValidateQueryEndpoints(const QueryParams& params, Logger* logger) {
 QueryInternal::QueryInternal(DatabaseInternal* database,
                              const QuerySpec& query_spec)
     : database_(database), query_spec_(query_spec) {
-  database_->future_manager().AllocFutureApi(&future_api_id_, kQueryFnCount);
+  if (database_) {
+    database_->future_manager().AllocFutureApi(&future_api_id_, kQueryFnCount);
+  }
 }
 
 QueryInternal::QueryInternal(const QueryInternal& internal)
     : database_(internal.database_), query_spec_(internal.query_spec_) {
-  database_->future_manager().AllocFutureApi(&future_api_id_, kQueryFnCount);
+  if (database_) {
+    database_->future_manager().AllocFutureApi(&future_api_id_, kQueryFnCount);
+  }
 }
 
 QueryInternal& QueryInternal::operator=(const QueryInternal& internal) {
@@ -104,15 +114,19 @@ QueryInternal& QueryInternal::operator=(const QueryInternal& internal) {
 QueryInternal::QueryInternal(QueryInternal&& internal)
     : database_(internal.database_),
       query_spec_(std::move(internal.query_spec_)) {
-  database_->future_manager().MoveFutureApi(&internal.future_api_id_,
-                                            &future_api_id_);
+  if (database_) {
+    database_->future_manager().MoveFutureApi(&internal.future_api_id_,
+                                              &future_api_id_);
+  }
 }
 
 QueryInternal& QueryInternal::operator=(QueryInternal&& internal) {
   database_ = internal.database_;
   query_spec_ = std::move(internal.query_spec_);
-  database_->future_manager().MoveFutureApi(&internal.future_api_id_,
-                                            &future_api_id_);
+  if (database_) {
+    database_->future_manager().MoveFutureApi(&internal.future_api_id_,
+                                              &future_api_id_);
+  }
   return *this;
 }
 #endif  // defined(FIREBASE_USE_MOVE_OPERATORS) || defined(DOXYGEN)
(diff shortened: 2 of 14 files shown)


alexames:
By the way, this is currently split up in to several individual commits. I thought it made it easier to understand what each change was to do it this way, and it would make for clearer blame history too. If you want me to squash it I can, but I think this way is better.

database/src/desktop/persistence/persistence_manager.cc
```diff
@@ -206,10 +206,8 @@ void PersistenceManager::DoPruneCheckAfterServerUpdate() {
           tracked_query_manager_->PruneOldQueries(*cache_policy_);
       PruneForestRef prune_forest_ref(&prune_forest);
       if (prune_forest_ref.PrunesAnything()) {
-        std::cout << __LINE__ << std::endl;
```
DellaBitta, line 209: what were these for?

database/tests/desktop/core/sync_point_spec_test.cc
```diff
@@ -0,0 +1,837 @@
...
+            registrations[callback_id] = event_registration;
+          }
+        }
+        std::vector<Event> actual = sync_tree_->AddEventRegistration(
+            UniquePtr<EventRegistration>(event_registration));
+        EXPECT_THAT(actual, Pointwise(EventEq(), expected));
+        break;
+      }
+      case test_data::StepType_unlisten: {
+        EventRegistration* event_registration = nullptr;
```
DellaBitta, line 272: This is bit inconsistent.  Above the event_registration isn't deliberately initialized.  Maybe add a nullptr assignment above as well?

database/tests/desktop/core/sync_point_spec_test.cc
```diff
@@ -0,0 +1,837 @@
...
+        int callback_id = spec->callbackId();
+        EXPECT_TRUE(callback_id != 0 && registrations.count(callback_id) != 0);
+        event_registration = registrations[callback_id];
```
DellaBitta, line 275: Should there be a assert here to ensure that this actually returned a valid pointer?

database/tests/desktop/core/sync_point_spec_test.cc
```diff
@@ -0,0 +1,837 @@
...
+        std::vector<Event> actual = sync_tree_->ApplyUserMerge(
+            path, merges, merges, current_write_id++, kPersist);
+        EXPECT_THAT(actual, UnorderedPointwise(EventEq(), expected));
+        break;
+      }
+      case test_data::StepType_ackUserWrite: {
+        int to_clear = static_cast<int>(spec->writeId());
+        AckStatus ack_status = spec->revert() ? kAckRevert : kAckConfirm;
+        std::vector<Event> actual =
+            sync_tree_->AckUserWrite(to_clear, ack_status, kPersist, 0);
```
DellaBitta, line 332: could you add a `/*{parameter_name}=*/0` for the last parameter here?

database/src/desktop/persistence/persistence_manager.cc
alexames: I dunno, I was probably debugging something at some point and left them in by mistake.

DellaBitta: Ok, cool.

database/src/desktop/util_desktop.cc
```diff
@@ -990,7 +990,7 @@ const std::string& GetHash(const Variant& data, std::string* output) {
 }
 
 bool IsValidPriority(const Variant& variant) {
-  return variant.is_numeric() || variant.is_string();
+  return variant.is_null() || variant.is_numeric() || variant.is_string();
```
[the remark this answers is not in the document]
alexames: it's not so much that it's a valid priority, it's that you're allowed to have no priority, and setting null is allowed when you have no priority.

database/src/desktop/query_desktop.cc
```diff
@@ -306,99 +320,118 @@ QueryInternal* QueryInternal::OrderByValue() {
...
 
 QueryInternal* QueryInternal::StartAt(const Variant& value,
                                       const char* child_key) {
-  Logger* logger = database_->logger();
-  if (!value.is_numeric() && !value.is_string() && !value.is_bool()) {
-    logger->LogWarning(
-        "Query::StartAt: Only strings, numbers, and boolean values are "
-        "allowed. (URL = %s)",
-        query_spec_.path.c_str());
+  Logger* logger = database_ ? database_->logger() : nullptr;
```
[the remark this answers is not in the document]
alexames: Good call.

database/src/desktop/query_desktop.cc
```diff
@@ -306,99 +320,118 @@ QueryInternal* QueryInternal::OrderByValue() {
...
 }
 
 QueryInternal* QueryInternal::EqualTo(const Variant& value) {
-  Logger* logger = database_->logger();
-  if (!value.is_numeric() && !value.is_string() && !value.is_bool()) {
-    logger->LogWarning(
-        "Query::EqualTo: Only strings, numbers, and boolean values are "
-        "allowed. (URL = %s)",
-        query_spec_.path.c_str());
+  Logger* logger = database_ ? database_->logger() : nullptr;
```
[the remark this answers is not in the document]
alexames: ack

database/src/desktop/query_desktop.cc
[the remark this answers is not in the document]
alexames: On second look, the logger is also being passed to the ValidateQueryEndpoints at the end of these functions, so leaving it at the top level makes sense I think.

database/src/desktop/util_desktop.cc
```diff
@@ -1229,6 +1231,25 @@ std::vector<std::string> split_string(const std::string& s, const char delimiter
...
+std::map<Path, Variant> VariantToPathMap(const Variant& data) {
+  std::map<Path, Variant> path_map;
+  if (data.is_map()) {
+    for (const auto& key_value : data.map()) {
+      Variant key_string_variant;
+      const char* key;
+      if (key_value.first.is_string()) {
+        key = key_value.first.string_value();
+      } else {
+        key_string_variant = key_value.first.AsString();
```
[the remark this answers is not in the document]
alexames: Agreed.

database/tests/desktop/core/sync_point_spec_test.cc
alexames: Done. I don't think it matters much as in both cases its guaranteed to be registered by the time it's used, but consistency is good.

database/tests/desktop/core/sync_point_spec_test.cc
alexames: Done.

database/tests/desktop/core/sync_point_spec_test.cc
alexames: Done

database/src/desktop/query_desktop.cc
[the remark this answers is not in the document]
DellaBitta: 👍

database/src/desktop/util_desktop.cc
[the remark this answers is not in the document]
DellaBitta: Understood.

DellaBitta (approved):

DellaBitta (changes requested):
New tests which fail should be disabled.  Could you create a story to revist them and put it on top of our Composing backlog.

alexames:
The tests that fail *are* being skipped, so something else is going on here. When I run the test executable directly it runs fine. I wonder if there's an issue with loading the data file?

alexames:
Figured out the issue. It was failing to load the data file. When I was running it locally the data that needed to be loaded was in the correct place relative to my current directory, but the tests run from a different directory than where I was so it failed to load. I've corrected the directory.

database/tests/desktop/core/sync_point_spec_test.cc
```diff
@@ -0,0 +1,883 @@
...
+  }
+
+  void SetUp() override {
+    sync_tree_ = new SyncTree(MakeUnique<WriteTree>(),
+                              MakeUnique<NoopPersistenceManager>(),
+                              MakeUnique<FakeListenProvider>(&logger_));
+  }
+
+  void TearDown() override { delete sync_tree_; }
+  void runOne(const char* name);
```
jonsimantov, line 119: nit: why aren't these named RunOne and RunTest?

database/tests/desktop/core/sync_point_spec_test.cc
```diff
@@ -0,0 +1,883 @@
...
+                     ? FlexbufferToVariant(event_spec->data_flexbuffer_root())
+                     : Variant::Null();
+  DataSnapshotInternal snapshot(nullptr, data, QuerySpec(path));
+
+  const char* prev_name =
+      event_spec->prevName() != nullptr ? event_spec->prevName()->c_str() : "";
+  return Event(type, nullptr, snapshot, prev_name);
+}
+
+void SyncTreeTest::runTest(const test_data::TestCase* test_spec,
```
jonsimantov, line 233: Is it possible to give any more context in the expect/assert output to help narrow down the issue if there's a failure?
For example, EXPECT_THAT(...) << "Listen step failed in " << test_spec->name() << ...any additional info

jonsimantov (changes requested):

database/tests/desktop/core/sync_point_spec_test.cc
alexames: No good reason, it was an oversight. Fixed.

database/tests/desktop/core/sync_point_spec_test.cc
alexames: The asserts in the library are pretty good about spitting out an error message saying why they're crashing specifically. While the tests have names, the individual steps do not. However, the tests that do interesting things have a "comment" field that describes what they are doing, so if you do get a failure you can see what the most recent Test and comment are. Additionally, the EXPECT_THAT(..., Pointwise(...)) construction will tell you whether there are elements missing, or elements that aren't the expected value, and at what index the element appears, etc.

I can add the step index if you think it would be helpful, but I'm not convinced it would speed debugging up any.

I do have a commit stashed in my local repo that contains a bunch of `operator<<` functions for various classes that print out the values of the class, I can check these in if you'd like. This can help with printing out error messages when, for example, an event isn't the expected value.

jonsimantov: I'd definitely err on the side of giving more info, but if you want to do that in a follow-up that's fine.

jonsimantov (dismissed):

DellaBitta (dismissed):

jonsimantov (approved):

jonsimantov (approved):
Stop rebasing :P

PR #361, opened 2021-04-09
Removing google play requirement.

vimanyu:
Firebase Android SDKs for Auth, Database, Storage, Functions do not require Google Play Services anymore.
https://firebase.google.com/docs/android/android-play-services

Removing checks for Google Play services for these products from the C++ SDK. 
Tested by running integration tests on, 
- devices with Google play services (https://github.com/firebase/firebase-cpp-sdk/actions/runs/730345555)
- devices without Google Play services (tested on Kindle)
- android emulator images with and without google play services. 

Addresses https://github.com/firebase/quickstart-unity/issues/975


vimanyu:
Not sure if we still need the App callback but I left it in there for now. Let me know if that can be safely removed too.

a-maurice (dismissed):

jonsimantov:
> Not sure if we still need the App callback but I left it in there for now. Let me know if that can be safely removed too.

What do you mean by "App callback"?

jonsimantov (changes requested):
Could you add an update to the readme file?

release_build_files/readme.md
```diff
@@ -584,6 +584,8 @@ code.
     -   General: Fixed rare crashes at application exit when destructors were
         being executed
         ([#345](https://github.com/firebase/firebase-cpp-sdk/pull/345)).
+    -   General: Removed checks for Google Play Services in C++ for Auth, Database,
```
jonsimantov, line 587: General (Android):

jonsimantov: Also, it should be spelled "Google Play services" (with a lower-case "s")

jonsimantov (approved):

jonsimantov:
> > Not sure if we still need the App callback but I left it in there for now. Let me know if that can be safely removed too.
> 
> What do you mean by "App callback"?

I figured out what you meant - those are needed for the Unity SDK to initialize the various C++ modules, I believe, and should be kept in place for now.

PR #365, opened 2021-04-12
Mark STLPort support deprecated

var-const:


cmake/external/firestore.cmake
```diff
@@ -18,9 +18,7 @@ if(TARGET firestore)
   return()
 endif()
 
-# Use a Firestore version slightly after 7.5.1, as it
-# contains a minor crash fix we'd like to include.
-set(version 5e76aa968bfce5e9c8fb0f28f523793f0e9ed5f3)
+set(version CocoaPods-7.7.0)
```
jonsimantov, line 21: If we're updating to Cocoapods 7.10.0 today, do I need to update this in my PR?

release_build_files/readme.md
```diff
@@ -571,6 +571,31 @@ code.
 
 ## Release Notes
 
+### 7.3.0
```
jonsimantov, line 574: Please put this into 7.2.0, we're putting together that version shortly today.

cmake/external/firestore.cmake
var-const: Sorry, I branched this off an older version of `main`. Closing in favor of #366.

release_build_files/readme.md
var-const: Done (in #366).

PR #366, opened 2021-04-12
Deprecate the STLPort support

var-const:


release_build_files/readme.md
```diff
@@ -571,13 +571,10 @@ code.
 
 ## Release Notes
 
-### 7.3.0
+### 7.2.0
 -   Changes
     -   General (Android): the STLPort support is deprecated and will be removed
         in the next major release. Please [use libc++ instead](https://developer.android.com/ndk/guides/cpp-support#cs).
```
jonsimantov, line 577: "Support for the STLPort STL variant is deprecated..."
Also wrap to 80 (put the link on the next line)?

jonsimantov: Or just copy the warning text exactly, I like that.

jonsimantov (changes requested):

release_build_files/readme.md
var-const: > Or just copy the warning text exactly, I like that.

Done.

Put the whole URL on the next line -- I'm not sure if splitting the markdown element in two could break the formatting?

jonsimantov: I was thinking wrapping it like this:
 -   General (Android): the STLPort support is deprecated and will be removed
        in the next major release. Please
        [use libc++ instead](https://developer.android.com/ndk/guides/cpp-support#cs).

jonsimantov (approved):

PR #368, opened 2021-04-12
Fix docs for setting and getting the dispatch queue in Firestore's settings class

dconeybe:
Fix a few documentation issues in the methods `dispatch_queue()` and `set_dispatch_queue(dispatch_queue_t)` of the `firebase::firestore::Settings` class:

1. The documentation is missing from https://firebase.google.com/docs/reference/cpp/class/firebase/firestore/settings because it is only compiled if `__OBJC__` is defined.
2. The documentation incorrectly states that the main thread is used for callbacks by default, which hasn't been true since 6.16.0 when commit 8d92e73b56308499c05ebc2d147fa3b0d79ad1e2 changed it to use a dedicated callback thread.
3. `set_dispatch_queue()` was not documented at all.


jonsimantov (approved):

firestore/src/include/firebase/firestore/settings.h
```diff
@@ -155,15 +155,41 @@ class Settings final {
...
   /**
-   * A dispatch queue to be used to execute all completion handlers and event
-   * handlers. By default, the main queue is used.
+   * Returns a dispatch queue that Firestore will use to execute callbacks.
+   *
+   * The returned dispatch queue is used for all completion handlers and event
+   * handlers.
+   *
+   * If no dispatch queue is explictly set by calling `set_dispatch_queue()`
+   * then a dedicated "callback thread" will be used; namely, the main thread
```
var-const, line 166: Nit: `s/thread/queue/`? "Thread" seems a little low-level here (and IIRC, libdispatch may run callbacks on different threads for the same queue).

var-const (approved):

var-const (approved):

PR #369, opened 2021-04-12
Feature/update brew version

vimanyu:
Fix for errors observed while installing packages with homebrew. Until Github runners are updated to use latest hombrew versions, manually updating homebrew in Github workflows.

https://github.com/Homebrew/brew/pull/11070
https://github.com/Homebrew/discussions/discussions/691

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -80,7 +80,6 @@ jobs:
 
       - name: Update homebrew (avoid bintray errors)
         if: startsWith(matrix.os, 'macos')
-        with:
         run: |
           brew update
 
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -81,6 +81,9 @@ jobs:
       - name: Update homebrew (avoid bintray errors)
         if: startsWith(matrix.os, 'macos')
         run: |
+          # Temporarily here until Github runners have updated their version of
+          # homebrew. This prevents errors arising from the shut down of
+          # binutils, used by older version of homebrew for hosting packages.
           brew update
 
       - name: Install prerequisites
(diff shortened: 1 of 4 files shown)


jonsimantov (commented):
Do the Integration tests and CPP packaging workflows need this as well?

.github/workflows/android.yml
```diff
@@ -78,6 +78,14 @@ jobs:
           path: /tmp/android-ndk-r16b
           key: android-ndk-${{ matrix.os }}-r16b
 
+      - name: Update homebrew (avoid bintray errors)
+        if: startsWith(matrix.os, 'macos')
```
jonsimantov, line 82: any particular reason this uses startsWith(matrix.os) and the below uses runner.os ==? Can you unify them?

jonsimantov (changes requested):

vimanyu:
Thanks for pointing it out. Yes, they needed this update too. 
I have scanned through the usages of install_prereqs scripts in our workflows and added this step prior.
Running integration_tests workflow just to confirm.

.github/workflows/android.yml
vimanyu: So I was trying to keep it consistent with any other macos checks in those blocks. 
For a couple of the blocks, I found the comparison used runner.os and for others, we used macos.

vimanyu: For example, cpp-packaging uses `runner.os` comparison,
https://github.com/firebase/firebase-cpp-sdk/commit/c87c8a4abf71d3c31f8f29fd9b4ca04d68b289f2

integration_tests.yml uses "matrix.os"
https://github.com/firebase/firebase-cpp-sdk/commit/ce9c9fa52ae3f96f43971b2040b545c439efbc00

jonsimantov (approved):

PR #370, opened 2021-04-12
Fix Firestore 7.10.0 build in C++ SDK, and match NanoPB versions.

jonsimantov:
Add the new Firestore nanopb proto header path to the build.
Update NanoPB package to match Firestore version.

### DIFF
--- a/firestore/CMakeLists.txt
+++ b/firestore/CMakeLists.txt
@@ -313,7 +313,7 @@ else()
   target_include_directories(firebase_firestore
     PUBLIC
       ${FIRESTORE_SOURCE_DIR}/Firestore/core/include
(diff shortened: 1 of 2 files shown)


firestore/CMakeLists.txt
```diff
@@ -312,6 +313,7 @@ else()
   target_include_directories(firebase_firestore
     PUBLIC
       ${FIRESTORE_SOURCE_DIR}/Firestore/core/include
+      ${FIRESTORE_SOURCE_DIR}/Firestore/Protos/nanpob
```
var-const, line 316: s/nanpob/nanopb/

jonsimantov: nice catch

var-const (approved):

PR #377, opened 2021-04-14
Added a code format python script to drive clang-format

DellaBitta:
Wrote the script in python for versatility.  Script will eventually be used by:

 - Github Actions to detect if formatting changes need to be made before merging to master,
 - Git hooks on all of our hosted platforms (Linux, Mac, Windows) where as shell scripts, bat files would have been a messy combination
 - Engineers who want to quickly format their sources.

The files the script checks are determine by command line arguments.  Options include:
   - Git Diff with master to determine which files have been altered.
   - A Directory
   - A Directory and recursively its subdirectories.

PR includes a formatted file for your reviewing pleasure.


scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+  "Append the directory, and recurse through its children, and add their "
+  "contents to the list of files to check.")
+flags.DEFINE_boolean("format_file", True, "Format files in place")
+flags.DEFINE_boolean("verbose", False, "Execute in verbose mode")
+
+# Constants:
+
+# The list of file types to run clang-format on.   Used to filter out
+# results when searching across directories or git diffs.
+FILE_TYPE_EXTENSIONS = (".ccp", ".cc", ".c", ".h")
```
jonsimantov, line 63: .cpp

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+CLANG_FORMAT_BASE = "clang-format -style=file "
+
+# Flag which informs clang-format to alter the file in place, overwriting
+# the original file with thew newly formatted version.
+CLANG_FORMAT_PARM_IN_PLACE = "-i "
+
+# Flag which informs clang-format to use stdout to log the changes it would
+# make to a file.  Logging in XML format allows us to easily parse the results 
+# to detect if the file requires formatting changes.
+CLANG_FORMAT_PARAM_LOG_REPLACEMENTS = "-output-replacements-xml "
```
jonsimantov, line 75: One nicer way than doing this (having concatenated strings with spaces included) is to just build all the parameters in a list in the first place. That way you also won't have to split them on " " which could cause an issue if filenames or paths have spaces in them.

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+  formatting changes the user needs to fix.
+    
+  Returns:
+    True if the file requires format changes, False if formatting would produce
+    an identical file.
+  """
+  command_line = CLANG_FORMAT_BASE + CLANG_FORMAT_PARAM_LOG_REPLACEMENTS
+  command_line += filename
+  args = command_line.split(" ")
+  proc = subprocess.Popen(command_line.split(" "), stdout=subprocess.PIPE)
```
jonsimantov, line 98: see above comment re: paths with spaces. If command_line is just a list the whole time, you won't have this issue.

jonsimantov (changes requested):

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+   1: Files which need formatting were found but weren't formatted
+     because -noformat_file was set.
+   2: The application wasn't configured properly and could not execute.
+"""
+import io
+import os
+import subprocess
+import sys
+
+from os import listdir
```
vimanyu, line 36: Since os is already imported you dont have to specifically import functions from that module.
You can just use `os.listdir` (also easier to read and follow that listdir is from os module) and `os.path.isfile` etc.

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+from os.path import isfile, join
+from absl import app
+from absl import flags
+
+
+# Flag Definitions:
+
+FLAGS = flags.FLAGS
```
vimanyu, line 44: nit: I feel absl is an overkill (another dependency) for just flag parsing as python provides an inbuilt module argparse which is pretty flexible. Feel free to ignore this comment though.

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+flags.DEFINE_boolean("git_diff", False, "Use git-diff to assemble a file list")
+flags.DEFINE_string("git_range", "main..", "the range string when using "
+  "git-diff")
+flags.DEFINE_multi_string("file", None, "Append the filename to the list of "
+  "files to check.")
+flags.DEFINE_multi_string("d", None, 
+  "Append the directory to the file list to format.")
+flags.DEFINE_multi_string("dr", None, 
+  "Append the directory, and recurse through its children, and add their "
+  "contents to the list of files to check.")
```
vimanyu, line 55: Generally, file and dir handling could be easier if modeled after common unix commands. Files/directories as positional arguments and a "-r" flag for recursive. 

`format_code.py <file1> <file2> <dir1> ... -r --git_diff`

Incase we want to stick to flags for these inputs, does absl have short hand notation for flags? If not, we should stay consistent as files is "file" but directory is "d". 

nit: for multi arguments, maybe "files" and "dirs" are better than singular names?

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+# The list of file types to run clang-format on.   Used to filter out
+# results when searching across directories or git diffs.
+FILE_TYPE_EXTENSIONS = (".ccp", ".cc", ".c", ".h")
+
+# Common execution arguments across all clang-format executions.
+CLANG_FORMAT_BASE = "clang-format -style=file "
+
+# Flag which informs clang-format to alter the file in place, overwriting
+# the original file with thew newly formatted version.
+CLANG_FORMAT_PARM_IN_PLACE = "-i "
```
vimanyu, line 70: nit: for consistency, PARAM?

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+
+# Flag which informs clang-format to use stdout to log the changes it would
+# make to a file.  Logging in XML format allows us to easily parse the results 
+# to detect if the file requires formatting changes.
+CLANG_FORMAT_PARAM_LOG_REPLACEMENTS = "-output-replacements-xml "
+
+# Conmmon exeuction arguments across all git diff executions.
```
vimanyu, line 77: nit: common execution

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+
+# The list of file types to run clang-format on.   Used to filter out
+# results when searching across directories or git diffs.
+FILE_TYPE_EXTENSIONS = (".ccp", ".cc", ".c", ".h")
+
+# Common execution arguments across all clang-format executions.
+CLANG_FORMAT_BASE = "clang-format -style=file "
+
+# Flag which informs clang-format to alter the file in place, overwriting
+# the original file with thew newly formatted version.
```
vimanyu, line 69: nit: the

scripts/format_code.py
vimanyu, line 98: For Python 3.5+, `subprocess.run` is more convenient and recommended way of running blocking commands. Checkout `capture_output` and `encoding` flags.
https://docs.python.org/3.7/library/subprocess.html

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+  formatted in place, with the results overwriting the previous version
+  of the file. Otherwise the result of the file reformat will be logged
+  to standard out.
+
+  Args:
+   filename: path to the file to format.
+  """
+  command_line = CLANG_FORMAT_BASE + CLANG_FORMAT_PARM_IN_PLACE + filename
+  args = command_line.split(" ")
+  proc = subprocess.Popen(args)
```
vimanyu, line 116: subprocess.run

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+      subdirectories found in path.
+  
+  Returns:
+    A list of matching filenames found in the directory.
+  """
+  filenames = []
+  for root, dirs, files in os.walk(path):
+    path = root.split(os.sep)
+    for filename in files:
+      if filename.endswith(FILE_TYPE_EXTENSIONS):
```
vimanyu, line 160: endswith will not work with a tuple/list.

vimanyu:
```
_, ext = os.path.splitext(filename)
if ext in FILE_TYPE_EXTENSION:
```

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+    path: the path to a directory to start searching from.
+    recurse: when True, will also recursively search for files in any
+      subdirectories found in path.
+  
+  Returns:
+    A list of matching filenames found in the directory.
+  """
+  filenames = []
+  for root, dirs, files in os.walk(path):
+    path = root.split(os.sep)
```
vimanyu, line 158: nit: doesn't seem to be used anywhere

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+  
+  Returns:
+    A list of filenames which match one of the extensions in
+    FILE_TYPE_EXTENSIONS.
+  """
+  filenames = []
+  if FLAGS.dr:
+    if FLAGS.verbose:
+      print()
+      print("Adding files by recursivly traversing directories:")
```
vimanyu, line 181: nit: recursively

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+    for directory in FLAGS.dr:
+      if FLAGS.verbose:
+        print("\tRecursive searching directory: \""+directory+"\"")
+      filenames += list_files_from_directory(directory, True)
+  if FLAGS.d:
+    for directory in FLAGS.d:
+      print("\Searching files in directory: \""+directory+"\"")
+      filenames += list_files_from_directory(directory, False)
+  if FLAGS.verbose:
+    print("  Found", len(filenames), "files.")
```
vimanyu, line 191: nit: easier to read with fomatting string,
```
print("  Found {0} files".format(len(filenames))

File: scripts/format_code.py [hunk 15]
```diff
@@ -0,0 +1,257 @@
...
+    if FLAGS.verbose:
+      print()
+      print("Adding files by recursivly traversing directories:")
+    for directory in FLAGS.dr:
+      if FLAGS.verbose:
+        print("\tRecursive searching directory: \""+directory+"\"")
+      filenames += list_files_from_directory(directory, True)
+  if FLAGS.d:
+    for directory in FLAGS.d:
+      print("\Searching files in directory: \""+directory+"\"")
```
vimanyu, line 188: Awesome feature of Python, you can use single, double, triple quotes for strings. And if any other quotes are within the string, they dont have to be escaped.

```
s = 'Recursively searching directory: "{0}"'.format(directory)
```

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+def validate_arguments():
+  """Ensures that a proper command line configuration exists to create a file
+  list either via git-diff operations or via manuallyd defined --file lists.
+  Logs an error if theres an errant configuration.
+  
+  Returns:
+    True if the either directories or git diff is defined, or both.  Returns
+    False otherwise signallingt hat execution should end.
+  """
+  if not FLAGS.git_diff:
```
vimanyu, line 204: Can be combined into one conditional.
```
if not FLAGS.git_diff and not FLAGS.file
```

scripts/format_code.py
```diff
@@ -0,0 +1,257 @@
...
+          format_file(filename)
+          count += 1
+        else:
+           print("  - File already formatted: \""+filename+"\"")
+    print("  - Formatted", count, "files.")
+  else:
+    count = 0
+    for filename in filenames:      
+      if does_file_need_formatting(filename):  
+        exit_code = 1
```
vimanyu, line 245: nit: non zero exit code usually indicates command failure but the script doesn't really fail here as we did not explicitly ask for formatting files. 
I am guessing this is for github server side to know that files weren't formatted and we shouldn't really continue. In that case, instead of using exit codes, it might be better to raise a `ValueError` with a nice descriptive message.

```
raise ValueError('something went wrong')
```

vimanyu (changes requested):
Thanks for attaching the sample file :). 
Made some general python comments.
Also, google style python docstrings specify the type if it is known beforehand.
https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

scripts/format_code.py
DellaBitta: I was following this pattern from the merge_libraries script in this same directory.  Noted for future reference but I don't think it's worth rewriting it at this point.   Thanks!

scripts/format_code.py
DellaBitta: Maybe I was hungry for a chicken parm!

scripts/format_code.py
DellaBitta: I wanted to use this script in a git hook, and [the pre commit git hook uses exit codes to signal errors](https://git-scm.com/docs/githooks#_pre_commit), so I'm going to stick witht his current implementation until I can test that out properly.   I assume that I can just check the exit code of the script in GHA as well.

scripts/format_code.py
DellaBitta: Actually it does work with tuples.   I tried it with a list and ran into the problem I think you are refering to, but making it a tuple smoothed it out.

scripts/format_code.py
vimanyu: Makes sense. Thanks for clarifying!

scripts/format_code.py
vimanyu: Whoa! TIL! Weirdly it doesn't work with lists but works with tuples.

scripts/format_code.py
DellaBitta: Done.

scripts/format_code.py
jonsimantov: Don't follow that, I barely knew python when I wrote it ;)

scripts/format_code.py
```diff
@@ -0,0 +1,227 @@
...
+  """
+  # The ACMRT filter lists files that have been (A)dded (C)opied (M)odified
+  # (R)enamed or (T)ype changed.
+  args = ['git', 'diff', FLAGS.git_range, '--name-only', '--diff-filter=ACMRT']
+  filenames = []
+  if FLAGS.verbose:
+    print()
+    print('Executing git diff to detect changed files, git_range: "{0}"'
+        .format(FLAGS.git_range))
+  result = subprocess.run(args, stdout=subprocess.PIPE)
```
vimanyu, line 108: Worth testing if the output is captured without capture_output. Also, I think subprocess.PIPE is not required. This might work,
`subprocess.run(args, capture_output=True)`

vimanyu (approved):
Minor comment but looks good otherwise!

scripts/format_code.py
DellaBitta: 'capture_output' was added in 3.7, which perhaps not all of our users will have installed, so I'm going with this option.

jonsimantov (approved):
It's a bit wordy in its output for a command-line utility, but we can fix that in post. ;)

PR #380, opened 2021-04-16
Source Code Format Checks

DellaBitta:
This change includes some tools to keep a consistent format of the repository's source files, including:

- a `Checks` workflow which errors if any of a branch's changed files require formatting.   We can extend this workflow with copyright header checks in the future and the pre-existing `Check PR Labels` has been folded into it.
- updates to `install_prereqs_desktop.py` to ensure that clang-format is installed.
- updates to `scripts/format_code.py` to reduce logging in non-verbose mode.
- a pre-commit githook. Note that the githook has to be installed manually by copying it from `scripts/git/hooks/pre-commit` to `.git/hooks/pre-commit`
- an update to our Contribution guidelines with formatting instructions.


scripts/format_code.py
```diff
@@ -70,7 +69,7 @@ def does_file_need_formatting(filename):
     an identical file.
   """
   args = ['clang-format', '-style=file', '-output-replacements-xml', filename]
-  result = subprocess.run(args, stdout=subprocess.PIPE)
+  result = subprocess.run(args, stdout=subprocess.PIPE, check=True)
```
dconeybe, line 72: It looks like `result` is unused. Can it be removed? Also on line 89.

DellaBitta: It's in the following line:
```
 for line in result.stdout.decode('utf-8').splitlines():
 ```

File: .github/workflows/consistency_checks.yml [hunk 2]
```diff
@@ -0,0 +1,30 @@
+name: Consistency Checks
```
jonsimantov, line 1: Wonder if we should merge Check PR Labels into this, to reduce the number of workflows running. WDYT?

scripts/format_code.py
```diff
@@ -175,14 +173,31 @@ def validate_arguments():
...
+  constructs that appear in obj-c but not in C/C++.
+  
+  Args:
+    filename (string): the name of the file to check
+  
+  Returns:
+    bool: True if the header file contains known obj-c language definitions.
+    Returns False otherwise.
+  """
+  if '.h' not in filename:
```
jonsimantov, line 190: nit: typically it's best to use one quote style throughout the script. ' vs "

jonsimantov:
Suggestion (possibly for a new PR) - add a label to autoformat a PR and push the changes in another commit, in case the developer doesn't want to run the formatter locally.

DellaBitta:
> Suggestion (possibly for a new PR) - add a label to autoformat a PR and push the changes in another commit, in case the developer doesn't want to run the formatter locally.

Just a heads up that the PR that we've been reviewing from a third party user submission has had issues with our labels mechanisms.  They stem from the fact that all of the workflows are being run from the user's github account, not ours, and they don't have write permissions on firebase repos. This causes some of the label steps - maybe only the ones which post comments..? -- from completing successfully.

I don't know if that applies at all to what you're suggesting, but I think that we need to retool the label control & flow that we already have.

scripts/format_code.py
```diff
@@ -175,14 +173,31 @@ def validate_arguments():
...
+    return False
+  objective_c_tags = [ '@end', '@class', '#import']
+  for line in open(filename):
```
vimanyu, line 193: Better to use context manager as that will take care of closing the file descriptor automatically.
```
with open(filename) as some_file:
   lines = some_file.readlines()
   for line in lines:
       ...
```

scripts/format_code.py
```diff
@@ -175,14 +173,31 @@ def validate_arguments():
...
+  Args:
+    filename (string): the name of the file to check
+  
+  Returns:
+    bool: True if the header file contains known obj-c language definitions.
+    Returns False otherwise.
+  """
+  if '.h' not in filename:
+    return False
+  objective_c_tags = [ '@end', '@class', '#import']
```
vimanyu, line 192: nit: optimization, use a tuple instead of list. Tuples in Python are immutable and have way faster searching (O(1)).

scripts/format_code.py
DellaBitta: Noted.

.github/workflows/consistency_checks.yml
DellaBitta: Sure. I can do that tomorrow morning.

scripts/format_code.py
vimanyu: nit: a safer and easier to read check would be to exactly compare the extension
```
_, ext = os.path.splitext(filename)
if ext != '.h':
   return False
```

vimanyu (commented):
minor python readability comments.

jonsimantov:
> > Suggestion (possibly for a new PR) - add a label to autoformat a PR and push the changes in another commit, in case the developer doesn't want to run the formatter locally.
> Just a heads up that the PR that we've been reviewing from a third party user submission has had issues with our itest labels.

Yeah, I'm aware of those issues. I still feel that this is worth doing for our internal development. Also, I have been thinking about switching over to using firebase-workflow-trigger's token to make the changes, where we were previously using the github-action default token to do. Do you know if the itest runs from forked repos have access to our repo's secrets?

scripts/format_code.py
```diff
@@ -195,32 +191,40 @@ def main(argv):
   if FLAGS.git_diff:
     filenames += git_diff_list_files()
 
-  print('Checking the format of {0} file(s). Please wait:'
-    .format(len(filenames)))
   exit_code = 0
+  if 0 == len(filenames):
```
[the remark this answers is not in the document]
DellaBitta: Done.

scripts/format_code.py
DellaBitta: Done.

scripts/format_code.py
DellaBitta: Done.

scripts/format_code.py
DellaBitta: Done

.github/workflows/consistency_checks.yml
DellaBitta: Done.

scripts/git/hooks/pre-commit
```diff
@@ -0,0 +1,2 @@
+echo .git/hooks/presubmit Checking File Formats
+python3 scripts/format_code.py -git_diff -noformat_file
```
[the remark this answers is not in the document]
DellaBitta: Done?  But not by me!

vimanyu (approved):
LGTM!

jonsimantov (approved):

PR #381, opened 2021-04-16
Reformat the codebase

DellaBitta:
Ran `scripts/format_code.py` on the entire sourcebase.


analytics/ios_headers/FIRUserPropertyNames.h
```diff
@@ -19,28 +19,31 @@
 ///
```
alexames, line 19: I don't think we should be formatting the obj c headers, should we?

app/src/fake/FIRApp.h
```diff
@@ -21,7 +21,6 @@ extern "C" {
 void FIRAppCreateUsingDefaultOptions(const char* name);
```
alexames, line 21: Another ObjC file I think that we should be leaving alone.

app/rest/transport_curl.cc
```diff
@@ -426,9 +445,7 @@ void BackgroundTransportCurl::CompleteOperation() {
 }
 
 void BackgroundTransportCurl::CheckOk(CURLcode code, const char* msg) {
-  if (code == CURLE_OK) {
-    return;
-  }
+  if (code == CURLE_OK) { return; }
```
alexames, line 448: nit: There's a bunch of one-liners like this that scooped up the {}'s. Can we set it up to not one-line {} blocks? If we must one-line them, can we remove the {}'s?

I ran the formatter within google3 on a few test files and it *wasn't* turning the bodies of if statements into one liners if they were in curly brackets. Is this just an inconsistency in formatter rules? Which one is more correct?

.clang-format
```diff
@@ -3,8 +3,12 @@ Standard: Cpp11
 ColumnLimit: 80
```
alexames, line 3: What's the rationale for the changes to this file? It seems to introduce more changes than it preserves. Shouldn't we stick with whatever the defaults of the Google style are?

performance/src_ios/fake/FIRHTTPMetric.h
```diff
@@ -5,18 +5,19 @@
...
-  FIRHTTPMethodPUT NS_SWIFT_NAME(put),
-  FIRHTTPMethodPOST NS_SWIFT_NAME(post),
-  FIRHTTPMethodDELETE NS_SWIFT_NAME(delete),
-  FIRHTTPMethodHEAD NS_SWIFT_NAME(head),
-  FIRHTTPMethodPATCH NS_SWIFT_NAME(patch),
-  FIRHTTPMethodOPTIONS NS_SWIFT_NAME(options),
-  FIRHTTPMethodTRACE NS_SWIFT_NAME(trace),
-  FIRHTTPMethodCONNECT NS_SWIFT_NAME(connect)
-} NS_SWIFT_NAME(HTTPMethod);
+      FIRHTTPMethodPUT NS_SWIFT_NAME(put),
```
alexames, line 8: hahaha what is going on here.

(probably a side effect of this being an ObjC file)

DellaBitta:
Closing this PR for a new formatting run.

app/rest/transport_curl.cc
jonsimantov: +1

.clang-format
jonsimantov: +1

PR #383, opened 2021-04-19
Code Format using clang-format's Google style defaults.

DellaBitta:
Updated codebase using clang-format's Google style basis without any further one-off customizations.

Clang-format seemingly. does some odd things to objective-c files.  Therefore directories known to contain objective-c files are ignored by defining a local .clang-format file which disables clang-format. Objective C .mm files are ignored in mixed-language directories, and Obj-C headers are detected by the format_code.py script which search for Obj-c language tags.


app/src/locale.cc
```diff
@@ -59,13 +59,11 @@ std::string GetLocale() {
...
   // default locale (empty in real life, or placeholder when running in a unit
   // test, as the test environment has no locale variables set).
-  std::string output = std::locale().name() != "C"
-                           ? std::locale().name()
-                           : getenv("LANG")
-                                 ? getenv("LANG")
-                                 : getenv("LC_CTYPE")
-                                       ? getenv("LC_CTYPE")
-                                       : getenv("TEST_TMPDIR") ? "en_US" : "";
+  std::string output = std::locale().name() != "C" ? std::locale().name()
```
alexames, line 62: nit: This is a case where I think the formatter is going to do worse than the nicely formatted table no matter what settings we use. If we go through with committing with these settings, can this block get wrapped in `// clang-format off/on`?

app/tests/optional_test.cc
```diff
@@ -168,10 +164,7 @@ TEST_F(OptionalTest, CopyConstructor) {
   EXPECT_TRUE(another_copy_of_optional_int.has_value());
   EXPECT_EQ(another_copy_of_optional_int.value(), 9999);
 
-  SetupExpectCall()
-    .Construct(1)
-    .CopyAndMove(2, 1)
-    .Destruct(4);
+  SetupExpectCall().Construct(1).CopyAndMove(2, 1).Destruct(4);
```
alexames, line 167: As mentioned above, I think this is a case where having the manual formatting is arguably superior (though I feel less strongly about this one). Another // clang-format off/on?

alexames (commented):
Overall I'm much happier with how this formatting pass went, though I think there are a small handful of places we ought to throw in some // clang-format off/on statements on custom-formatted blocks.

app/tests/optional_test.cc
DellaBitta: Done.

Also found a few more.

app/src/locale.cc
DellaBitta: Done.

DellaBitta:
Added so clang-format on / off toggles around code which flows better in the authored state.
Also removed the firestore/.clang-format file that I inherited from their repo as it seemed to cause more changes than our standard baseline .clang-format file.

app/src/log_stdio.cc
```diff
@@ -48,7 +48,7 @@ static const char* kLogLevelPrefix[] = {
 #if FIREBASE_PLATFORM_WINDOWS
 // Guards the log buffer on Windows.
 static Mutex g_log_mutex;  // NOLINT
-#endif  // FIREBASE_PLATFORM_WINDOWS
+#endif                     // FIREBASE_PLATFORM_WINDOWS
```
alexames, line 51: (This is the kind of thing I was talking about above)

alexames (approved):
This looks good to me. The one minor nitpick I have is that I'm not a huge fan of when the formatter decides to line up comments at the end of lines, because it can end up with some awkwardly formatted comments, but I don't think it's worth overriding google style to turn that off.

DellaBitta:
> This looks good to me. The one minor nitpick I have is that I'm not a huge fan of when the formatter decides to line up comments at the end of lines, because it can end up with some awkwardly formatted comments, but I don't think it's worth overriding google style to turn that off.

Yeah, I saw this too and it made me sigh.

PR #385, opened 2021-04-20
Fix NPE in android_abis.gradle if ANDROID_NDK_HOME is not set

dconeybe:
This fixes the following cryptic build error if the "ANDROID_NDK_HOME" environment variable is not set:

```
java.lang.NullPointerException: Cannot invoke method contains() on null object
```

### DIFF
--- a/android_build_files/android_abis.gradle
+++ b/android_build_files/android_abis.gradle
@@ -24,8 +24,9 @@ android {
       // Default list of ABIs available in up-to-date NDK.
       abiFilters "x86", "armeabi-v7a", "arm64-v8a", "x86_64"
 
-      if (System.getenv("ANDROID_NDK_HOME").contains("r16b") ||
-          System.getenv("ANDROID_NDK_HOME").contains("r11c")) {
+      if (System.getenv("ANDROID_NDK_HOME") != null && (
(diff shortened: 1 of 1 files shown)


jonsimantov (approved):

android_build_files/android_abis.gradle
```diff
@@ -24,8 +24,9 @@ android {
       // Default list of ABIs available in up-to-date NDK.
       abiFilters "x86", "armeabi-v7a", "arm64-v8a", "x86_64"
 
-      if (System.getenv("ANDROID_NDK_HOME").contains("r16b") ||
-          System.getenv("ANDROID_NDK_HOME").contains("r11c")) {
+      if (System.getenv("ANDROID_NDK_HOME") != null && (
+              System.getenv("ANDROID_NDK_HOME").contains("r16b") ||
```
jonsimantov, line 28: Could you replace these tabs with spaces?

jonsimantov (changes requested):

android_build_files/android_abis.gradle
dconeybe: It *was* spaces. Android Studio had formatted it like this. I tweaked it a bit. Let me know if you're okay with it now.

jonsimantov (approved):

PR #390, opened 2021-04-23
use native android call with app as parameter

vimanyu:
For Remote Config, use the method that allows passing in a custom app,
In the Android SDK, we call`getInstance(app)` instead of `getInstance()`.
Fixes: https://github.com/firebase/quickstart-unity/issues/991


cynthiajoan (approved):

release_build_files/readme.md
```diff
@@ -571,6 +571,12 @@ code.
 
 ## Release Notes
 
+### 7.3.1
```
cynthiajoan, line 574: Hmmm is this change going out with 7.3.0?

vimanyu: Done! Changed it to TBD

cynthiajoan (approved):

PR #392, opened 2021-04-23
Fixed a number of minor issues with database.

alexames:
After adding the new unit tests, a number of tests representing rare
edge cases were failing. This commit fixes all but one of the failing
tests.


database/src/desktop/view/view.cc
```diff
@@ -152,7 +152,7 @@ std::vector<Event> View::ApplyOperation(
     const Variant* opt_complete_server_cache,
     std::vector<Change>* out_changes) {
   if (operation.type == Operation::kTypeMerge &&
-      !operation.source.query_params.has_value()) {
+      operation.source.query_params.has_value()) {
```
DellaBitta, line 155: Could you explain this change a bit?  Seems like a pretty fundamental turn.

alexames: When I originally transcribed this code from the Java reference implementation I just got the condition exactly backwards. Whoops

The [reference implementation](https://github.com/firebase/firebase-android-sdk/blob/d92096257b3e750c72757fc7c4bc9d7393db2046/firebase-database/src/main/java/com/google/firebase/database/core/view/View.java#L163) had:

    if (operation.getType() == Operation.OperationType.Merge
        && operation.getSource().getQueryParams() != null) {

And checking for `!= null` is equivalent to checking if an optional's `.has_value()` is true.

DellaBitta: Gotcha, thanks.  👍

DellaBitta (approved):

PR #394, opened 2021-04-25
Implement Firestore Bundles

wu-hui:


firestore/src/android/firestore_android.cc
```diff
@@ -515,6 +528,49 @@ void FirestoreInternal::ClearListeners() {
...
+Future<LoadBundleTaskProgress> FirestoreInternal::LoadBundle(
+    const std::string& bundle,
+    std::function<void(const LoadBundleTaskProgress&)> progress_callback) {
+  Env env = GetEnv();
+  Local<String> bundle_jstring = env.NewStringUtf(bundle.c_str());
+  Local<String> encoding = env.NewStringUtf("UTF-8");
+  Local<Array<uint8_t>> bytes = bundle_jstring.GetBytes(env, encoding);
+  Local<LoadBundleTaskInternal> task = env.Call(obj_, kLoadBundle, bytes.get());
+
+  auto* listener = new LambdaEventListener<LoadBundleTaskProgress>(
```
dconeybe, line 552: Where does `listener` get deleted? Does the `LambdaEventListener` look after this? I poked around the code quickly and couldn't find where it gets deleted.

firestore/src/android/load_bundle_task_android.h
```diff
@@ -0,0 +1,42 @@
...
+
+/** A C++ proxy for a Java `LoadBundleTask`, which is a subclass of `Task`. */
+class LoadBundleTaskInternal : public jni::Task {
+ public:
+  using Task::Task;
+
+  static void Initialize(jni::Loader& loader);
+
+  void AddProgressListener(jni::Env& env,
+                           const jni::Global<jni::Object>& executor,
```
dconeybe, line 35: In general, method arguments should not be `Global` or `Local` template types. Just accept `Object&` (or `const Object&`). That way the caller can specify either a local *or* a global reference.

firestore/src/android/load_bundle_task_progress_android.cc
```diff
@@ -0,0 +1,96 @@
...
+int64_t LoadBundleTaskProgressInternal::total_bytes() const {
+  Env env = GetEnv();
+  return env.Call(obj_, kGetTotalBytes);
+}
+
+LoadBundleTaskProgress::State LoadBundleTaskProgressInternal::state() const {
+  Env env = GetEnv();
+  Local<Object> state = env.Call(obj_, kGetTaskState);
+  // Get the name of the enum value for inspection.
+  std::string enum_name = env.Call(state, kName).ToString(env);
```
dconeybe, line 85: You should probably check the instances of the enum instead of hardcoding the string values.

Here is an example where the `java.lang.Thread.State` enum is used.

https://github.com/firebase/firebase-cpp-sdk/blob/8f3e19226639a3b91e5d879cdf5f3d867dbf8db4/firestore/integration_test_internal/src/android/jni_runnable_android_test.cc#L34

firestore/src/include/firebase/firestore/load_bundle_task_progress.h
```diff
@@ -0,0 +1,92 @@
...
+  explicit LoadBundleTaskProgress(LoadBundleTaskProgressInternal* internal);
+
+  int32_t documents_loaded_;
+  int32_t total_documents_;
+  int64_t bytes_loaded_;
+  int64_t total_bytes_;
+  State state_;
+};
+
+inline bool operator==(const LoadBundleTaskProgress lhs,
```
dconeybe, line 75: Is `lhs` supposed to be a reference? Also in `operator!=`.

firestore/src/common/load_bundle_task_progress.cc
```diff
@@ -0,0 +1,53 @@
...
+LoadBundleTaskProgress::LoadBundleTaskProgress(
+    LoadBundleTaskProgressInternal* internal) {
+  FIREBASE_ASSERT(internal != nullptr);
+  documents_loaded_ = internal->documents_loaded();
+  total_documents_ = internal->total_documents();
+  bytes_loaded_ = internal->bytes_loaded();
+  total_bytes_ = internal->total_bytes();
+  state_ = internal->state();
+
+  delete internal;
```
dconeybe, line 49: Is this conventional where the constructor deletes the `internal` object? This looks very risky. Is there a better way to achieve this?

firestore/src/ios/firestore_ios.cc
```diff
@@ -299,9 +322,74 @@ void Firestore::set_log_level(LogLevel log_level) {
...
+void FirestoreInternal::SetClientLanguage(const std::string &language_token) {
   api::Firestore::SetClientLanguage(language_token);
 }
 
+Future<LoadBundleTaskProgress> FirestoreInternal::LoadBundle(
+    const std::string &bundle) {
+  auto promise = promise_factory_.CreatePromise<LoadBundleTaskProgress>(
+      AsyncApi::kLoadBundle);
+  auto bundle_stream = absl::make_unique<util::ByteStreamCpp>(
+      absl::make_unique<std::stringstream>(std::stringstream(bundle)));
```
dconeybe, line 334: Is it necessary to create the temporary `stringstream` object for the call to `make_unique()`? Could you just specify `bundle` directly? Also on line 358.

firestore/src/ios/firestore_ios.cc
```diff
@@ -299,9 +322,74 @@ void Firestore::set_log_level(LogLevel log_level) {
...
+  std::shared_ptr<api::LoadBundleTask> task =
+      firestore_core_->LoadBundle(std::move(bundle_stream));
+  task->Observe(
+      [promise, task](const api::LoadBundleTaskProgress &progress) mutable {
+        if (progress.state() == api::LoadBundleTaskState::kSuccess) {
+          promise.SetValue(ToApiProgress(progress));
+          task->RemoveAllObservers();
+        }
```
dconeybe, line 342: nit: Use an "else if" here to be clearer about control flow. Also on line 367.

firestore/src/include/firebase/firestore/load_bundle_task_progress.h
```diff
@@ -0,0 +1,92 @@
...
+
+  /** Returns the total number of bytes in the bundle. Returns 0 if the bundle
+   * failed to parse. */
+  int64_t total_bytes() const { return total_bytes_; }
+
+  /** Returns the current state of the loading progress. */
+  State state() const { return state_; }
+
+ private:
+  friend class EventListenerInternal;
```
dconeybe, line 62: Why does this class need `friend` declarations? Can't all of its private members be directly accessed via the public getter methods?

firestore/src/ios/load_bundle_task_progress_ios.h
```diff
@@ -0,0 +1,47 @@
...
+#ifndef FIREBASE_FIRESTORE_CLIENT_CPP_SRC_IOS_LOAD_BUNDLE_TASK_PROGRESS_IOS_H_
+#define FIREBASE_FIRESTORE_CLIENT_CPP_SRC_IOS_LOAD_BUNDLE_TASK_PROGRESS_IOS_H_
+
+#include <cstdint>
+
+namespace firebase {
+namespace firestore {
+
+class LoadBundleTaskProgressInternal {
+ public:
```
dconeybe, line 25: How does this class get constructed? Wouldn't it need a constructor to initialize the instance variables? I think I'm missing something.

firestore/src/ios/load_bundle_task_progress_ios.h
```diff
@@ -0,0 +1,47 @@
...
+  int32_t total_documents() const { return total_documents_; }
+
+  int64_t bytes_loaded() const { return bytes_loaded_; }
+
+  int64_t total_bytes() const { return total_bytes_; }
+
+  LoadBundleTaskProgress::State state() const { return state_; }
+
+ private:
+  int32_t documents_loaded_;
```
dconeybe, line 37: Please initialize these instance variables. It's good practice to initialize instance variables of builtin type, to avoid any possibility of accessing uninitialized values.

e.g. `int32_t documents_loaded_ = 0;`

firestore/src/include/firebase/firestore.h
```diff
@@ -402,9 +403,41 @@ class Firestore {
...
+   * Loads a Firestore bundle into the local cache.
+   *
+   * @param bundle A string representing the bundle to be loaded.
+   * @return A `Future` that is resolved when the loading is either completed,
+   * or aborted due to an error.
+   */
+  virtual Future<LoadBundleTaskProgress> LoadBundle(const std::string &bundle);
+
+  /**
+   * Loads a Firestore bundle into the local cache, with provided callbacks
```
dconeybe, line 419: Please document which thread the callbacks are made on.

firestore/integration_test_internal/src/bundle_test.cc
```diff
@@ -0,0 +1,196 @@
...
+      auto limit_to_last_snapshot = *Await(limit_to_last.Get(Source::kCache));
+      EXPECT_THAT(
+          QuerySnapshotToValues(limit_to_last_snapshot),
+          testing::ElementsAre(MapFieldValue{{"k", FieldValue::String("a")},
+                                             {"bar", FieldValue::Integer(1)}}));
+    }
+  }
+};
+
+TEST_F(BundleTest, CanLoadBundlesWithProgressUpdates) {
```
dconeybe, line 76: Please add a test that explicitly tests loading bundles *without* progress updates.

firestore/integration_test_internal/src/bundle_test.cc
```diff
@@ -0,0 +1,196 @@
+#include <string>
```
dconeybe, line 1: Please add tests for `NamedQuery()`, especially specifying an empty and a non-existent name.

firestore/src/include/firebase/firestore.h
```diff
@@ -402,9 +403,41 @@ class Firestore {
...
+   * updates.
+   * @return A `Future` that is resolved when the loading is either completed,
+   * or aborted due to an error.
+   */
+  virtual Future<LoadBundleTaskProgress> LoadBundle(
+      const std::string &bundle,
+      std::function<void(const LoadBundleTaskProgress &)> progress_callback);
+
+  /**
+   * Reads a Firestore `Query` from local cache, identified by the given name.
```
dconeybe, line 433: Document the behavior if the given `query_name` is not known.

dconeybe (commented):
I've done a first pass review. It's still a bit of a blur to me so I'll take a better look in the next day or two!

firestore/src/include/firebase/firestore/load_bundle_task_progress.h
wu-hui: They are the places that will call the private constructor. Many classes in this directory follow this.

firestore/src/include/firebase/firestore/load_bundle_task_progress.h
wu-hui: Thanks!

firestore/src/ios/firestore_ios.cc
wu-hui: Done.

firestore/src/ios/firestore_ios.cc
wu-hui: Done.

firestore/src/ios/load_bundle_task_progress_ios.h
wu-hui: This class is not used, and only here to satisfy the type system. I deleted it and went for conditional compilation instead.

firestore/src/ios/load_bundle_task_progress_ios.h
wu-hui: No longer applies.

firestore/firestore_resources/build.gradle
```diff
@@ -49,8 +49,8 @@ android {
 }
 
 dependencies {
-  implementation 'com.google.firebase:firebase-analytics:18.0.2'
-  implementation 'com.google.firebase:firebase-firestore:22.0.2'
+  implementation 'com.google.firebase:firebase-analytics:18.0.3'
```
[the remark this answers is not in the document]
wu-hui: I need the version bump to include bundles. Will request for review from FPL once I get internal LGTMs.

wu-hui (commented):
Thanks for the review!

var-const (changes requested):
@wu-hui We should make the references and pointers stick to the type, not the value. Otherwise, this produces a huge spurious delta (and will continue to do so as we work on other files).

Our configuration file for the iOS SDK is [here](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/.clang-format). The most relevant options are `DerivePointerAlignment` and `PointerAlignment`. We set that to `false` and `left`, correspondingly (the Google style is to set `DerivePointerAlignment` to `true). See the [full docs](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)

What we probably need is to add a `.clang-format` file to the `firestore/` directory. I'm surprised that `PointerAlignment: right` seems to be used because the `.clang-format` at the root of the repo is based on the Google style, which should derive the alignment.

wu-hui:
> @wu-hui We should make the references and pointers stick to the type, not the value. Otherwise, this produces a huge spurious delta (and will continue to do so as we work on other files).
> 
> Our configuration file for the iOS SDK is [here](https://github.com/firebase/firebase-ios-sdk/blob/master/Firestore/core/.clang-format). The most relevant options are `DerivePointerAlignment` and `PointerAlignment`. We set that to `false` and `left`, correspondingly (the Google style is to set `DerivePointerAlignment` to `true). See the [full docs](https://clang.llvm.org/docs/ClangFormatStyleOptions.html)
> 
> What we probably need is to add a `.clang-format` file to the `firestore/` directory. I'm surprised that `PointerAlignment: right` seems to be used because the `.clang-format` at the root of the repo is based on the Google style, which should derive the alignment.

Done.

dconeybe (commented):
Andy, I had a few minutes to re-review this today. I still need to do a thorough review, which I hope to do on Monday. But I thought I'd send some feedback based on another quick review. Apologies for the delay.

firestore/src/android/firestore_android.cc
```diff
@@ -515,6 +528,49 @@ void FirestoreInternal::ClearListeners() {
   listener_registrations_.clear();
 }
 
+Future<LoadBundleTaskProgress> FirestoreInternal::LoadBundle(
+    const std::string& bundle) {
+  Env env = GetEnv();
+  Local<String> bundle_jstring = env.NewStringUtf(bundle.c_str());
+  Local<String> encoding = env.NewStringUtf("UTF-8");
```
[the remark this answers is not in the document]
wu-hui: Done for real..

firestore/src/ios/firestore_ios.cc
wu-hui: Done for real..

firestore/integration_test_internal/src/bundle_test.cc
```diff
@@ -0,0 +1,278 @@
...
+          std::lock_guard<std::mutex> lk(lock);
+          DeleteFirestore(db);
+          cv.notify_one();
+        }
+      });
+
+  // Wait for the notification that the instance is deleted. We cannot rely on
+  // the returned future because it will not complete.
+  std::unique_lock<std::mutex> lk(lock);
+  cv.wait(lk);
```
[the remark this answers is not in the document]
var-const: Note that you can pass a predicate to `wait` to deal with spurious wakeups -- see https://en.cppreference.com/w/cpp/thread/condition_variable/wait. Using a promise is likely easier, too.

firestore/src/android/firestore_android.cc
wu-hui: As discussed offline, I decided to take up some extent of memory leak to support the unlikely case of deleting firestore in the middle of progress update.

dconeybe (dismissed):

firestore/integration_test_internal/src/bundle_test.cc
```diff
@@ -0,0 +1,278 @@
...
+  Future<LoadBundleTaskProgress> result = db->LoadBundle(
+      bundle, [&progresses](const LoadBundleTaskProgress& progress) {
+        progresses.push_back(progress);
+      });
+
+  auto final_progress = AwaitResult(result);
+
+  // 4 progresses will be reported: initial, document 1, document 2, final
+  // success.
+  EXPECT_EQ(progresses.size(), 4);
```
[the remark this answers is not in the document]
wu-hui: Done.

firestore/src/android/firestore_android.cc
```diff
@@ -562,14 +563,18 @@ Future<LoadBundleTaskProgress> FirestoreInternal::LoadBundle(
         // listeners.
         progress_callback(p);
       });
+  // TODO(b/187420421): The listener is owned by the firestore instance, longer
+  // than ideal. This is to support the unlikely case when user try to delete
+  // Firestore instance in the listener. Once the referred bug is fixed, it can
+  // be managed through a `shared_ptr`.
+  bundle_listeners_.push_back(
```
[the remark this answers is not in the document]
wu-hui: Done.

wu-hui (commented):
Thanks!

firestore/src/android/firestore_android.h
```diff
@@ -214,6 +214,8 @@ class FirestoreInternal {
 #else   //  defined(_STLPORT_VERSION)
   std::unordered_set<ListenerRegistrationInternal*> listener_registrations_;
 #endif  //  defined(_STLPORT_VERSION)
+
+  Mutex bundle_listener_mutex_;
```
dconeybe, line 218: ultranit: The mutex should be named `bundle_listeners_mutex_`, instead of `bundle_listener_mutex_`.

dconeybe (approved):

var-const (changes requested):

var-const (changes requested):
For the public documentation, can you please run it by a tech writer?

Also, sorry about the delay!

firestore/src/include/firebase/firestore.h
```diff
@@ -405,6 +406,41 @@ class Firestore {
...
+   * updates.
+   * @return A `Future` that is resolved when the loading is either completed,
+   * or aborted due to an error.
+   */
+  virtual Future<LoadBundleTaskProgress> LoadBundle(
+      const std::string& bundle,
+      std::function<void(const LoadBundleTaskProgress&)> progress_callback);
+
+  /**
+   * Reads a Firestore `Query` from local cache, identified by the given name.
```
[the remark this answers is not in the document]
wu-hui: I choose to model after Android because the API resemble more than iOS. I did a pass from `FIRFirestore.h` to bring in some more change.

cmake/external/firestore.cmake
```diff
@@ -18,12 +18,12 @@ if(TARGET firestore)
   return()
 endif()
 
-set(version 9bae3613d885fb8cc7d74c612c90f821a0e1981f)
 ExternalProject_Add(
   firestore
 
   DOWNLOAD_DIR ${FIREBASE_DOWNLOAD_DIR}
-  URL https://github.com/firebase/firebase-ios-sdk/archive/${version}.tar.gz
+  GIT_REPOSITORY https://github.com/firebase/firebase-ios-sdk
```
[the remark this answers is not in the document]
dconeybe: If we make this change, we should probably also add `GIT_TAG` to pin it to a specific commit. As it is I'm guessing it will just get HEAD from the master branch.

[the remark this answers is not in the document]
dconeybe: Nevermind my comment. I didn't see the entire diff and you *did* specify `GIT_TAG`.

firestore/src/android/firestore_android.h
```diff
@@ -192,6 +215,10 @@ class FirestoreInternal {
   std::unordered_set<ListenerRegistrationInternal*> listener_registrations_;
 #endif  //  defined(_STLPORT_VERSION)
 
+  Mutex bundle_listener_mutex_;
+  std::vector<std::unique_ptr<LambdaEventListener<LoadBundleTaskProgress>>>
```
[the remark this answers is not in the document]
dconeybe: It's probably a good idea to add an inline comment here to explain that the entire purpose of this vector is to have the `LambdaEventListener` objects destroyed when the `FirestoreInternal` is destroyed. Otherwise, it just looks odd to have this vector of unique pointers that are never read.

firestore/src/android/firestore_android.cc
```diff
@@ -515,6 +542,58 @@ void FirestoreInternal::ClearListeners() {
...
+  Local<LoadBundleTaskInternal> task = CreateLoadBundleTask(env, obj_, bundle);
+
+  auto* listener = new LambdaEventListener<LoadBundleTaskProgress>(
+      [progress_callback](const LoadBundleTaskProgress& p, Error e,
+                          const std::string& s) {
+        // `e` is always `kErrorOk` and `s` is always empty for progress
+        // listeners.
+        progress_callback(p);
+      });
+  // TODO(b/187420421): The listener is owned by the firestore instance, longer
```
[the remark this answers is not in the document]
dconeybe: Here is the bug tracking the feature to which @wu-hui alluded: b/187420421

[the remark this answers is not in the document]
var-const: Thanks for the explanation!

> Unique pointer is here to help these listener to be owned by the firestore instance, which make sure they only get deleted when the firestore instance is going away.

What I'm trying to figure out is, what's the advantage of using unique pointers instead of just plain values?

var-const (changes requested):

firestore/src/android/event_listener_android.cc
```diff
@@ -132,15 +153,36 @@ void EventListenerInternal::VoidEventListenerNativeOnEvent(JNIEnv*, jclass,
...
+    jobject progress) {
+  if (listener_ptr == 0) {
+    return;
+  }
+  auto* firestore = reinterpret_cast<FirestoreInternal*>(firestore_ptr);
+  auto* listener =
+      reinterpret_cast<EventListener<LoadBundleTaskProgress>*>(listener_ptr);
+  LoadBundleTaskProgress cpp_progress(
+      // Internal progress is owned by `cpp_progress`.
+      new LoadBundleTaskProgressInternal(firestore, Object(progress)));
```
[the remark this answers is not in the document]
wu-hui: I see, makes sense.

firestore/src/android/firestore_android.cc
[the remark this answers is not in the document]
wu-hui: Changed to value.

firestore/src/android/firestore_android.h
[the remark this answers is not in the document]
wu-hui: Done.

firestore/src/common/load_bundle_task_progress.cc
```diff
@@ -0,0 +1,55 @@
...
+                                               int64_t total_bytes,
+                                               State state)
+    : documents_loaded_(documents_loaded),
+      total_documents_(total_documents),
+      bytes_loaded_(bytes_loaded),
+      total_bytes_(total_bytes),
+      state_(state) {
+}
+
+#if defined(__ANDROID__)
```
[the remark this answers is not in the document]
wu-hui: Done.

firestore/src/common/load_bundle_task_progress.cc
```diff
@@ -37,6 +37,8 @@ LoadBundleTaskProgress::LoadBundleTaskProgress(int32_t documents_loaded,
       state_(state) {
 }
 
+// Android requires below constructor to create this object from internal
+// objects in a promise. See promise_android.h
 #if defined(__ANDROID__)
 LoadBundleTaskProgress::LoadBundleTaskProgress(
     LoadBundleTaskProgressInternal* internal) {
```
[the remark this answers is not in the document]
wu-hui: Done.

firestore/src/include/firebase/firestore.h
```diff
@@ -405,6 +406,42 @@ class Firestore {
...
+  virtual Future<LoadBundleTaskProgress> LoadBundle(
+      const std::string& bundle,
+      std::function<void(const LoadBundleTaskProgress&)> progress_callback);
+
+  /**
+   * Reads a Firestore `Query` from the local cache, identified by the given
+   * name.
+   *
+   * Named queries are packaged into bundles on the server side (along with
+   * resulting documents) and loaded to local cache using `LoadBundle`. Once in
```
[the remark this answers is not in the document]
wu-hui: Done.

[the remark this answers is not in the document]
wu-hui: Done.

firestore/src/include/firebase/firestore.h
```diff
@@ -405,6 +406,42 @@ class Firestore {
...
+   * the local cache, you can use this method to extract a query by name.
+   *
+   * If a query cannot be found, the returned future will complete with its
+   * `error()` set to a non-zero error code.
+   */
+  virtual Future<Query> NamedQuery(const std::string& query_name);
```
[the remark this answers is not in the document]
wu-hui: Done.

firestore/src/include/firebase/firestore/load_bundle_task_progress.h
```diff
@@ -0,0 +1,104 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#ifndef FIREBASE_FIRESTORE_CLIENT_CPP_SRC_INCLUDE_FIREBASE_FIRESTORE_LOAD_BUNDLE_TASK_PROGRESS_H_
+#define FIREBASE_FIRESTORE_CLIENT_CPP_SRC_INCLUDE_FIREBASE_FIRESTORE_LOAD_BUNDLE_TASK_PROGRESS_H_
+
+#include <cstdint>
```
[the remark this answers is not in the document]
wu-hui: Wired. Done now.

firestore/src/include/firebase/firestore/load_bundle_task_progress.h
```diff
@@ -77,11 +79,11 @@ class LoadBundleTaskProgress {
...
-  int32_t documents_loaded_;
-  int32_t total_documents_;
-  int64_t bytes_loaded_;
-  int64_t total_bytes_;
-  State state_;
+  int32_t documents_loaded_ = 0;
+  int32_t total_documents_ = 0;
+  int64_t bytes_loaded_ = 0;
+  int64_t total_bytes_ = 0;
+  State state_ = LoadBundleTaskProgress::State::kInProgress;
```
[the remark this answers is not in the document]
wu-hui: Done.

firestore/src_java/com/google/firebase/firestore/internal/cpp/FirestoreTasks.java
```diff
@@ -17,7 +17,7 @@ private FirestoreTasks() {
   }
 
   /**
-   * Takes a Task and returns a new Task, whose result will be the same as the incoming Task,
+   * Takes a {@link Task}and returns a new Task, whose result will be the same as the incoming Task,
```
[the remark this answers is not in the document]
wu-hui: Done.

firestore/src/ios/firestore_ios.cc
```diff
@@ -303,5 +327,68 @@ void FirestoreInternal::SetClientLanguage(const std::string& language_token) {
...
+    std::function<void(const LoadBundleTaskProgress&)> progress_callback) {
+  auto promise = promise_factory_.CreatePromise<LoadBundleTaskProgress>(
+      AsyncApi::kLoadBundle);
+  auto bundle_stream = absl::make_unique<util::ByteStreamCpp>(
+      absl::make_unique<std::stringstream>(bundle));
+  std::shared_ptr<api::LoadBundleTask> task =
+      firestore_core_->LoadBundle(std::move(bundle_stream));
+  task->Observe([promise, task, progress_callback](
+                    const api::LoadBundleTaskProgress& progress) mutable {
+    progress_callback(ToApiProgress(progress));
```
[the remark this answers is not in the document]
wu-hui: For iOS, the progress observer and success observer are always called from a single thread, and it is always progress first, success later. See: https://osscs.corp.google.com/firebase-sdk/firebase-ios-sdk/+/master:Firestore/core/src/api/load_bundle_task.cc;l=105?q=load_bundle_task.cc

For Android, a single thread executor is always supplied by the C++ layer to underlying Android SDK, so the behavior is the same. See https://osscs.corp.google.com/firebase-sdk/firebase-android-sdk/+/master:firebase-firestore/src/main/java/com/google/firebase/firestore/LoadBundleTask.java;l=510?q=loadbundletask.java

firestore/src/android/firestore_android.h
```diff
@@ -216,8 +216,8 @@ class FirestoreInternal {
 #endif  //  defined(_STLPORT_VERSION)
 
   Mutex bundle_listeners_mutex_;
-  std::vector<std::unique_ptr<LambdaEventListener<LoadBundleTaskProgress>>>
-      bundle_listeners_;
+  // Using a vector to ensure FirestoreInternal owns all listener instance.
```
var-const, line 219: Optional: consider `to ensure listener instances cannot outlive FirestoreInternal`.

var-const (dismissed):

firestore/src/android/firestore_android.h
wu-hui: Done.

firestore/src/android/firestore_android.cc
```diff
@@ -515,6 +542,59 @@ void FirestoreInternal::ClearListeners() {
...
+
+  EventListener<LoadBundleTaskProgress>* listener_ptr = nullptr;
+  // TODO(b/187420421): The listener is owned by the firestore instance, longer
+  // than ideal. This is to support the unlikely case when user try to delete
+  // Firestore instance in the listener. Once the referred bug is fixed, it can
+  // be managed through a `shared_ptr`.
+  {
+    MutexLock lock(bundle_listeners_mutex_);
+    bundle_listeners_.push_back(std::move(listener));
+    listener_ptr = &(bundle_listeners_.back());
```
dconeybe, line 576: Pointers to vector elements are not stable because they can be relocated with a future call to push_back(). This means that `listener_ptr` could become invalid in the future, which would cause undefined behavior in the progress updates.

Two possible solutions are (a) replace std::vector with a data structure that maintains pointer stability such as std::list or (b) change the vector to store pointers directly and allocate the objects with an explicit use of the `new` operator.

firestore/src/android/firestore_android.cc
```diff
@@ -515,6 +542,59 @@ void FirestoreInternal::ClearListeners() {
...
+
+  Local<Object> progress_listener =
+      EventListenerInternal::Create(env, this, listener_ptr);
+  task.AddProgressListener(env, user_callback_executor(), progress_listener);
+
+  return promises_->NewFuture<LoadBundleTaskProgress>(env, AsyncFn::kLoadBundle,
+                                                      task);
+}
+
+Future<Query> FirestoreInternal::NamedQuery(const std::string& query_name) {
```
dconeybe, line 587: Add a test that calls `NamedQuery()` with a malformed UTF8 string. This should cause `NewStringUtf()` to fail (throws an exception?). Make sure that this doesn't cause a crash and that the returned `Future` object is "invalid".

firestore/src/android/firestore_android.h
```diff
@@ -192,6 +215,11 @@ class FirestoreInternal {
   std::unordered_set<ListenerRegistrationInternal*> listener_registrations_;
 #endif  //  defined(_STLPORT_VERSION)
 
+  Mutex bundle_listeners_mutex_;
+  // Using a vector to ensure listener instances cannot outlive
+  // FirestoreInternal.
+  std::vector<LambdaEventListener<LoadBundleTaskProgress>> bundle_listeners_;
```
dconeybe, line 221: I think I have a better idea than using this vector to keep the event listeners alive. What if you create them via the `new` operator then add a `delete` for them to the `CleanupNotifier` returned from `cleanup()`?

firestore/src/android/firestore_android.cc
```diff
@@ -206,6 +223,14 @@ enum class InitialLogState {
   kSetDisabled,
 } initial_log_state = InitialLogState::kUnset;
 
+Local<LoadBundleTaskInternal> CreateLoadBundleTask(Env& env,
```
dconeybe, line 226: Add a test that calls `CreateLoadBundleTask()` with a string that is invalid UTF-8 encoding. In this case, it should return an invalid (null?) object.

dconeybe (changes requested):

firestore/src/android/firestore_android.cc
wu-hui: Done.

firestore/src/android/firestore_android.cc
wu-hui: Done.

.gitignore
```diff
@@ -18,3 +18,4 @@ gcs_key_file.json
 
 # Folders for cmake/test output
 *_build/
+*cmake-build-debug/
```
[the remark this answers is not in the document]
dconeybe: Ahh I see. I thought it was just some build directory that you came up with. If it's a CLion default, then this change makes sense.

firestore/integration_test_internal/src/bundle_test.cc
```diff
@@ -0,0 +1,298 @@
...
+          db_deleted.set_value();
+        }
+      });
+
+  // Wait for the notification that the instance is deleted. We cannot rely on
+  // the returned future because it will not complete.
+  db_deleted.get_future().wait();
+
+  // This future is not completed, and returns back a nullptr for result.
+  EXPECT_EQ(Await(result), nullptr);
```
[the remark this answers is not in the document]
dconeybe: The comment a few lines above at line 167 says "We cannot rely on the returned future because it will not complete" and on line 171 says "This future is not completed". If the `Future` is known to not complete, then it does not make sense to await its completion. Are those comment incorrect?

firestore/integration_test_internal/src/bundle_test.cc
```diff
@@ -199,6 +201,26 @@ TEST_F(BundleTest, LoadBundlesForASecondTimeSkips) {
   VerifyQueryResults(db);
 }
 
+TEST_F(BundleTest, LoadInvalidBundlesShouldFail) {
+  Firestore* db = TestFirestore();
+  std::vector<std::string> invalid_bundles{
+      "invalid bundle obviously", "\"(╯°□°）╯︵ ┻━┻\"",
+      "\0\1\2\3\ud7ff\ue000\uffff\""  // random bytes
```
dconeybe, line 208: This new `LoadInvalidBundlesShouldFail` test does not appear to include an invalid utf-8 byte sequence (I tested out the sequences and they were all able to be decoded by Python's utf8 decoder). If this is true, please add an invalid UTF-8 byte sequence to these test scenarios, such as "\xc3\x28" (https://stackoverflow.com/questions/1301402/example-invalid-utf8-string).

firestore/integration_test_internal/src/bundle_test.cc
```diff
@@ -291,6 +313,12 @@ TEST_F(BundleTest, GetInvalidNamedQuery) {
     EXPECT_EQ(future.status(), FutureStatus::kFutureStatusComplete);
     EXPECT_EQ(future.error(), Error::kErrorNotFound);
   }
+  {
+    auto future = db->NamedQuery("\0\1\2\3\4");
```
dconeybe, line 317: This test should have an invalid utf8 byte sequence as well.

firestore/src/android/firestore_android.h
dconeybe: IMO it's better because it's explicit about the memory management. Using a vector/list to indirectly keep objects alive was non-obvious when I read the code. This is an optional comment so feel free to leave things as-is.

dconeybe (approved):

firestore/integration_test_internal/src/util/bundle_builder.cc
```diff
@@ -0,0 +1,227 @@
...
+namespace firestore {
+namespace {
+
+// Copied from
+// https://stackoverflow.com/questions/2896600/how-to-replace-all-occurrences-of-a-character-in-string
+std::string ReplaceAll(std::string str,
+                       const std::string& from,
+                       const std::string& to) {
+  if (from.empty()) return str;
+  size_t start_pos = 0;
```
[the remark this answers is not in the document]
var-const: Note that while technically true, this is very unlikely to matter in practice. GCC and Clang have `string::size_type` be the same type as `size_t` across all versions (at least any semi-recent ones); I'm not aware of a Visual Studio online compiler, but I'd be surprised if it wasn't the case for MSVC as well.

firestore/src/android/firestore_android.cc
var-const: Great catch, Denver! Sorry, I missed that the pointer gets stored for later usage.

var-const (approved):

firestore/src/include/firebase/firestore.h
```diff
@@ -405,6 +406,41 @@ class Firestore {
...
+   * @return A `Future` that is resolved when the loading is either completed,
+   * or aborted due to an error.
+   */
+  virtual Future<LoadBundleTaskProgress> LoadBundle(const std::string& bundle);
+
+  /**
+   * Loads a Firestore bundle into the local cache, with provided callbacks
+   * executed for progress updates.
+   *
+   * @param bundle A string representing the bundle to be loaded.
```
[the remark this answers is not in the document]
markarndt: Yeah, in Android things are clearer since the objects passed to loadBundle obviously ARE the bundle, cast in some data structure. In C++, does this string (a) "identify" the bundle to be loaded, or (b) actually contain the bundle data? If (a) then I'd say "A string identifying the bundle...". If (b), then yes it may be a little clear to say "A string containing the bundle...."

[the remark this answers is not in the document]
markarndt: Ping.

markarndt (approved):

firestore/src/include/firebase/firestore.h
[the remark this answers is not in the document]
wu-hui: Done.

dconeybe (dismissed):

firestore/integration_test_internal/src/bundle_test.cc
```diff
@@ -0,0 +1,335 @@
...
+  // local run is fine. We need to figure out why and re-enable it.
+  SKIP_TEST_ON_WINDOWS;
+
+  Firestore* db = TestFirestore();
+  std::vector<std::string> invalid_bundles{
+      "invalid bundle obviously", "\"(╯°□°）╯︵ ┻━┻\"",
+      "\xc3\x28"  // random bytes
+  };
+  for (const auto& bundle : invalid_bundles) {
+    std::vector<LoadBundleTaskProgress> progresses;
```
dconeybe, line 220: Using a `ScopedTrace` RAII object would be especially useful here to help determine which iteration of the loop failed.

firestore/firestore_resources/build.gradle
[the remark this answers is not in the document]
dconeybe: Did you run this change to `build.gradle` by the FPL team?

dconeybe (approved):
LGTM. I left two minor suggestions for the tests (that can be done in a follow-up PR if you like the suggestions) and also confirm with the FPL team about the `build.gradle` version bumps.

firestore/firestore_resources/build.gradle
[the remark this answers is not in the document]
wu-hui: Confirmed with Anthony.

PR #396, opened 2021-04-26
Applied clang-tidy modernize fixups

alexames:
Applied the clang-tidy rules:

modernize-use-override
modernize-use-nullptr
modernize-use-equals-delete
modernize-use-equals-default
modernize-use-emplace
modernize-pass-by-value
modernize-loop-convert
modernize-deprecated-headers

### DIFF
--- a/admob/src/common/admob_common.cc
+++ b/admob/src/common/admob_common.cc
@@ -133,10 +133,10 @@ FutureCallbackData* CreateFutureCallbackData(FutureData* future_data,
 
 // Non-inline implementation of the Listeners' virtual destructors, to prevent
 // their vtables from being emitted in each translation unit.
-BannerView::Listener::~Listener() {}
-InterstitialAd::Listener::~Listener() {}
-NativeExpressAdView::Listener::~Listener() {}
-rewarded_video::Listener::~Listener() {}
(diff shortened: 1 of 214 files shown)


alexames:
When I ran these I intended to clang-format because these checks didn't always leave the code formatted correctly. However, the little script that I had run accidentally applied the formatting to the subsequent commit. I'm going to squash these all in the end so I don't think it matters though.

There's a lot of files that changed, so when reviewing it's probably easier to look at individual commits. Each commit just has one flavor of change that it's applying so you can see how it affects the whole codebase.

alexames:
I noticed a few changes to files in the `include` subdirectory, but they all seem benign. Even though it's a change to a header file I don't think any of the changes made technically qualify as an API change. If anyone feels differently though I'm happy to revert those files.

jonsimantov:
I'm fairly certain that this PR will have to wait until the PR that removes STLPort goes in.

alexames:
> I'm fairly certain that this PR will have to wait until the PR that removes STLPort goes in.

I'll just abandon this change for the time being and if we decide to set up a clang tidy rule for the repo we can re-run these checks then.

PR #408, opened 2021-05-03
Removed SetCurrentScreen method from Analytics

vimanyu:
Syncing with the removal from underlying native mobile SDKs.

### DIFF
--- a/release_build_files/readme.md
+++ b/release_build_files/readme.md
@@ -365,8 +365,8 @@ code.
 ### 8.0.0
     
 -   Changes
-    -   Analytics: Removed `SetCurrentScreen()` following the removal from underlying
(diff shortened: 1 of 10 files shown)


release_build_files/readme.md
```diff
@@ -365,6 +365,8 @@ code.
 ### 8.0.0
     
 -   Changes
+    -   Analytics: Removed `SetCurrentScreen()` following the removal from underlying
+        native SDKs.
```
jonsimantov, line 369: Maybe add something about how it was deprecated from the native SDKs? Though it's a bummer that we didn't have it marked deprecated ourselves.

DellaBitta (approved):

PR #409, opened 2021-05-04
Create build matrices & test matrices for integration test workflow

sunmou99:
- Divided the matrix into build matrices & test matrices. Now we can test on different devices in parallel.
- Summary log has two parts: build summary & test summary. Changed the summarize script accordingly.
- Both testapps & logs are uploaded as artifacts. retention: 2 days.
- Download iOS Simulator if the device doesn't exist.
- Android Emulator test log is moved to Emulator's internal storage, because some Emulators don't have external storage.
- other small fixes


jonsimantov:
I merged in some pretty major changes to the integration tests workflow, probably worth resolving the conflicts. Note that it uses a non-default ref for checking out the repo now.

PR #410, opened 2021-05-04
Add Firestore source tests

ehsannas:


firestore/integration_test_internal/src/source_test.cc
```diff
@@ -0,0 +1,523 @@
...
+
+TEST_F(SourceTest, TestGetDocumentWhileOnlineWithDefaultGetOptions) {
+  MapFieldValue initialData = {{"key", FieldValue::String("value")}};
+  DocumentReference docRef = DocumentWithData(initialData);
+
+  Future<DocumentSnapshot> future = docRef.Get();
+  WaitFor(future);
+
+  DocumentSnapshot snapshot = *future.result();
+  ASSERT_TRUE(snapshot.exists());
```
jonsimantov, line 24: Is there a reason you're using ASSERT for everything and not EXPECT? ASSERT should really only be used when the test cannot continue if the thing failed.

firestore/integration_test_internal/src/source_test.cc
```diff
@@ -0,0 +1,523 @@
...
+  const std::map<std::string, MapFieldValue> initialDocs = {
+      {"doc1", {{"key1", FieldValue::String("value1")}}},
+      {"doc2", {{"key2", FieldValue::String("value2")}}},
+      {"doc3", {{"key3", FieldValue::String("value3")}}}};
+  CollectionReference colRef = Collection(initialDocs);
+
+  Future<QuerySnapshot> future = colRef.Get();
+  WaitFor(future);
+
+  QuerySnapshot querySnapshot = *future.result();
```
var-const, line 40: Nit: it seems a little inconsistent that tests combine heavily abbreviated variable names (e.g. `col_ref`) with fully spelled out names (e.g. `query_snapshot`). Android uses abbreviated names throughout. I don't mind either, but perhaps choose a consistent approach? If you feel differently, let's discuss.

firestore/integration_test_internal/src/source_test.cc
```diff
@@ -0,0 +1,523 @@
...
+}
+
+TEST_F(SourceTest, TestGetDocumentWhileOfflineWithDifferentGetOptions) {
+  MapFieldValue initialData = {{"key", FieldValue::String("value")}};
+  DocumentReference docRef = DocumentWithData(initialData);
+
+  WaitFor(docRef.Get());
+  WaitFor(TestFirestore()->DisableNetwork());
+
+  // Create an initial listener for this query (to attempt to disrupt the
```
var-const, line 248: Nit: missing word (`disrupt the _gets_ below`).

firestore/integration_test_internal/src/source_test.cc
```diff
@@ -0,0 +1,523 @@
...
+  // continuing.
+  std::promise<Error> errorPromise;
+  std::future<Error> errorFuture = errorPromise.get_future();
+
+  docRef.AddSnapshotListener([&errorPromise](const DocumentSnapshot& snapshot,
+                                             Error error_code,
+                                             const std::string& error_message) {
+    errorPromise.set_value(error_code);
+  });
+  errorFuture.wait();
```
var-const, line 259: Nit: please add a check that the value is not an error (I think Android would raise an exception in the error case, so some kind of an error check is warranted). Applies below as well.

firestore/integration_test_internal/src/source_test.cc
```diff
@@ -0,0 +1,523 @@
...
+  WaitFor(colRef.Get());
+  WaitFor(TestFirestore()->DisableNetwork());
+
+  // Since we're offline, the returned futures won't complete.
+  colRef.Document("doc2").Set({{"key2b", FieldValue::String("value2b")}},
+                              SetOptions().Merge());
+  colRef.Document("doc3").Set({{"key3b", FieldValue::String("value3b")}});
+  colRef.Document("doc4").Set({{"key4", FieldValue::String("value4")}});
+
+  // Create an initial listener for this query (to attempt to disrupt the
```
var-const, line 299: Nit: `gets` is missing here too.

firestore/integration_test_internal/src/source_test.cc
```diff
@@ -0,0 +1,523 @@
...
+  WaitFor(docRef.Delete());
+
+  WaitFor(TestFirestore()->DisableNetwork());
+  Future<DocumentSnapshot> future = docRef.Get();
+  WaitFor(future);
+
+  DocumentSnapshot snapshot = *future.result();
+  ASSERT_FALSE(snapshot.exists());
+  ASSERT_EQ(0, snapshot.GetData().size());
+  // TODO(ehsann): is_from_cache() is false. why?
```
var-const, line 389: It looks like there's a related TODO in the Android test -- can you copy that comment here and mark the test as disabled (it's `@Ignore`d on Android)? (To disable a test, add `DISABLED_` to its name, i.e., it should be `DISABLED_GetDeletedDocWhileOfflineWithDefaultGetOptions`).

var-const (changes requested):

firestore/integration_test_internal/src/source_test.cc
ehsannas: Done

firestore/integration_test_internal/src/source_test.cc
```diff
@@ -0,0 +1,523 @@
...
+
+  WaitFor(TestFirestore()->DisableNetwork());
+  Future<QuerySnapshot> future = colRef.Get(Source::kServer);
+  Await(future);
+  EXPECT_EQ(future.status(), FutureStatus::kFutureStatusComplete);
+  EXPECT_EQ(future.error(), Error::kErrorUnavailable);
+}
+
+}  // namespace firestore
+}  // namespace firebase
```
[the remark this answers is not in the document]
ehsannas: Done

ehsannas (commented):
Thanks for all the pointers. PTAL.

firestore/integration_test_internal/src/source_test.cc
```diff
@@ -13,511 +13,514 @@ namespace firestore {
...
+  std::promise<Error> error_promise;
+  std::future<Error> error_future = error_promise.get_future();
+
+  doc_ref.AddSnapshotListener(
+      [&error_promise](const DocumentSnapshot& snapshot, Error error_code,
+                       const std::string& error_message) {
+        error_promise.set_value(error_code);
+      });
+  // Note that future::get() will wait until the future has a valid result and
+  // retrieves it. Calling wait() before get() is not needed.
```
var-const, line 260: Ultranit: I think it should be `s/retrieves/retrieve/` (the way I read it, it connects to the previous `will`: `will wait... and (will) retrieve it`). Feel free to disagree.

firestore/integration_test_internal/src/source_test.cc
```diff
@@ -0,0 +1,526 @@
+#include <future>
```
var-const, line 1: Nit: please add includes for `std::map` and `std::string`. See go/cstyle#Names_and_Order_of_Includes:

> You should include all the headers that define the symbols you rely upon... If you rely on symbols from `bar.h`, don't count on the fact that you included `foo.h` which (currently) includes `bar.h`: include `bar.h` yourself, unless `foo.h` explicitly demonstrates its intent to provide you the symbols of `bar.h`.

var-const (approved):

firestore/integration_test_internal/src/source_test.cc
ehsannas: Ah, I had enabled auto-merge and it has landed the PR upon your approval. I'll send a follow up PR to address these. Thanks!

PR #411, opened 2021-05-05
Minor fixes to Firestore Source Tests.

ehsannas:


var-const (approved):
Thanks for the follow-up!

PR #412, opened 2021-05-05
Remove the STLPort mention from the docs and mark gnustl as deprecated

var-const:


var-const:
Sorry, missed this one in the previous PR.

jonsimantov (approved):

PR #416, opened 2021-05-06
Disable a test that seems to time out on Android.

ehsannas:
Will investigate and re-enable the test when the cause is found.


var-const:
For the record, here's the link to a failing run: https://github.com/firebase/firebase-cpp-sdk/runs/2516720057

jonsimantov (dismissed):
You can also disable this only on Android by adding SKIP_TEST_ON_ANDROID; as the first line of the function body (macro is defined in firebase_test.h)

var-const (approved):

ehsannas:
Closing this in favor of https://github.com/firebase/firebase-cpp-sdk/pull/417

PR #425, opened 2021-05-13
Provide an easy way to edit the googletest arguments in integration tests

dconeybe:
Android and iOS do not have a way to take command-line arguments as input, and therefore there is no way to specify useful arguments like `--gtest_test_filter` when developing/debugging a specific test. This PR modifies `firebase_test_framework.cc` to at least provide an easy way to hardcode googletest arguments during development.

### DIFF
--- a/testing/test_framework/src/firebase_test_framework.cc
+++ b/testing/test_framework/src/firebase_test_framework.cc
@@ -15,6 +15,9 @@
 #include "firebase_test_framework.h"  // NOLINT
 
 #include <cstdio>
+#include <cstring>
+#include <string>
+#include <vector>
 
 #include "firebase/future.h"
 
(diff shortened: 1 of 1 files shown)


testing/test_framework/src/firebase_test_framework.cc
```diff
@@ -294,7 +297,57 @@ std::ostream& operator<<(std::ostream& os, const Variant& v) {
...
+    args_vector.push_back(argv[i]);
+  }
+
+  // This is where you can add elements to the `args_vector` vector that will be
+  // specified to googletest.
+  // e.g. args_vector.push_back("--gtest_list_tests");
+
+  // Write the elements of the vector back into argv and modify argc.
+  // The memory leaks produced below are acceptable because they would last the
+  // entire lifetime of the application anyways.
```
ehsannas, line 334: More importantly, I would say they are acceptable because this may only be used during development / debugging.

We could alter the code slightly so that it doesn't leak memory except for when someone is adding new flags during dev/debugging process. Meaning, we could add an early exit here:

```cpp
if (args_vector.size() == argc) {
  return argv;
}
```

ehsannas (changes requested):
This is generally useful. I used this method when I was trying to reproduce an issue for a specific test. A minor suggestion inlined.

Also a couple of questions for my own benefit:

Are the changes clang-formatted? (If not, does GitHub automatically format it before merging?)

How is this `firebase_test_framework` invoked? The cleaner approach seems to be to pass `--gtest_filter` to that invocation (so that it's already in `argv` when we get to `common_main`)

dconeybe:
> Are the changes clang-formatted? (If not, does GitHub automatically format it before merging?)
Yes, the file has been formatted with `clang-format`. GitHub does not automatically format but there is a presubmit check that verifies that code is formatted correctly and also an (optional) commit hook that you can install locally. See https://github.com/firebase/firebase-cpp-sdk/pull/380 for details.

> How is this `firebase_test_framework` invoked? The cleaner approach seems to be to pass `--gtest_filter` to that invocation (so that it's already in `argv` when we get to `common_main`)
It would definitely be cleaner to just specify arguments on the command line. You can indeed do this when running the desktop tests. However, Android has no means of specifying command-line arguments and we have not plumbed through the command-line arguments in iOS (if it's even possible). There are probably some platform-specific mechanisms that could be used to specify the arguments without having to edit source code; however, this is just a quick-and-dirty way to get it done.

ehsannas:
Got it. Thanks

testing/test_framework/src/firebase_test_framework.cc
dconeybe: Good idea. I've implemented this suggestion and refactored the code a bit. PTAL.

ehsannas (approved):

PR #426, opened 2021-05-13
Fix Firestore transaction_extra_test.cc on iOS and re-enable the tests

dconeybe:
The tests were hanging indefinitely and therefore were disabled. The root cause was the call to `Await()` in the transaction callback thread, which eventually called `ProcessEvents()` in `ios_app_framework.mm`. The `ProcessEvents()` function was coded in such a way that it would only handle being called from the main thread, and, therefore, when called from a non-main thread, caused undefined behavior (in this case, a hang). I re-wrote the concurrency logic in `ios_app_framework.mm` and that fixed the tests.

Previously, this code would only lock `g_shutdown_signal` once, before calling `common_main()`. Let's say the thread that does this is T1. `ProcessEvents()` requires the calling thread to possess the lock, which is only true if it is called from T1; if `ProcessEvents()` is called from any other thread, it will hang (because it can never acquire the lock).


testing/sample_framework/src/ios/ios_app_framework.mm
```diff
@@ -47,10 +47,19 @@ @interface FTAViewController : UIViewController
...
+  kShuttingDown,
+  kShutDown
+};
+
+// Note: g_running_status and g_exit_status must only be accessed while holding the lock from
+// g_running_status_condition; also, any changes to these values should be followed up with a
+// call to [g_running_status_condition broadcast].
+static NSCondition *g_running_status_condition;
+static RunningStatus g_running_status = RunningStatus::kRunning;
+static int g_exit_status = -1;
```
var-const, line 61: Nit: it seems a little inconsistent that `status` is used in two closely related variables but has slightly different meanings. How about `s/RunningStatus/RunningState/`?

testing/sample_framework/src/ios/ios_app_framework.mm
```diff
@@ -70,9 +79,14 @@ - (void)viewDidLoad {
     char *argv[1];
     argv[0] = new char[strlen(TESTAPP_NAME) + 1];
     strcpy(argv[0], TESTAPP_NAME);  // NOLINT
-    [g_shutdown_signal lock];
-    g_exit_status = common_main(1, argv);
-    [g_shutdown_complete signal];
+    auto common_main_result = common_main(1, argv);
```
var-const, line 82: Nit: consider spelling out the type here.

testing/sample_framework/src/ios/ios_app_framework.mm
```diff
@@ -87,9 +101,20 @@ - (void)viewDidLoad {
...
 bool ProcessEvents(int msec) {
-  [g_shutdown_signal
-      waitUntilDate:[NSDate dateWithTimeIntervalSinceNow:static_cast<float>(msec) / 1000.0f]];
-  return g_shutdown;
+  NSDate* endDate = [NSDate dateWithTimeIntervalSinceNow:static_cast<float>(msec) / 1000.0f];
+  [g_running_status_condition lock];
+
+  if (g_running_status == RunningStatus::kRunning) {
+    NSDate* currentDate = [NSDate date];
+    if ([currentDate compare:endDate] == NSOrderedAscending) {
```
var-const, line 109: Question: what is the reason to check this manually? Is there any advantage to calling `wait` unconditionally and letting its implementation figure out if enough time has passed?

var-const (changes requested):
Overall looking good! Just some nits.

var-const (approved):
Consider expanding the rationale in the PR description -- some of the details you shared over chat seem very useful in general. Perhaps something like (feel free to change as necessary):
```
Previously, this code would only lock `g_shutdown_signal` once, before calling `common_main`. Let's say the thread that does this is `T1`. `ProcessEvents` requires the calling thread to possess the lock, which is only true if it is called from `T1`; if `ProcessEvents` is called from any other thread, it will hang (because it can never acquire the lock).
```

dconeybe:
> Consider expanding the rationale in the PR description

Done. I just copied your suggestion verbatim :)

PR #429, opened 2021-05-17
Fix GetPlatformAppByName() crash on KitKat when the app was not found

dconeybe:
This PR fixes a bug in `GetPlatformAppByName()` if an exception is thrown by the Java method call. The issue is that if an exception is thrown then the return value from `CallStaticObjectMethod()` is `NULL`... *except* on KitKat, where it appears to return garbage instead. This garbage value was being returned and the caller was incorrectly treating it as a valid object reference. The fix is to check if an exception was thrown and explicitly returning `NULL` in that case, instead of relying on the return value from `CallStaticObjectMethod()` being `NULL`.

This bug surfaced in the integration test `FirestoreIntegrationTest.CanPageThroughItems` which was crashing on KitKat. The root cause was that it was using a custom app name and the call to `GetPlatformAppByName()` was returning garbage in that case.

https://github.com/firebase/firebase-cpp-sdk/blob/4a34eddf71866ec230b9d7b92c86814c9795161b/app/src/app_android.cc#L302-L317

https://github.com/firebase/firebase-cpp-sdk/blob/4a34eddf71866ec230b9d7b92c86814c9795161b/firestore/integration_test_internal/src/cursor_test.cc#L19-L26

### DIFF
--- a/app/src/app_android.cc
+++ b/app/src/app_android.cc
@@ -311,7 +311,13 @@ static jobject GetPlatformAppByName(JNIEnv* jni_env, const char* name) {
         name_string);
     jni_env->DeleteLocalRef(name_string);
   }
-  jni_env->ExceptionCheck();
+  if (jni_env->ExceptionCheck()) {
+    // Explicitly set `platform_app` to `NULL` if an exception was thrown
+    // because on KitKat (API 19) `CallStaticObjectMethod()` may return garbage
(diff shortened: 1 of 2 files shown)


alexames (approved):

dconeybe:
Note: I'm going to ignore the test failures because they are all flakes that could not possibly be triggered by this PR.

PR #431, opened 2021-05-18
Fix Firestore's snapshot_metadata_android_test

dconeybe:
For some reason, the call to `env.FindClass()` was failing because it couldn't find the class. So I updated the code to use our new, modern JNI wrapper and it all worked. I'm not sure why but I'm happy with the result.

### DIFF
--- a/firestore/integration_test_internal/src/android/snapshot_metadata_android_test.cc
+++ b/firestore/integration_test_internal/src/android/snapshot_metadata_android_test.cc
@@ -3,6 +3,7 @@
 #include "android/firestore_integration_test_android.h"
 #include "firebase_test_framework.h"
 #include "firestore/src/include/firebase/firestore/snapshot_metadata.h"
+#include "firestore/src/jni/declaration.h"
 #include "firestore/src/jni/env.h"
 #include "firestore_integration_test.h"
 #include "gmock/gmock.h"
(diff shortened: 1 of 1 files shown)


wu-hui (approved):
The PR looks good.

But have you tried to figure out what are being executed btw the two different implementation?

dconeybe:
> But have you tried to figure out what are being executed btw the two different implementation?

No, I did not figure out what was being done differently. Although I'm curious and would like to dig into it, I feel that my time is better spent elsewhere.

dconeybe:
> > But have you tried to figure out what are being executed btw the two different implementation?
> 
> No, I did not figure out what was being done differently. Although I'm curious and would like to dig into it, I feel that my time is better spent elsewhere.

I think I know why the call to `env.FindClass()` was failing. It's a known issue/limitation of the Android JNI APIs where it was using a ClassLoader that only had the built-in standard libraries available. The workaround would have been to use the ClassLoader retrieved from another class. This logic is already implemented in a helper function of whose existence I was unaware at the time: `app_framework::FindClass()`: https://github.com/firebase/firebase-cpp-sdk/blob/892560304722dd4debdf9fcaba5e8aa986bbc860/testing/sample_framework/src/android/android_app_framework.cc#L82

So I probably could have just swapped out the call to `env.FindClass()` with `app_framework::FindClass()`.

PR #436, opened 2021-05-26
Converted RegistrationIntentService to a JobIntentService

alexames:
There appears to be a hard-to-reproduce hang that in the
RegistrationIntentService, which is probably related to it being an
IntentService instead of a JobIntentService.

### DIFF
--- a/messaging/src/android/java/com/google/firebase/messaging/cpp/RegistrationIntentService.java
+++ b/messaging/src/android/java/com/google/firebase/messaging/cpp/RegistrationIntentService.java
@@ -14,7 +14,6 @@
 
 package com.google.firebase.messaging.cpp;
 
-import android.app.IntentService;
 import android.content.Context;
 import android.content.Intent;
 import com.google.firebase.iid.FirebaseInstanceId;
@@ -28,7 +27,7 @@ import java.nio.channels.FileLock;
  * A class that manages Registration Token generation and passes the generated tokens to the native
  * OnTokenReceived function.
  */
-public class RegistrationIntentService extends IntentService {
+public class RegistrationIntentService extends JobIntentService {
   private static final String TAG = "FirebaseRegService";
 
   public RegistrationIntentService() {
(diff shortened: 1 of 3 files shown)


messaging/src/android/java/com/google/firebase/messaging/cpp/RegistrationIntentService.java
```diff
@@ -37,11 +36,20 @@ public RegistrationIntentService() {
     super(TAG);
   }
 
+  /**
+   * Convenience wrapper over enqueueWork to either directly start the service (when running on
+   * pre-O platforms) or enqueue work for it as a job (when running on Android O and later).
+   */
+  public static void enqueueWork(Context context, Intent intent) {
```
a-maurice, line 43: Is this getting called from somewhere else?

messaging/src/android/java/com/google/firebase/messaging/cpp/RegistrationIntentService.java
```diff
@@ -28,7 +27,7 @@
  * A class that manages Registration Token generation and passes the generated tokens to the native
  * OnTokenReceived function.
  */
-public class RegistrationIntentService extends IntentService {
+public class RegistrationIntentService extends JobIntentService {
```
DellaBitta, line 30: The CI tests are failing and I think it might be becuase you need to import the JobInentService here.

DellaBitta (changes requested):

messaging/src/android/java/com/google/firebase/messaging/cpp/RegistrationIntentService.java
alexames: This was included by mistake. I've removed it

messaging/src/android/java/com/google/firebase/messaging/cpp/RegistrationIntentService.java
alexames: Fixed

DellaBitta (approved):

PR #443, opened 2021-06-02
Feature/tvos support

vimanyu:
This is specifically for building tvOS libraries for both device and simulator. For testing the built libraries, there will be another PR that ensures all integration tests are working.

- single toolchain file that could possibly be used in the future for iOS as well as mac builds. Currently being used only for tvOS.
- Disabled AdMob and Dynamic Links as they are not supported on tvOS.
- Disabled some functionality from auth (PhoneAuthProvider) that doesn't exist on tvOS.
- Disabled some functionality in FCM (since some of the frameworks we are using do not exist on tvOS). FCM might require more work for full functionality on tvOS but it doesn't block the builds. 

Easiest way to build for tvos (including xcframeworks) is to use the helper bash scripts,
`build_scripts/tvos/build.sh`


build_scripts/tvos/install_prereqs.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash -e
+
+if [[ $(uname) != "Darwin" ]]; then
+    echo "Unsupported platform, iOS can only be build on a MacOS machine."
```
DellaBitta, line 4: tvOS, not iOS

build_scripts/tvos/package.sh
```diff
@@ -0,0 +1,56 @@
+#!/bin/bash -e
+
```
DellaBitta, line 2: This should have some sort of header, like our other package scripts.

cmake/toolchains/apple.toolchain.cmake
```diff
@@ -0,0 +1,926 @@
+# This file is part of the ios-cmake project. It was retrieved from
```
DellaBitta, line 1: I think you should add the Google copyright to the top of this file, too, and also create a entry for this in the `release_build_files/NOTICES` file.

DellaBitta (changes requested):

app/src/app_common.cc
```diff
@@ -88,7 +88,7 @@ const char* kCpuArchitecture = "arm32";
 
 #elif defined(__APPLE__)
 const char* kCppRuntimeOrStl = "libcpp";
-#if FIREBASE_PLATFORM_IOS
+#if FIREBASE_PLATFORM_IOS || FIREBASE_PLATFORM_TVOS
 const char* kOperatingSystem = "ios";
```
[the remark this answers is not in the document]
vimanyu: Great catch and thank you for spotting this!

vimanyu:
Thanks for the comments! All have been addressed!

build_scripts/tvos/package.sh
```diff
@@ -0,0 +1,65 @@
...
+    echo "Usage: $0 <built tvOS SDK path> <path to put packaged files into>
+example:
+  build_scripts/tvos/package.sh tvos_build package_out"
+}
+
+builtpath=$1
+packagepath=$2
+
+if [[ -z "${builtpath}" || -z "${packagepath}" ]]; then
+    echo "Usage: $0 <built tvOS SDK path> <path to put packaged files into>"
```
DellaBitta, line 16: call usage() here intead.

build_scripts/tvos/package.sh
```diff
@@ -0,0 +1,65 @@
...
+    exit 1
+fi
+
+if [[ ! -d "${builtpath}/xcframeworks" ]]; then
+    echo "Built tvOS SDK not found at path '${builtpath}'."
```
DellaBitta, line 21: Call usage() here, too.

DellaBitta (changes requested):

build_scripts/tvos/package.sh
vimanyu: oops! Missed it! Done!

DellaBitta (approved):

PR #445, opened 2021-06-03
Database URLs are no longer necessary.

alexames:
Made several fields in plist files optional. Most notably the
DATABASE_URL, because that was causing problems for users building for
iOS, but also the other fields like the STORAGE_BUCKET.

### DIFF
--- a/generate_xml_from_google_services_json.py
+++ b/generate_xml_from_google_services_json.py
@@ -113,9 +113,9 @@ def update_dict_keys(key_map, input_dict):
 
   Two dictionaries are passed to this function: the key_map that represents a
   mapping of source keys to destination keys, and the input_dict that is the
-  dictionary that is to duplicated, replacing any key that matches a source key
-  with a destination key. Source keys that are not present in the input_dict
-  will not have their destination key represented in the result.
+  dictionary that is to be duplicated, replacing any key that matches a source
+  key with a destination key. Source keys that are not present in the
+  input_dict will not have their destination key represented in the result.
 
   In other words, if key_map is `{'old': 'new', 'foo': 'bar'}`, and input_dict
   is `{'old': 10}`, the result will be `{'new': 10}`.
--- a/generate_xml_from_google_services_json.py
+++ b/generate_xml_from_google_services_json.py
@@ -121,12 +121,14 @@ def update_dict_keys(key_map, input_dict):
   is `{'old': 10}`, the result will be `{'new': 10}`.
 
   Args:
-    key_map: A dictionary of source keys to destination keys.
-    input_dict: The dictionary that is to be duplicated, replacing source keys
-      with the corresponding destination keys from key_map.
+    key_map: A dictionary of strings to strings that maps source keys to
+      destination keys.
+    input_dict: The dictionary of string keys to any value type, which
+      is to be duplicated, replacing source keys with the corresponding
+      destination keys from key_map.
 
   Returns:
-    A new map with updated keys.
+    A new dictionary with updated keys.
   """
   return {
       new_key: input_dict[old_key]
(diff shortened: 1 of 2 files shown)


alexames:
In addition to this change, we need to update the .exe file that goes along with these. See the readme [here](https://github.com/firebase/firebase-cpp-sdk/blob/main/readme_for_generate_xml_from_google_services_json_exe.txt). Unfortunately, I don't have access to my Windows computer at the moment. @DellaBitta, Can I get you to build that for me and upload the changes here?

generate_xml_from_google_services_json.py
```diff
@@ -108,6 +108,33 @@ def construct_plist_dictionary(xml_root):
...
+    input_dict: The dictionary that is to be duplicated, replacing source keys
+      with the corresponding destination keys from key_map.
+
+  Returns:
+    A new map with updated keys.
+  """
+  return {
+      new_key: input_dict[old_key]
+      for (old_key, new_key) in key_map.items()
+      if old_key in input_dict
```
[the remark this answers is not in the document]
alexames: Yeah, this is a dictionary comprehension. I'll loop in VJ

generate_xml_from_google_services_json.py
```diff
@@ -108,6 +108,33 @@ def construct_plist_dictionary(xml_root):
...
+  mapping of source keys to destination keys, and the input_dict that is the
+  dictionary that is to duplicated, replacing any key that matches a source key
+  with a destination key. Source keys that are not present in the input_dict
+  will not have their destination key represented in the result.
+
+  In other words, if key_map is `{'old': 'new', 'foo': 'bar'}`, and input_dict
+  is `{'old': 10}`, the result will be `{'new': 10}`.
+
+  Args:
+    key_map: A dictionary of source keys to destination keys.
```
vimanyu, line 124: Though it is obvious here but it is always a good idea to put the argument type in the docstring.
Also, any reason why we want to call one thing a map and other a dictionary?

generate_xml_from_google_services_json.py
```diff
@@ -108,6 +108,33 @@ def construct_plist_dictionary(xml_root):
...
+    input_dict: The dictionary that is to be duplicated, replacing source keys
+      with the corresponding destination keys from key_map.
+
+  Returns:
+    A new map with updated keys.
```
vimanyu, line 129: Same: Return type as in google style doc strings.

generate_xml_from_google_services_json.py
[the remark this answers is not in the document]
vimanyu: Looks good! Clever way of updating the dictionaries!

generate_xml_from_google_services_json.py
```diff
@@ -108,6 +108,33 @@ def construct_plist_dictionary(xml_root):
   return plist_dict
 
 
+def update_dict_keys(key_map, input_dict):
+  """Creates a dict from input_dict with the same values but new keys.
+
+  Two dictionaries are passed to this function: the key_map that represents a
+  mapping of source keys to destination keys, and the input_dict that is the
+  dictionary that is to duplicated, replacing any key that matches a source key
```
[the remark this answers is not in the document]
alexames: Fixed

vimanyu (commented):
Made some minor comments. Looks good otherwise.

generate_xml_from_google_services_json.py
alexames: I don't have a good reason for the names, other than it sounded natural. The first argument maps old keys to new keys, the other is just some arbitrary dictionary. I could call it `key_dict` or `key_map_dict` if you prefer, or rename `input_dict` to just `input` or something. 

Naming is hard.

generate_xml_from_google_services_json.py
alexames: Done

generate_xml_from_google_services_json.py
vimanyu: Yup, this one is tricky. I think it does sound natural. Let's leave it as it is.

vimanyu (approved):
LGTM

PR #446, opened 2021-06-04
Rationalize C++ source layout

ehsannas:
See b/174877213 for details and doc.


ehsannas:
I have broken the PR down to smaller contained commits that can be reviewed.

The last commit is just the result of running the `scripts/format_code.py` script.

var-const (changes requested):
A couple of points:

1.
`grep` still shows a few mentions of `_ios` (including those in comments):
```
$ grep -RI _ios firestore

./integration_test_internal/src/field_value_test.cc:#include "firestore/src/ios/field_value_ios.h"
./integration_test_internal/src/util/integration_test_util.cc:#include "firestore/src/ios/firestore_ios.h"
./integration_test_internal/src/firestore_test.cc:  // Also, the logic in `util_ios.h` can be modified to make sure that
./src/android/field_value_android.h:  // these constructors must match the equivalents in field_value_ios.h.
```
(omitting some irrelevant stuff)

2.
I would urge you to import this change _before_ merging it (I think Copybara should be able to handle an unmerged branch) and doing all the necessary changes to the `BUILD` files, just in case it uncovers any issues with this PR.

3.
Consider reverting the formatting changes, they increase the delta without being directly related to the theme of this PR. I'd prefer to have those committed via a separate PR.

ehsannas:
(1) Uh, that's a great catch! Thanks. I have addressed things that were found via grep. Done.
(3) I have dropped the formatting commit (so the check_format bot will fail). Done.
(2) I'll focus on this, and once I'm done I'll assign the PR back to you.
Thanks

ehsannas:
I got a successful import (a few minor changes were needed to internal files) via cl/377574378

ehsannas:
The integration test failures seemed random/flakey 😢  (and different tests were failing on different platforms). I've restarted the GitHub Actions...

ehsannas:
Integration tests passed 👍

PR #451, opened 2021-06-08
Fix GHA CI Windows Path Limits

DellaBitta:
GHA Windows runners have an enviornment option `LongPathsEnabled` which defaults to true (1).  

This can be tested by running  ```(Get-ItemProperty "HKLM:System\CurrentControlSet\Control\FileSystem").LongPathsEnabled``` in a Windows runner.

This value isn't honored within bash shells on Windows and the path limit caused build errors in the Packaging Build workflow for Firestore targets.

This change removes 'bash' as the shell when building and instead executes`python scripts/gha/build_desktop.py` via the host operating system's default shell.


alexames (approved):
lgtm

vimanyu (approved):
LGTM

PR #452, opened 2021-06-09
Temporarily remove `NamedQuery` from the public API

var-const:
See b/190545340.


jonsimantov (approved):
Approving assuming tests pass.

jonsimantov:
Failed test looks like just a timeout - untagging.

PR #454, opened 2021-06-09
Git Patch BoringSSL to disable Windows build warnings/errors

DellaBitta:
This change fixes the boringSSL sources so that they no longer causes warnings on `windows-latest` runners.  

- Switched the boringSSL dependency to a git checkout instead of a tarball download.
- Created a git patch to alter boringSSL's cmake files, adding C4255 as a warning that shall be suppressed.
- Updated GHA configuration to use windows-latest again.  This was previously changed to windows-2016 since those tools didn't report C4255 as a warning. See #444 for more information about these reverted changes.

Other tests:
 - [Packaging Test Run](https://github.com/firebase/firebase-cpp-sdk/actions/runs/922541062)


vimanyu (approved):
LGTM

DellaBitta:
Tests failed during a download of the cloud sdk and not in any of the build steps.  Packaging flow was fully succesful.

PR #462, opened 2021-06-11
format_code.py changed to print the diff when not formatting files inline

dconeybe:
This PR modifies `format_code.py` to print a unified diff of incorrectly-formatted files when both `-noformat_file` and `-verbose` are specified. This can make it easier to correct formatting problems based on the output from GitHub Actions without having to run `clang-format` yourself. This is especially useful if the version of `clang-format` on the GitHub Actions runners differs from that on your local machine, as they may produce different results.

Here is an example output of `format_code.py`:

```
$ python3 scripts/format_code.py -noformat_file -verbose -f example.cc                  
Found 1 file(s). Checking their format.
  - Requires reformatting: "example.cc"
------ BEGIN FORMATTING DIFF OF example.cc
---
+++
@@ -1,9 +1,9 @@
+#include <string>
 #include <vector>
-#include <string>

 std::vector<std::string> GetStrings() {
-    std::vector<std::string> v;
+  std::vector<std::string> v;
   v.emplace_back("string1");
-  v.emplace_back( "string2" );
+  v.emplace_back("string2");
   return v;
 }
------ END FORMATTING DIFF OF example.cc
```

### DIFF
--- a/scripts/format_code.py
+++ b/scripts/format_code.py
@@ -60,8 +60,8 @@ Used to filter out results when searching across directories or git diffs.
 
 # Functions:    
 def get_formatting_diff_lines(filename):
-  """Executes clang-format on the file and returns a human-friendly string
-  containing the diff of the original file against the formatted file.
+  """Calculates and returns a printable diff of the formatting changes that
+  would be applied to the given file by clang-format.
 
   Args:
     filename (string): path to the file whose formatting diff to calculate.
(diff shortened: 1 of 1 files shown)


scripts/format_code.py
```diff
@@ -28,6 +28,7 @@
      because -noformat_file was set.
    2: The application wasn't configured properly and could not execute.
 """
+import difflib
```
DellaBitta, line 31: is this in python by default or do we need to instruct users to install it for this script?

dconeybe: It's part of the Python standard library.

https://docs.python.org/3/library/difflib.html

DellaBitta (approved):

PR #466, opened 2021-06-15
Feature/tvos ios one build script

vimanyu:
Single build script that builds for ios and tvos. 
```
python3 scripts/gha/build_ios_tvos.py

Build for iOS and tvOS.

optional arguments:
  -h, --help            show this help message and exit
  -b BUILD_DIR, --build_dir BUILD_DIR
                        Name of build directory.
  -s SOURCE_DIR, --source_dir SOURCE_DIR
                        Directory containing source code (top level CMakeLists.txt)
  -p PLATFORM [PLATFORM ...], --platform PLATFORM [PLATFORM ...]
                        List of platforms to build for.
  -a ARCHITECTURE [ARCHITECTURE ...], --architecture ARCHITECTURE [ARCHITECTURE ...]
                        List of architectures to build for.
  -t TARGET [TARGET ...], --target TARGET [TARGET ...]
                        List of CMake build targets
  -o OS [OS ...], --os OS [OS ...]
                        List of operating systems to build for.
```

In addition to building frameworks for all os/platform/architecture combinations, it will also build "universal" frameworks and "xcframeworks".
The xcframeworks being generated contain both ios and tvos libraries. 

Other notable change is a new "merged" Info.plist template file that contains entries for both ios and tvos.


jonsimantov:
Is there still a build script in build_scripts/ios?

vimanyu:
> Is there still a build script in build_scripts/ios?

For now, yes as our documentation needs to be updated too. Infact, there are build scripts specifically for tvos. 
I will clean things up after ensuring the new xcframeworks work as expected.

scripts/gha/build_ios_tvos.py
```diff
@@ -0,0 +1,579 @@
...
+
+
+def parse_cmdline_args():
+  parser = argparse.ArgumentParser(description='Build for iOS and tvOS.')
+  parser.add_argument('-b', '--build_dir',
+    default='ios_tvos_build', help='Name of build directory.')
+  parser.add_argument('-s', '--source_dir',
+    default=os.getcwd(),
+    help='Directory containing source code (top level CMakeLists.txt)')
+  parser.add_argument('-p', '--platform', nargs='+',
```
sunmou99, line 541: "platform" have a different meaning in our project. Could you provide a different name here?

sunmou99 (commented):
argparse is good. But I wasn't sure if you noticed Google's practice of [Command line parsing](https://engdoc.corp.google.com/eng/doc/devguide/py/common_modules.md?cl=head#command-line-parsing)

scripts/gha/build_ios_tvos.py
vimanyu: Good point. iOS plist files use the term "SupportedPlatformVariant". Do you think '-variant' might be good name?

vimanyu:
> argparse is good. But I wasn't sure if you noticed Google's practice of [Command line parsing](https://engdoc.corp.google.com/eng/doc/devguide/py/common_modules.md?cl=head#command-line-parsing)

I personally feel the benefit of using a standard library module argparse for simple standalone scripts like this is that we do not have any unnecessary external dependencies.

scripts/gha/build_ios_tvos.py
sunmou99: I think anything other than "platform" and easy to understand would work. like "ios_ platform", "ios_device_type". (I'm bad at names😅)

sunmou99:
> > argparse is good. But I wasn't sure if you noticed Google's practice of [Command line parsing](https://engdoc.corp.google.com/eng/doc/devguide/py/common_modules.md?cl=head#command-line-parsing)
> 
> I personally feel the benefit of using a standard library module argparse for simple standalone scripts like this is that we do not have any unnecessary external dependencies.

Just wanna share that "absl" is an option and is used in our project.
And sure, I agree with you that it's good not to unnecessary external dependencies.

scripts/gha/build_ios_tvos.py
```diff
@@ -0,0 +1,578 @@
...
+It does the following,
+- Build ios and tvos libraries via cmake (after downloading corresponding
+cocoapods)
+- Create a "universal" framework if all platforms and architectures were built.
+- Create "xcframeworks" combining both ios and tvos frameworks under one roof.
+
+Usage examples:
+# Build all supported targets and architectures on all platforms.
+python3 scripts/gha/build_ios_tvos.py
+
```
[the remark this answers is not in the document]
vimanyu: Done!

scripts/gha/build_ios_tvos.py
vimanyu: Renamed to "--platform_variant" as per the iOS plist files.

scripts/gha/build_ios_tvos.py
```diff
@@ -0,0 +1,581 @@
...
+    default=os.getcwd(),
+    help='Directory containing source code (top level CMakeLists.txt)')
+  parser.add_argument('-v', '--platform_variant', nargs='+',
+    default=('device', 'simulator'),
+    help='List of platforms to build for.')
+  parser.add_argument('-a', '--architecture', nargs='+',
+    default=('arm64', 'armv7', 'x86_64', 'i386'),
+    help='List of architectures to build for.')
+  parser.add_argument('-t', '--target', nargs='+',
+    default=('firebase_admob', 'firebase_analytics',
```
sunmou99, line 550: I wonder if we could reuse the "Config" variable.

sunmou99 (commented):
The iOS build workflow takes longer than before: https://github.com/firebase/firebase-cpp-sdk/actions/workflows/ios.yml
Originally, the builds for different architectures are in subprocesses. 
Could you add similar features here with Thread?

vimanyu:
> The iOS build workflow takes longer than before: https://github.com/firebase/firebase-cpp-sdk/actions/workflows/ios.yml
> Originally, the builds for different architectures are in subprocesses.
> Could you add similar features here with Thread?

Thanks for pointing this out. I did some quick tests to see what works best and I think multiprocessing is working great. For debugging purpose, threading/multiprocess makes it a little hard as the log output will be interleaved but we can look into it once it becomes a bigger issue.

Options (and local tests from macbook pro with 12 logical cores):
Comparison with current state of things (building everything serially):  29 minutes
1. Use `cmake --build -j <cpu_count>`
This options is great as just the build for a specific architecture is parallelized by cmake but it is very slightly slower than multiprocessing.
Time: 9m 32 s

2. Multithread
Time: 10m

3. Multiprocess
Time: 7m 56s

Since github runners have only 3 cores, these times are different but in general multiprocessing is definitely faster.

scripts/gha/build_ios_tvos.py
```diff
@@ -0,0 +1,604 @@
...
+
+Usage examples:
+# Build all supported targets and architectures on all platforms.
+python3 scripts/gha/build_ios_tvos.py
+
+# Build specific targets
+python3 scripts/gha/build_ios_tvos.py -t firebase_auth firebase_database
+
+# Build for specific architectures
+python3 scripts/gha/build_ios_tvos.py -p arm64 -t firebase_remote_config
```
sunmou99, line 35: -a arm64

scripts/gha/build_ios_tvos.py
```diff
@@ -0,0 +1,604 @@
...
+  parser.add_argument('-s', '--source_dir',
+    default=os.getcwd(),
+    help='Directory containing source code (top level CMakeLists.txt)')
+  parser.add_argument('-v', '--platform_variant', nargs='+',
+    default=('device', 'simulator'),
+    help='List of platforms to build for.')
+  parser.add_argument('-a', '--architecture', nargs='+',
+    default=('arm64', 'armv7', 'x86_64', 'i386'),
+    help='List of architectures to build for.')
+  parser.add_argument('-t', '--target', nargs='+',
```
sunmou99, line 572: added flag -t firebase_remote_config
Generate empty folders for other xcframeworks

.github/workflows/ios.yml
```diff
@@ -55,4 +55,4 @@ jobs:
 
       - name: Build SDK
         run: |
-          build_scripts/ios/build.sh -b ios_build -s .
+          python scripts/gha/build_ios_tvos.py -b ios_tvos_build -s .
```
sunmou99, line 58: Also, update the file name & workflow name.

sunmou99 (commented):
What's the plan about tvos? 
If add tvos to the packaging workflow, probably need to update the release note later.

vimanyu:
> What's the plan about tvos?
> If add tvos to the packaging workflow, probably need to update the release note later.

Thinking more about it, I think I will revert changes to github workflow files for now. It definitely helped to test the packaging workflow and make sure nothing broke in the process but we should truly include tvos once all the work on simulators on github is done.

scripts/gha/build_ios_tvos.py
vimanyu: Fixed.

.github/workflows/ios.yml
vimanyu: Will take care of this in a new PR since I will be removing the workflow files from this one.

scripts/gha/build_ios_tvos.py
vimanyu: Good idea but I think having the list here explicitly makes it obvious for users of this standalone script. In addition, the `build_testapps.json` contains config for settings like "test_lab". It might make sense to use that for integration testing.

vimanyu:
Have removed changes to workflow files from this PR as we will wait until the work on tvos simulator integration tests on Github is done. 
Though, this script was put through a test in cpp-packaging workflow and worked,
https://github.com/firebase/firebase-cpp-sdk/actions/runs/965593738

sunmou99 (approved):
LGTM

PR #467, opened 2021-06-15
Fix for headers not copied over to firebase.framework while building ios/tvos

vimanyu:
There is a step that copies over all the firebase products include headers over to `firebase.framework`. 
This step is triggered after looking for the word "firebase" in the current path. However, if the current path (checkout of repo) already has "firebase" sub directory anywhere else, this logic fails and headers are not copied over.

For example,
Building ios/tvos `./build_scripts/tvos/build.sh` from these 2 checkout locations will have different results
```
# Headers won't be copied
~/foo/firebase/firebase-cpp-sdk/ 

# Headers will be copied
~/foo/firebase_bar/firebase-cpp-sdk
```


sunmou99 (approved):
GTLM

sunmou99 (changes requested):
Took a second look. "src/include/firebase/" doesn't cover all headers. e.g.
https://github.com/firebase/firebase-cpp-sdk/blob/42137bf462bbf518a20789fee08f3259ad712ab7/app/CMakeLists.txt#L486

The solutions could be: 1) search 'include/firebase/' in the path 2) or search "firebase/" backwards.  3) or add one more step to cover the Firestore headers.

sunmou99:
https://github.com/firebase/firebase-cpp-sdk/blob/42137bf462bbf518a20789fee08f3259ad712ab7/app/CMakeLists.txt#L486
Double checked, these files are included in "xcframework".
"framework" is for macOS Desktop, thus doesn't include these files.

vimanyu:
> https://github.com/firebase/firebase-cpp-sdk/blob/42137bf462bbf518a20789fee08f3259ad712ab7/app/CMakeLists.txt#L486
> 
> 
> Double checked, these files are included in "xcframework".
> "framework" is for macOS Desktop, thus doesn't include these files.

Thanks for flagging this! Changed the pattern to "/include/firebase" to catch the Firestore headers.

sunmou99 (approved):
LGTM

cynthiajoan (approved):

PR #468, opened 2021-06-15
Prevent Windows GHA CI disk exhausation

DellaBitta:
Windows BoringSSL builds are now larger due to #454. The checkout & build partition **(D:)** hits capacity when attempting to install the Google Cloud SDK, which is what we use to upload the final integraion_test build artifacts to storage.

This change moves the arhives destined for GCS to the **C:** drive, then installs the Cloud SDK after they're off of **D:**.  The migration of the integration tests only occurs on Windows, for now.

Some notes:
- We can’t set the build_testapps.py’s  --output_directory flag to directly output to the **C:** directory because python’s relpath operations fail when working across different drives.
- The **github.workspace** env var can be changed to override check-out & installation paths, but the [GHA team notes potential unknown side effects](https://github.community/t/is-it-possible-to-change-github-workspace/136094/3). Therefore I didn’t test this.
- We couldn't  use **mv** the results to **C:** as Firestore's build results include symlinks.  Using **tar** to ferry the files instead.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -374,7 +374,7 @@ jobs:
         # Copy the artifacts and also the upload scripts since python's relpath
         # produces errors when called across two different drives. Use tar to
         # move the files as it neatly deepcopies firestore's symlinks.
-        if:  ${{ (env.GCS_UPLOAD_DIR != github.workspace) && !canceled() }}
+        if:  ${{ (env.GCS_UPLOAD_DIR != github.workspace) && !cancelled() }}
         shell: bash
         run: |
           set -x
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -253,20 +253,20 @@ jobs:
         with:
           ref: ${{needs.prepare_matrix.outputs.github_ref}}
           submodules: true
-      - name: Set env vars (ubuntu)
+      - name: Set env vars (Linux)
         if: startsWith(matrix.os, 'ubuntu')
         run: echo "VCPKG_TRIPLET=x64-linux" >> $GITHUB_ENV
-      - name: Set env vars (macos)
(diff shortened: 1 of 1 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -356,6 +356,38 @@ jobs:
...
+        # Copy the artifacts and also the upload scripts since python's relpath
+        # produces errors when called across two different drives. Use tar to
+        # move the files as it neatly deepcopies firestore's symlinks.
+        if:  ${{ (env.GCS_UPLOAD_DIR != github.workspace) && !cancelled() }}
+        shell: bash
+        run: |
+          set -x
+          mkdir -p ${{env.GCS_UPLOAD_DIR}}/scripts/gha/integration_testing
+          mkdir -p ${{env.GCS_UPLOAD_DIR}}/scripts/gha-encrypted
+          tar -chf ${{env.GCS_UPLOAD_DIR}}/ta.tar ta
```
vimanyu, line 383: Just for readability, could you put a comment here what the "ta" directory is? Others are easy to follow.

vimanyu (dismissed):
Thanks for taking care of this! One minor note but LGTM otherwise.

vimanyu:
Just a note, the boringssl change (download tarball -> checkout specific tag in repo) increased the boringssl source code directory size from 100M to 280M.

vimanyu (approved):

jonsimantov:
> Just a note, the boringssl change (download tarball -> checkout specific tag in repo) increased the boringssl source code directory size from 100M to 280M.

Would checking out with--depth 1 help this?

DellaBitta:
I'm going to try out some new git cleanup code that I've found to see if I can reduce the size of the checkout, but I'd like to merge this for now so that we can effectively use integration tests moving forward and I can roll back this change if it's not necessary for now.

PR #474, opened 2021-06-16
Fix "Undefined symbols" linker error when DocumentChange::npos was used

dconeybe:
Fix a bug where accessing `firebase::firestore::DocumentChange::npos` from a non-Android platform results in a linker error:

```
Undefined symbols for architecture x86_64:
  "firebase::firestore::DocumentChange::npos", referenced from:
      firebase_testapp_automated::FirebaseFirestoreBasicTest_TestDocumentChangeNpos_Test::TestBody() in integration_test.cc.o
ld: symbol(s) not found for architecture x86_64
```

Here is the declaration of `npos`:

https://github.com/firebase/firebase-cpp-sdk/blob/7ae9569eb9aa8628d7206bdf9553b2f639b02462/firestore/src/include/firebase/firestore/document_change.h#L66-L72

The fix is to add a *definition* of `npos` into the corresponding .cc file

### DIFF
--- a/firestore/src/common/document_change.cc
+++ b/firestore/src/common/document_change.cc
@@ -23,6 +23,8 @@ using Type = DocumentChange::Type;
 // Older NDK (r16b) fails to define this properly. Fix this when support for
 // the older NDK is removed.
 const std::size_t DocumentChange::npos = static_cast<std::size_t>(-1);
+#else
+constexpr std::size_t DocumentChange::npos;
 #endif  // defined(ANDROID)
 
 DocumentChange::DocumentChange() {}
--- a/release_build_files/readme.md
+++ b/release_build_files/readme.md
@@ -567,7 +567,7 @@ code.
 
 ## Release Notes
 
-### 8.0.1
+### 8.1.0
 -   Changes
     -   Firestore: Fixed a linker error when DocumentChange::npos was used.
         ([#474](https://github.com/firebase/firebase-cpp-sdk/pull/474)).
--- a/release_build_files/readme.md
+++ b/release_build_files/readme.md
@@ -569,7 +569,7 @@ code.
 
 ### 8.1.0
 -   Changes
-    -   Firestore: Fixed a linker error when DocumentChange::npos was used.
+    -   Firestore: Fixed a linker error when `DocumentChange::npos` was used.
         ([#474](https://github.com/firebase/firebase-cpp-sdk/pull/474)).
 
 ### 8.0.0
--- a/release_build_files/readme.md
+++ b/release_build_files/readme.md
@@ -567,6 +567,11 @@ code.
 
 ## Release Notes
 
+### 8.0.1
+-   Changes
+    -   Firestore: Fixed a linker error when DocumentChange::npos was used.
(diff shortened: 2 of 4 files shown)


firestore/src/common/document_change.cc
```diff
@@ -23,6 +23,8 @@ using Type = DocumentChange::Type;
 // Older NDK (r16b) fails to define this properly. Fix this when support for
 // the older NDK is removed.
 const std::size_t DocumentChange::npos = static_cast<std::size_t>(-1);
+#else
+constexpr std::size_t DocumentChange::npos;
```
dconeybe, line 27: @var-const Do I have to initialize `npos` here, as well as in the header file? I did some research about `constexpr` and couldn't find a definitive answer. The tests that I added work fine, but it's possible that they're just inlining the value of `npos` from the header file.

dconeybe: Nevermind, it's a compiler error to re-define it here. I thought I had tried that. Obviously not lol.

release_build_files/readme.md
```diff
@@ -567,6 +567,11 @@ code.
 
 ## Release Notes
 
+### 8.1.0
+-   Changes
+    -   Firestore: Fixed a linker error when DocumentChange::npos was used.
```
var-const, line 572: Nit: please wrap `DocumentChange::npos` in backticks.

release_build_files/readme.md
```diff
@@ -567,6 +567,11 @@ code.
...
+        ([#474](https://github.com/firebase/firebase-cpp-sdk/pull/474)).
```
var-const, line 573: Optional: I think the changelog usually links to issues, not PRs, so this link could probably be omited.

firestore/src/common/document_change.cc
var-const: If you're interested, this is something that changed in C++17 -- before C++17, defining a static `constexpr` variable like this was necessary, but starting from C++17 it's unnecessary and deprecated (see the last code snippet and the text around it [here](https://en.cppreference.com/w/cpp/language/static)). Like you said, this issue didn't come up previously because the variable wasn't used ("odr-used"). This [TotW](go/totw/140#a-static-class-member) is also relevant.

var-const (approved):
Thanks for fixing this! Please make sure that this is ok to merge/the version number in the changelog is up-to-date.

release_build_files/readme.md
dconeybe: Done.

release_build_files/readme.md
dconeybe: It looks like there is a mixture of issues and PRs linked to in previous change log entries. I see your point but I personally feel like adding a link to additional context is useful for future readers. If this issue had been discovered externally then there would be an issue number; however, since it was discovered by me, a developer on the Firestore team, no issue was created. IMO, providing this link adds value at no cost. If you feel strongly I can remove it, but I like having it here.

dconeybe (commented):
> Please make sure that this is ok to merge/the version number in the changelog is up-to-date.

I confirmed the version number at go/firebase-cpp-release.

var-const (approved):

PR #477, opened 2021-06-16
Format Firestore source code with clang-format

dconeybe:


firestore/src/android/document_reference_android.h
```diff
@@ -37,7 +37,8 @@ class DocumentReferenceInternal : public Wrapper {
 
   DocumentReferenceInternal(FirestoreInternal* firestore,
                             const jni::Object& object)
-      : Wrapper(firestore, object), promises_(firestore) {}
+      : Wrapper(firestore, object), promises_(firestore) {
+  }
```
wu-hui, line 41: I prefer the old way. 

Can you find out which config we can use to get the old way? I am OK to leave it if it takes you more than 10 minutes.

wu-hui (dismissed):

firestore/src/android/document_reference_android.h
dconeybe: Done. I just had to remove `AllowShortFunctionsOnASingleLine: None` from `firestore/.clang-format`.

https://github.com/firebase/firebase-cpp-sdk/blob/7ae9569eb9aa8628d7206bdf9553b2f639b02462/firestore/.clang-format#L9

wu-hui (approved):

PR #480, opened 2021-06-18
Added build source to user agent string.

alexames:
This allows the library to determine whether the library was built on
github or not.

### DIFF
--- a/app/src/app_common.cc
+++ b/app/src/app_common.cc
@@ -167,7 +167,7 @@ const char* kCpuArchitecture = "x86";
 
 #if FIREBASE_GITHUB_ACTION_BUILD
 const char* kBuildSource = "github_action_built";
-#elif
+#else
 const char* kBuildSource = "custom_built";
 #endif
 // clang-format=on
--- a/app/src/app_common.cc
+++ b/app/src/app_common.cc
@@ -164,6 +164,12 @@ const char* kCpuArchitecture = "x86";
 #else
 #error Unknown operating system.
 #endif  // Operating system
+
+#if FIREBASE_GITHUB_ACTION_BUILD
+const char* kBuildSource = "github_action_built";
+#elif
+const char* kBuildSource = "custom_built";
+#endif
 // clang-format=on
 
 const char* kApiClientHeader = "x-firebase-client";
@@ -303,6 +309,8 @@ App* AddApp(App* app, std::map<std::string, InitResult>* results) {
                          kCpuArchitecture);
     App::RegisterLibrary(FIREBASE_CPP_USER_AGENT_PREFIX "-stl",
                          kCppRuntimeOrStl);
+    App::RegisterLibrary(FIREBASE_CPP_USER_AGENT_PREFIX "-buildsrc",
+                         kBuildSource);
   }
   callback::Initialize();
   AppCallback::NotifyAllAppCreated(app, results);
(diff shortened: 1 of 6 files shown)


scripts/gha/build_desktop.py
```diff
@@ -182,6 +182,8 @@ def cmake_configure(build_dir, arch, msvc_runtime_library='static', linux_abi='l
   else:
     # workaround, absl doesn't build without tests enabled
     cmd.append('-DBUILD_TESTING=off')
+  # When building from GitHub Actions, this should always be set.
+  cmd.append('-DFIREBASE_GITHUB_ACTION_BUILD=ON')
```
DellaBitta, line 186: This shouldn't be hard coded here.  Our build scripts are run locally by the team and so local builds will be reported as GHA builds.  Instead create a runtime flag on build_desktop.py that appends this -D command, and then alter `.github/workflows/desltop.yml` and `.github/workflows/cpp-packaging.yml` to add a `--gha_build` flag (or something) when the workflow invokes `build_desktop.py`

app/src/app_common.cc
```diff
@@ -164,6 +164,12 @@ const char* kCpuArchitecture = "x86";
 #else
 #error Unknown operating system.
 #endif  // Operating system
+
+#if FIREBASE_GITHUB_ACTION_BUILD
```
DellaBitta, line 168: This should be an `#ifdef` instead.
 
`FIREBASE_GITHUB_ACTION_BUILD` isn't defined as `1` when  the cmake variable `FIREBASE_GITHUB_ACTION_BUILD` isn't defined, and this will cause a preprocessor error in non GHA builds.  That is to say, there's no scenario where `FIREBASE_GITHUB_ACTION_BUILD` defaults to `0`, but its actually undefined instead, and so `#if FIREBASE_GITHUB_ACTION_BUILD` will fail.

DellaBitta (changes requested):

app/src/app_common.cc
alexames: My understanding is that undefined symbols default to 0. I can add and `#else` statement to set it to 0 if you want to be explicit though. 

As a test, I just ran this, and it worked exactly as expected:
```
#include <iostream>
int main() {
#if UNDEFINED_SYMBOL
  std::cout << "Hello, world\n";
#else
  std::cout << "Goodbye, world\n";
#endif
  return 0;
}
```

```
g++ test.cpp -o test && ./test
```

```
Goodbye, world
```

alexames: Either way I now have it set to set the symbol to 1 or 0 instead of just setting it to 1.

scripts/gha/build_desktop.py
alexames: I think I made the changes you wanted. Take a look and make sure I set it up right.

app/src/app_common.cc
```diff
@@ -164,6 +164,12 @@ const char* kCpuArchitecture = "x86";
...
+const char* kBuildSource = "github_action_built";
+#elif
```
DellaBitta, line 170: This is causing a compilation error.  Should just be `#else`

DellaBitta (changes requested):

app/src/app_common.cc
alexames: Whoops. that's what I get for thinking I can get away with minor changes without compiling.🤦

scripts/gha/build_desktop.py
```diff
@@ -150,7 +150,7 @@ def install_cpp_dependencies_with_vcpkg(arch, msvc_runtime_library, cleanup=True
 
 def cmake_configure(build_dir, arch, msvc_runtime_library='static', linux_abi='legacy',
                     build_tests=True, config=None, target_format=None,
-                    use_openssl=False, disable_vcpkg=False, verbose=False):
+                    use_openssl=False, disable_vcpkg=False, gha_build=False,verbose=False):
```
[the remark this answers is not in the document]
alexames: Done.

DellaBitta (approved):

PR #483, opened 2021-06-21
Integration test remove "update podfile" logic

sunmou99:
As we have `update_ios_android_dependencies.py` script. It's save to remove the similar logic in integration tests


vimanyu (approved):
Thanks for cleaning this up! LGTM!

PR #485, opened 2021-06-22
boringSSL clone specific commit

DellaBitta:
Found a way to clone the boringSSL sources to a specific commit, instead of cloning the repo first and then checking out a commit.  

This frees up enough disk space that we no longer need to copy the resulting artifacts over to the C drive on Windows hosts in order to make enough room for the google cloud sdk installation.  Therefore I've removed that step from the CI saving up to 20 minutes per windows integration test build.


vimanyu (commented):
Possible to use the commit tag variable here to avoid duplication?

cmake/external/boringssl.cmake
```diff
@@ -21,15 +21,20 @@ endif()
...
   ${CMAKE_CURRENT_LIST_DIR}/../../scripts/git/patches/boringssl/0001-disable-C4255-converting-empty-params-to-void.patch)
 
+set(commit_tag, 83da28a68f32023fd3b95a8ae94991a07b1f6c62)
+
 ExternalProject_Add(
   boringssl
+  
+  DOWNLOAD_COMMAND 
+    COMMAND git init boringssl
+    COMMAND cd boringssl && git fetch --depth=1 https://github.com/google/boringssl 83da28a68f32023fd3b95a8ae94991a07b1f6c62 && git reset --hard FETCH_HEAD
```
vimanyu, line 31: Possible to use `commit_tag` variable here?

vimanyu (approved):
LGTM!

PR #486, opened 2021-06-22
Resolve macros related to STLPort.

ehsannas:
The following should be true going forward:
`FIREBASE_USE_STD_FUNCTION`: Always defined
`STLPORT`: Always undefined
`_STLPORT_VERSION`: Always undefined
So we can simplify the code on this basis.

More changes to come related to removal of STLPort:
* Removing `AddSnapshotListener` and `AddSnapshotsInSyncListener` (and `RunTransaction`?) variants that use `EventListener`
* Removing `EventListener`
* Removing STLPort-specific workarounds in the code on a case-by-case basis
* Move to C++11 standard definitions of move, unique_ptr, shared_ptr, etc. where applicable


firestore/src/common/compiler_info.cc
```diff
@@ -163,9 +163,7 @@ std::ostream& operator<<(std::ostream& os, LanguageVersion) {
 struct StandardLibraryVendor {};
 
 std::ostream& operator<<(std::ostream& os, StandardLibraryVendor) {
-#if defined(_STLPORT_VERSION)
```
var-const, line 166: I'm a little torn about removing this -- I guess it's possible (even if very unlikely) that someone could compile the SDK from the source code and modify it to work on STLPort. Also, unlike other cases, this isn't really a workaround, so it doesn't add to our tech debt, IMO. Unless you feel strongly about this, I'm leaning towards leaving this in.

var-const (changes requested):

firestore/src/android/firestore_android.h
```diff
@@ -97,10 +94,9 @@ class FirestoreInternal {
   // Runs transaction atomically.
   Future<void> RunTransaction(TransactionFunction* update,
                               bool is_lambda = false);
```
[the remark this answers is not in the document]
ehsannas: Done

firestore/src/android/document_reference_android.cc
```diff
@@ -173,8 +171,6 @@ ListenerRegistration DocumentReferenceInternal::AddSnapshotListener(
                              /*passing_listener_ownership=*/true);
```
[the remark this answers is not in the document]
ehsannas: Done

firestore/integration_test_internal/src/transaction_extra_test.cc
```diff
@@ -26,8 +26,6 @@ namespace firestore {
 
 using TransactionExtraTest = FirestoreIntegrationTest;
```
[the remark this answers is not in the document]
ehsannas: Filed b/191970743

ehsannas:
Also rebased against `main`

firestore/src/common/compiler_info.cc
var-const: It would be quite a bit of work, but not impossible. My main point is that if we're acknowledging that there is a possibility of the SDK running on platforms we're not aware of (outputting `unknown`), we might as well have a check for STLPort, just for completeness' sake. I don't see how it costs us anything in terms of maintenance.

var-const (dismissed):

ehsannas:
Rebased again to resolve merge conflicts

firestore/src/common/compiler_info.cc
var-const: Bump.

ehsannas: Done

var-const (approved):

PR #487, opened 2021-06-23
Feature/tvos cleanup integration tests

vimanyu:
Tons of cleanup on tvos integration test setup. Retaining only the bare minimum required to setup the XCode projects.

- Removed any user specific directories for XCode across the repo (xcuserdata). Added this to `.gitignore` too. 
- Removed `.xcworkspace` as it is created automatically on `pod install`
- Removed `Pods` sub project and linker reference to pod libraries in xcode project. These are automatically done on `pod install` too.
- Removed "Assets" and "plist" file for tvos target. Aside from getting rid of unncessary files, we also get rid of deeply nested directories under Assets folder which caused errors while checkout out the repo on Windows (260 character limit for paths)
- Sharing the same plist file for ios and tvos targets. This is more convenient for running simulation tests on CI as our existing infrastructure for iOS for post processing Info.plist files should just work for tvos too.
- Setting tvOS deployment target to 10.1 to include older devices in our tests.

To test,
Since this PR mostly involves xcode files, the best way to check would be to try out any of the integration tests locally.
- Get GoogleService-Info.plist
- `python3 setup_integration_tests.py`
- cd <firebase_product>/integration_test
- `pod install`
- `open integration_test.xcworkspace`
- Drag and drop xcframeworks. Prebuilt ones can be obtained from this [run](https://github.com/firebase/firebase-cpp-sdk/actions/runs/959812320)
- Build and run integration_test_tvos


sunmou99 (approved):
Tested Manually with iOS+tvOS XCFramework.
LGTM

PR #491, opened 2021-06-24
Fix absl leakage by using Firebase 8.2.0

wu-hui:


wu-hui:
> Added `Firestore::NamedQuery` that allows reading the queries used to build a Firestore Data Bundle.

Done.

firestore/src/include/firebase/firestore.h
```diff
@@ -429,6 +429,21 @@ class Firestore {
...
+   * Named queries are packaged into bundles on the server side (along with the
+   * resulting documents) and loaded into local cache using `LoadBundle`. Once
+   * in the local cache, you can use this method to extract a query by name.
+   *
+   * If a query cannot be found, the returned future will complete with its
+   * `error()` set to a non-zero error code.
+   *
+   * @param query_name The name of the query to read from saved bundles.
+   */
+  virtual Future<Query> NamedQuery(const std::string& query_name);
```
var-const, line 445: Sorry about missing this earlier -- for all other functions in the public API that take string parameters, we have both a `const std::string&` overload and a `const char*` overload. Can you please file a bug to add the `const char*` overload here? (I'd rather keep it out of this PR)

var-const (approved):

jonsimantov (approved):
@vimanyu Make sure this is included in the next release

PR #493, opened 2021-06-24
Address STLPort-specific workarounds

ehsannas:
This covers all mentions of b/163140650

### DIFF
--- a/firestore/src/common/field_value.cc
+++ b/firestore/src/common/field_value.cc
@@ -25,14 +25,6 @@ namespace {
 
 using Type = FieldValue::Type;
 
-template <typename T>
-std::string ValueToString(const T& value) {
-  // TODO(stlport): use `std::to_string` once possible.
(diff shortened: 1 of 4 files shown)


firestore/src/common/settings.cc
```diff
@@ -18,13 +18,7 @@ const char kDefaultHost[] = "firestore.googleapis.com";
 
 std::string ToStr(bool v) { return v ? "true" : "false"; }
 
-std::string ToStr(int64_t v) {
```
var-const, line 21: You can remove this overload entirely and just directly call `std::to_string` below.

ehsannas: Done

var-const (approved):

ehsannas:
Had to rebase and also we still can't use `std::to_string` :(

var-const (approved):

PR #494, opened 2021-06-25
Don't use EventListener in integration tests.

ehsannas:


ehsannas:
There's still 1 test failure (TransactionTest.TestCanHaveGetsWithoutMutations) only on one configuration(test-android-macos-latest-android_latest).

ehsannas:
I will roll back the original PR for now

PR #496, opened 2021-06-26
Rework how tests deal with future timeout

var-const:


firestore/integration_test_internal/src/firestore_integration_test.h
```diff
@@ -1,6 +1,7 @@
 #ifndef FIREBASE_FIRESTORE_CLIENT_CPP_SRC_TESTS_FIRESTORE_INTEGRATION_TEST_H_
 #define FIREBASE_FIRESTORE_CLIENT_CPP_SRC_TESTS_FIRESTORE_INTEGRATION_TEST_H_
 
+#include <chrono>  // NOLINT(build/c++11)
```
dconeybe, line 4: Do we still need these "NOLINT" comments? I thought they were only required in google3?

firestore/integration_test_internal/src/firestore_integration_test.h
```diff
@@ -43,11 +44,15 @@ bool ProcessEvents(int msec);
...
-// Waits for a Future to complete. If a timeout is reached then this method
-// returns as if successful; therefore, the caller should verify the status of
-// the given Future after this function returns. Returns the number of cycles
-// that were left before a timeout would have occurred.
-int WaitFor(const FutureBase& future);
+// Blocks until either the given predicate `pred` returns `true` or timeout
+// occurs. Returns `true` on success, `false` on timeout or if exit signal was
+// received.
+template <typename PredT>
+bool WaitUntil(const PredT& pred, int timeout_ms = kTimeOutMillis);
```
dconeybe, line 51: Consider returning an enum from `WaitUntil()` and `WaitUntilFutureCompletes()` instead of conflating timeout and exit signal with a `false` return value. This will improve readability at the call sites since it is not immediately obvious without reading the function's documentation or implementation what the return value means and will avoid misleading "timeout" error messages when the exit signal occurs.

firestore/integration_test_internal/src/firestore_integration_test.h
```diff
@@ -280,37 +285,27 @@ class FirestoreIntegrationTest : public testing::Test {
 
   // TODO(zxu): add a helper function to block on signal.
 
-  // A helper function to block until the future completes.
+  // Blocks until the future completes and returns the future result.
```
dconeybe, line 288: ultranit: Consider adding a comma after "completes".

firestore/integration_test_internal/src/firestore_integration_test.h
```diff
@@ -280,37 +285,27 @@ class FirestoreIntegrationTest : public testing::Test {
...
+    EXPECT_NE(future.result(), nullptr) << DescribeFailedFuture(future);
+
     return future.result();
   }
 
+  // Blocks until the future completes.
   static void Await(const Future<void>& future);
 
-  // A helper function to block until there is at least n event.
+  // Blocks until there is at least `event_count` events.
```
dconeybe, line 303: ultranit: "is" -> "are"

firestore/integration_test_internal/src/firestore_integration_test.h
```diff
@@ -355,6 +350,24 @@ class FirestoreIntegrationTest : public testing::Test {
   mutable std::unordered_map<Firestore*, FirestoreInfo> firestores_;
 };
 
+template <typename PredT>
+bool WaitUntil(const PredT& pred, int timeout_ms) {
+  auto now = std::chrono::steady_clock::now();
```
dconeybe, line 355: The logic in this method is a bit difficult to follow IMO because `now` changes value and meaning over time. Consider getting rid of the `now` variable and just calling `std::chrono::steady_clock::now()` when the current time is needed.

I realize that `now` is used to determine the final result on line 368, but this could be avoided if the `while` loop were modified to be `while (true)`, something like this:

```
while (true) {
  if (pred()) {
    return true;
  } else if (std::chrono::steady_clock::now() >= timeout_time) {
    return false;
  }
  ...
}
```

I think a change like this would improve the overall readability of the method.

dconeybe (changes requested):
Overall, this looks fantastic! It's great to see the timeout logic de-duplicated in several places. Thanks for doing this.

firestore/integration_test_internal/src/firestore_integration_test.h
var-const: I believe we do (we started doing them way back when everything was still in the iOS repo). I'm not sure if the C++ repo has the same linter checks enabled. We might want to revisit this at some point.

firestore/integration_test_internal/src/firestore_integration_test.h
var-const: Hmm, my understanding is that this is a compound predicate (`Blocks... and returns`) and thus doesn't require a comma. I can't dig up a very authoritative reference quickly but the few links that come up in a search seem to agree -- see e.g. [this](https://grammar.yourdictionary.com/sentences/what/what-is-a-predicate.html).

firestore/integration_test_internal/src/firestore_integration_test.h
var-const: I usually try to avoid `while (true)` loops where possible -- personally, it feels like the loop simply obscures the loop condition (effectively there still is a loop condition, it's just implicit in the loop body).

If you feel strongly about getting rid of `now`, the last line could instead be something like `return pred();`. What do you think?

Alternatively, perhaps renaming `now` could help? Maybe something like `last_checked_time`?

firestore/integration_test_internal/src/firestore_integration_test.h
var-const: Thanks!

firestore/integration_test_internal/src/firestore_integration_test.h
var-const: I thought about this but was leaning towards not doing it. It seems to complicate the common case for the sake of a very rare occurrence. I'm not sure this even works correctly -- e.g. on Linux the integration tests seem to completely ignore `Ctrl-C`.

var-const: > I'm not sure this even works correctly -- e.g. on Linux the integration tests seem to completely ignore Ctrl-C.

Please disregard -- this is actually working correctly. I'm still hesitant to complicate the common case, though.

var-const: Hmm, it seems like it actually works in some tests and doesn't work in others.

dconeybe: My concern still stands, that returning a `bool` detracts from readability at the call sites. And confounding exit signal with timeout could lead to misleading test failure messages. When tests timeout they are already hard enough to debug and unclear error messages can further make that harder. This doesn't block my approval of this PR though.

dconeybe (approved):

dconeybe (approved):

firestore/integration_test_internal/src/firestore_integration_test.h
var-const: Ah, I see what you mean, and I fully agree. How about `Blocks until the future is completed and returns...`? I think it could still be read the wrong way but it seems like an improvement. What do you think?

firestore/integration_test_internal/src/firestore_integration_test.h
var-const: It seems to me that the root cause here is that these functions just don't have a great way of reporting this event, which really seems more akin to an exception. There also isn't much tests can do about it, so I'm not sure they need the ability to check for this condition. I think that this type of exit usually happens due to user input, so it shouldn't be confusing to the user, at least in the general case.

dconeybe (dismissed):

firestore/integration_test_internal/src/firestore_integration_test.cc
```diff
@@ -280,18 +272,12 @@ FirestoreIntegrationTest::QuerySnapshotToMap(
   return result;
 }
 
-/* static */
-void FirestoreIntegrationTest::Await(const Future<void>& future) {
-  while (future.status() == FutureStatus::kFutureStatusPending) {
```
var-const, line 285: Previously, the overload of `Await` taking a `Future<void>` didn't have a timeout, unlike its `Future<T>` counterpart. Adding a timeout revealed that some of our transaction tests that do retries can take a varying amount of time, depending on the exact values of exponential backoff during that run, that just happens to be within a range of 11-18 seconds. The default timeout is 15 seconds, so that resulted in flaky tests. Adding a way to override the default timeout fixes the issue.

firestore/integration_test_internal/src/firestore_integration_test.h
dconeybe: Maybe we *should* throw an exception now that exceptions are enabled in tests, especially once #561 is merged. In any case, that's out of scope for this PR so I'm fine with this code as-is.

firestore/integration_test_internal/src/util/future_test_util.h
```diff
@@ -12,9 +12,11 @@
 namespace firebase {
 
 // Prints a human-friendly representation of a Future to an ostream.
-// This function is found dynamically by googletest to print a Future object
-// in a test failure message.
-void PrintTo(const Future<void>& future, std::ostream* os);
+// This function is via ADL by Googletest to print a Future object in a test
```
dconeybe, line 15: I think there is a missing verb in this sentence: "This function is via ADL...". Should it be "is *found*"?

dconeybe (approved):
Approved, with one very minor comment about a comment wording.

jonsimantov:
@dconeybe Is this PR still relevant? Should it be merged or closed?

dconeybe:
Yes, it is still relevant. I've merged in HEAD and assigned to myself.

PR #499, opened 2021-06-29
Add generated docs check and generated headers target, and remove FIREBASE_NAMESPACE macro.

jonsimantov:
Run Doxygen as part of the common PR checks.

This will fail if Doxygen returns any errors or warnings. This could include missing or misspelled parameters, missing method documentation, or other issues.

Because we don't perform a full header scrub here, the Doxygen config file uses _sed_ to remove everything between `<SWIG>` and `</SWIG>`. We could potentially perform a second pass with SWIG enabled, or do this in the Unity repo in the future, to check C# documentation.

This PR also fixes some of those same issues present in our build right now, including a few missing method docs in Firestore, a missing parameter in Database, and bad cross-references in a few different places, including all references to App and Future, which were failing due to the FIREBASE_NAMESPACE macro which is no longer needed. _This PR removes all usages of the FIREBASE_NAMESPACE macro._

Note that because we also run Doxygen on some generated headers, this PR includes a new CMake target to build those headers without building the entire build. Any future generated headers should be added to that target.

This PR also contains a small change to print the list of files requiring code formatting directly to the GitHub workflow log.


remote_config/src/include/firebase/remote_config.h
```diff
@@ -258,7 +258,7 @@ class RemoteConfig {
   ///
   /// @note This does not actually apply the data or make it accessible,
   /// it merely retrieves it and caches it.  To accept and access the newly
-  /// retrieved values, you must call @ref ActivateFetched().
+  /// retrieved values, you must call @ref Activate().
```
alexames, line 261: This should have been fixed in #481. Can you do a pull to make sure your branch is up to date?

alexames (approved):

remote_config/src/include/firebase/remote_config.h
alexames: Nevermind, looks like it auto-merged cleanly and closed the branch.

app/src/include/firebase/variant.h
```diff
@@ -28,17 +28,14 @@
...
-#endif
 
 namespace firebase {
 namespace internal {
 class VariantInternal;
 }
 }  // namespace firebase
 
-namespace FIREBASE_NAMESPACE {
+namespace firebase {
```
jonsimantov, line 38: This is a test comment. <hidden value="cpplint-file-comment"></hidden>

app/src/jobject_reference.h
```diff
@@ -20,38 +20,31 @@
...
+        : firebase::internal::JObjectReference(obj) {}                      \
+    classname(JNIEnv* env, jobject obj)                                     \
+        : firebase::internal::JObjectReference(env, obj) {}                 \
+    classname& operator=(const firebase::internal::JObjectReference& rhs) { \
+      firebase::internal::JObjectReference::operator=(rhs);                 \
+      return *this;                                                         \
+    }                                                                       \
+    classname& operator=(firebase::internal::JObjectReference&& rhs) {      \
+      firebase::internal::JObjectReference::operator=(rhs);                 \
+      return *this;                                                         \
```
jonsimantov, line 46: This is a second comment. <hidden value="cpplint-file-comment"></hidden>

PR #502, opened 2021-07-01
Update Auth ReauthenticateAndRetrieveData comments

DellaBitta:
The behavior of ReauthenticateAndRetrieveData errors differ per platform. The preexisting comments documented the iOS behavior only, which was confusing for users of other platforms. 

This PR clarifies that errors on Android and Desktop platforms may cause the user to be signed out.


vimanyu (approved):
LGTM!

PR #507, opened 2021-07-02
Use git apply patch to fix uWebsockets crash, previously fixed by python script in #344

DellaBitta:
Previously we downloaded a tarball release of uWebSockets and then patched the code to fix a crash issue. The patch was executed by a custom python script which somewhat blindly overwrote at set of line numbers in the file.  See [PR #344](https://github.com/firebase/firebase-cpp-sdk/pull/344)

Git apply patch wasn't used because it requires a initialized git repo to work.  Cloning the the uWebsockets repo with history and then checking out to an older commit would explode the disk usage, causing parition exhaustion issues on Windows, so we were forced to use the non-git-initialized tarball.

Thankfully I found a way to fetch a repo at a specific commit, removing the overhead of cloning all of the history.  This change updates the cmake download step for uWebSockets to fetch the commit instead of downloading a tarball.  With a git initialized repo we can our change with a far safer `git apply` patch instead of our custom python script, akin to what was done for boringSSL in [PR #485](https://github.com/firebase/firebase-cpp-sdk/pull/485)

Integration Tests have been flaky, but I was able to get a successful run on the latest commit [here](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1007959550)


cmake/external/boringssl.cmake
```diff
@@ -21,14 +21,14 @@ endif()
...
-set(commit_tag 83da28a68f32023fd3b95a8ae94991a07b1f6c62)
+set(boringssl_commit_tag 83da28a68f32023fd3b95a8ae94991a07b1f6c62)
 
 ExternalProject_Add(
   boringssl
   
   DOWNLOAD_COMMAND 
     COMMAND git init boringssl
-    COMMAND cd boringssl && git fetch --depth=1 https://github.com/google/boringssl ${commit_tag} && git reset --hard FETCH_HEAD
+    COMMAND cd boringssl && git fetch --depth=1 https://github.com/google/boringssl ${boringssl_commit_tag} && git reset --hard FETCH_HEAD
```
vimanyu, line 31: Should this be https//github.com/google/boringssl.git?

scripts/git/patches/uWebsockets/0001-fix-want-write-crash.patch
```diff
@@ -0,0 +1,57 @@
...
+             if (ssl) {
++              /* BEG Patched by firebase-cpp-sdk 00001-fix-want-write-crash.patch */
++              bool continue_loop = true;
++              do {
+                 sent = SSL_write(ssl, message->data, message->length);
+                 if (sent == (ssize_t) message->length) {
+-                    wasTransferred = false;
+-                    return true;
++                  wasTransferred = false;
++                  return true;
```
vimanyu, line 26: Is it possible to use the original indentation to have a cleaner diff?

vimanyu (commented):
Minor comment on cleaning up the diff (patch) file but looks good otherwise.

cmake/external/boringssl.cmake
DellaBitta: Seems like it works, so no?

scripts/git/patches/uWebsockets/0001-fix-want-write-crash.patch
DellaBitta: Done.

vimanyu (dismissed):
LGTM!

cmake/external/boringssl.cmake
```diff
@@ -21,14 +21,14 @@ endif()
 set(patch_file 
   ${CMAKE_CURRENT_LIST_DIR}/../../scripts/git/patches/boringssl/0001-disable-C4255-converting-empty-params-to-void.patch)
 
-set(commit_tag 83da28a68f32023fd3b95a8ae94991a07b1f6c62)
+set(boringssl_commit_tag 83da28a68f32023fd3b95a8ae94991a07b1f6c62)
```
jonsimantov, line 24: Just curious, why do these need to be changed?

jonsimantov (approved):

vimanyu (approved):

cmake/external/boringssl.cmake
DellaBitta: I had mistakenly typoed the commit tag in the applicaiton of it, and it was still set to the boring ssl commit tag from the previous cmake file, which had a nasty `silent` error that the commit couldn't be found (but the build kept on going).  Having an explicity variable for each checkout will make things easier to debug later, **I think**.

PR #510, opened 2021-07-05
Remove STLPort support from public API.

ehsannas:


ehsannas:
This PR is a combination of #486 (which I rolled back) and #494.

firestore/src/include/firebase/firestore.h
```diff
@@ -253,13 +249,9 @@ class Firestore {
    * The string reference parameter can be used to set the error message.
    *
    * @return A Future that will be resolved when the transaction finishes.
-   *
-   * @note This method is not available when using the STLPort C++ runtime
-   * library.
    */
   virtual Future<void> RunTransaction(
       std::function<Error(Transaction&, std::string&)> update);
```
var-const, line 254: I cannot remember why we didn't `ifdef` out the other version of `RunTransaction` (the one taking a `TransactionFunction`), but I strongly suspect it was an oversight. I think that `TransactionFunction` (defined in `transaction.h`) and everything related to it can be removed now as well, similarly to `EventListener`.

However, given that we forgot to make the `RunTransaction` that takes a `TransactionFunction` conditionally-defined, we need to mark it deprecated first. Can you please create a follow-up PR to do just that? (using the `FIREBASE_DEPRECATED` macro)

var-const (approved):

firestore/src/include/firebase/firestore.h
ehsannas: +1  yep, will do.

ehsannas: Posted https://github.com/firebase/firebase-cpp-sdk/pull/512

var-const: Thanks!

PR #511, opened 2021-07-06
Make C++ validation tests run on all platforms

var-const:
Also:
* updated the GoogleTest version (to be able to use the `Throws` matcher);
* enabled exceptions in Firestore integration tests (otherwise validation tests would be ignored);
* add some missing validation (see the linked bug for remaining gaps);
* replace all usage of `FIREBASE_ASSERT` macros with `SIMPLE_HARD_ASSERT` (which throws an exception on Unity / crash in C++).

See b/136012313 for more context.


firestore/integration_test_internal/src/validation_test.cc
```diff
@@ -19,27 +19,276 @@
...
+      return "FirebaseFirestore has already been started and its settings can "
+             "no longer be changed. You can only call setFirestoreSettings() "
+             "before calling any other methods on a FirebaseFirestore object.";
+#else
+      return "Firestore instance has already been started and its settings can "
+             "no longer be changed. You can only set settings before calling "
+             "any other methods on a Firestore instance.";
+#endif
+
+    case ErrorCase::kSettingsDisableSsl:
```
dconeybe, line 90: The two values for `kSettingsDisableSsl` are identical. Can you get rid of the `#ifdef` for this one?

firestore/integration_test_internal/src/validation_test.cc
```diff
@@ -120,49 +364,24 @@ class ValidationTest : public FirestoreIntegrationTest {
...
-          AnyOf(
-              // When validated by iOS or common C++ code
-              StrEq(reason),
-              // When validated by Android Java code
-              StrEq("Use FieldPath.of() for field names containing '~*/[]'.")));
-    }
+    EXPECT_ERROR_EITHER(
+        document.Update(MapFieldValue{{path, FieldValue::Integer(1)}}), reason,
+        // TODO(b/171990785): Unify Android and C++ validation error messages.
+        // When validated by Android Java code
```
dconeybe, line 380: This comment `When validated by Android Java code` doesn't make sense to me. Is it a copy-paste error? If not, could you re-write it as a full English sentence?

firestore/integration_test_internal/src/validation_test.cc
```diff
@@ -350,8 +515,12 @@ TEST_F(ValidationTest, WritesMayContainIndirectlyNestedLists) {
 #endif  // defined(FIREBASE_USE_STD_FUNCTION)
 }
 
-// TODO(zxu): There is no way to create Firestore with different project id yet.
-TEST_F(ValidationTest, WritesMustNotContainReferencesToADifferentDatabase) {}
+// TODO(b/136012313): batches and transactions don't currently throw.
+// TEST_F(ValidationTest, WritesMustNotContainReferencesToADifferentDatabase) {
```
dconeybe, line 519: Consider uncommenting this test, but skipping it via the `GTEST_SKIP()` macro. This will at least highlight in the test result count that a test has been skipped. Leaving it commented here makes it less visible.

There is precedence for this too, like all of the macros in https://github.com/firebase/firebase-cpp-sdk/blob/8ad99ec49ff5d01e909d5cedc5366cf5325460ec/testing/test_framework/src/firebase_test_framework.h.

firestore/src/main/transaction_main.cc
```diff
@@ -156,5 +163,14 @@ void TransactionInternal::MarkPermanentlyFailed() {
   transaction_->MarkPermanentlyFailed();
 }
 
+void TransactionInternal::ValidateReference(const DocumentReference& document) {
+  auto* internal_doc = GetInternal(&document);
```
dconeybe, line 167: Isn't it possible that the "internal" DocumentReference is nullptr? If so, you should probably check if it's null before using it.

dconeybe (changes requested):
I think you should add a change log entry, since there are externally-visible behavior changes being made by this PR.

firestore/integration_test_internal/src/validation_test.cc
var-const: Thanks for spotting this! I think I originally thought that the C++ validation message would spell it `ssl_enabled`, but it turned out to be not true.

firestore/integration_test_internal/src/validation_test.cc
var-const: Done. I think the original idea was something like "The message below is produced when validation is performed by the Android SDK", hopefully it reads a little better now.

firestore/integration_test_internal/src/validation_test.cc
var-const: Thanks, great suggestion.

firestore/src/main/transaction_main.cc
var-const: Good catch. The existing code would crash in that case. I'll add a validation test to cover this tomorrow.

dconeybe: Just waiting for this to be addressed. Then I'll approve.

var-const:
Some changes since the last review:
* added validation and tests for the case where an invalid `DocumentReference` is passed to a `Transaction` or a `WriteBatch` (previously it would crash);
* made `WritesMustNotContainReferencesToADifferentDatabase` work by adding a new test helper function, `TestFirestoreWithProjectId` (this necessitated a slight refactoring in `FirestoreIntegrationTest`);
* updated the GoogleTest version to be able to use the `Throws` matcher;
* enabled exceptions in Firestore integration tests;
* added a `README` entry.

firestore/src/main/transaction_main.cc
var-const: Added the test (also addressed the same problem with `WriteBatch`).

firestore/src/common/document_reference.cc
```diff
@@ -113,12 +114,23 @@ CollectionReference DocumentReference::Parent() const {
 
 CollectionReference DocumentReference::Collection(
     const char* collection_path) const {
+  if (!collection_path) {
+    SimpleThrowInvalidArgument("Collection path cannot be null.");
+  }
+  if (collection_path[0] == '\0') {
```
dconeybe, line 120: Is there a reason to not use `strlen(collection_path) == 0` here, instead of `collection_path[0] == '\0'`?

dconeybe (approved):

firestore/integration_test_internal/CMakeLists.txt
```diff
@@ -378,5 +378,9 @@ add_subdirectory(${FIREBASE_CPP_SDK_DIR} bin/ EXCLUDE_FROM_ALL)
 # Note that firebase_app needs to be last in the list.
 set(firebase_libs firebase_firestore firebase_auth firebase_app)
 set(gtest_libs gtest gmock)
+# Validation tests need exceptions to be enabled in order to run. The only
+# library that explicitly disables exceptions is App (but it gets propagated
+# everywhere), so overriding that effectively reenables exceptions.
+target_compile_options(firebase_app PUBLIC -fexceptions)
```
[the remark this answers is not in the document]
dconeybe: FYI This change probably won't be required once https://github.com/firebase/firebase-cpp-sdk/pull/561 is merged. Can you see if you can undo this change if you merge after it?

jonsimantov:
Has the latest _main_ been merged into this PR, to ensure that the GoogleTest update doesn't break the flaky-test-workaround code in test_framework?

firestore/integration_test_internal/CMakeLists.txt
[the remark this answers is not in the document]
var-const: Thanks for the pointer! I'll give it a try once that PR is merged.

firestore/integration_test_internal/src/validation_test.cc
```diff
@@ -172,44 +381,30 @@ TEST_F(ValidationTest, ChangingSettingsAfterUseFails) {
...
+TEST_F(ValidationTest, FirestoreGetInstanceWithNullAppFails) {
+  EXPECT_ERROR(
+      Firestore::GetInstance(/*app=*/nullptr, /*init_result=*/nullptr),
+      "firebase::App instance cannot be null. Use "
+      "firebase::App::GetInstance() without arguments if you'd like to use "
+      "the default instance.");
+}
 
 TEST_F(ValidationTest,
        FirestoreGetInstanceWithNonNullAppReturnsNonNullInstance) {
```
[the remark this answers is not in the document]
var-const: Added a null pointer check as well.

firestore/integration_test_internal/src/validation_test.cc
```diff
@@ -381,40 +545,107 @@ TEST_F(ValidationTest, WritesMustNotContainReservedFieldNames) {
...
+  auto* db2 = TestFirestore("db2");
+  EXPECT_NE(db1, db2);
+
+  std::string reason =
+      "Provided document reference is from a different Cloud Firestore "
+      "instance.";
+  MapFieldValue data{{"foo", FieldValue::Integer(1)}};
+  DocumentReference bad_ref = db2->Document("foo/bar");
+
+#if defined(__ANDROID__)
```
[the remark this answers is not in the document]
var-const: Added a TODO.

firestore/src/common/document_reference.cc
var-const: I don't have a strong preference on this -- it just seemed a little simpler to do.

var-const:
@jonsimantov My branch had some of those, but not all of them (e.g. I had some `installations` tests with flaky sections, but not `storage` tests). Just to be sure, I merged the latest `main` and will run all tests again.

firestore/src/common/document_reference.cc
dconeybe: Personally, I'd prefer `strlen` because IMO it is more readable; that is, it directly conveys what you are checking. Checking that the first element == 0 is indirectly checking the string's length. This won't block my approval though.

dconeybe (approved):

jonsimantov (approved):
LGTM for the googletest changes. Deferring to Denver's review for the Firestore-specific test changes.

PR #512, opened 2021-07-06
Deprecate RunTransaction which takes a TransactionFunction.

ehsannas:


firestore/src/include/firebase/firestore.h
```diff
@@ -264,8 +264,16 @@ class Firestore {
    * transaction is valid as long as the Future has not yet completed.)
    *
    * @return A Future that will be resolved when the transaction finishes.
+   *
+   * @note This method is only available when using the STLPort C++ runtime
```
var-const, line 268: Please remove the `@note` part. It would be true if this method was `#ifdef`ed, but we didn't do it, and adding the `#ifdef` now would be essentially equivalent to removing it.

firestore/src/include/firebase/firestore.h
```diff
@@ -264,8 +264,16 @@ class Firestore {
...
+   * library.
+   *
+   * @deprecated STLPort support in Firestore is deprecated and will be removed
```
var-const, line 271: I'd rephrase the comment a little bit, particularly to make it clear that the STLPort support was already removed. Consider something like:
```
@deprecated This method was originally added to support the STLPort C++ runtime on Android.
Firebase support for STLPort has been removed, and consequently this method is
deprecated and will be removed in a future release.
```

var-const (changes requested):

firestore/src/include/firebase/firestore.h
ehsannas: Done

firestore/src/include/firebase/firestore.h
ehsannas: Done

var-const (dismissed):

ehsannas:
We need to update our own test that uses this API as well:
```
/Users/runner/work/firebase-cpp-sdk/firebase-cpp-sdk/ta/firestore/iti/src/transaction_test.cc:119:42: error: 'RunTransaction' is deprecated [-Werror,-Wdeprecated-declarations]
  Future<void> future = TestFirestore()->RunTransaction(&transaction);
```

ehsannas:
All tests pass now.

firestore/integration_test_internal/src/transaction_test.cc
```diff
@@ -91,38 +91,26 @@ class TransactionTest : public FirestoreIntegrationTest {
...
-  std::string key() { return key_; }
-  std::string value() { return value_; }
-
- private:
-  DocumentReference doc_;
-  const std::string key_{"foo"};
-  const std::string value_{"bar"};
-};
-
 TEST_F(TransactionTest, TestGetNonexistentDocumentThenCreatePortableVersion) {
```
var-const, line 94: The `PortableVersion` part can now be removed from the name of the test.

var-const (changes requested):
Almost LGTM, just a couple of minor things:
* `TransactionFunction` is also used in `includes_test.cc`;
* file comments referring to `TransactionFunction` in `transaction_test.cc` and `transaction_extra_test.cc` (`We will be using lambda...`) can now be removed.

firestore/integration_test_internal/src/transaction_test.cc
ehsannas: Done

firestore/integration_test_internal/src/transaction_test.cc
```diff
@@ -91,38 +91,26 @@ class TransactionTest : public FirestoreIntegrationTest {
...
   DocumentReference doc = TestFirestore()->Collection("towns").Document();
-  TestTransactionFunction transaction{doc};
-  Future<void> future = TestFirestore()->RunTransaction(&transaction);
+  std::string key = "foo";
+  std::string value = "bar";
+  Future<void> future = TestFirestore()->RunTransaction(
+      [this, doc, key, value](Transaction& transaction,
```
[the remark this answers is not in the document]
ehsannas: Done

ehsannas:
The test failure is `FirebaseMessagingTest.TestSendMessageToToken` 😭 
restarting the tests

var-const (approved):

ehsannas:
in today's episode of github actions failures we have:
* SyntaxError: Non-ASCII character '\xe2' in file scripts/gha/it_workflow.py on line 60
* FirebaseAuthTest.TestAnonymousSignin test failure
* FirebaseDatabaseTest.TestInfoConnected test failure

this is super frustrating ... 😭

ehsannas:
This time around, I'm getting these failures:
```
FirebaseMessagingTest.TestSubscribeAndUnsubscribe
BundleTest.LoadInvalidBundlesShouldFail
```

At least this time one of the test failures is in Firestore (even though it's probably a flake as it didn't fail in the previous two runs)...

Let's try again for the 6th time.

PR #517, opened 2021-07-07
Temporarily disable tvOS integration test against packaged SDK

sunmou99:
Temporarily disable tvOS integration test against packaged SDK.
Running Packaging workflow now: https://github.com/firebase/firebase-cpp-sdk/actions/runs/1009798586
Integration triggered without running tvOS tests: https://github.com/firebase/firebase-cpp-sdk/actions/runs/1010009783


jonsimantov:
Is this something that will be re-enabled in the future? We'll definitely want to test tvOS against packaged SDK if we can.

sunmou99:
> Is this something that will be re-enabled in the future? We'll definitely want to test tvOS against packaged SDK if we can.

Yes, we will re-enable it after this ticket: https://b.corp.google.com/issues/193068220
Currently, we test tvOS against source.

vimanyu (approved):
LGTM

PR #524, opened 2021-07-12
Fix flaky test

jonsimantov:
Add retry logic to FirebaseMessagingTest.TestSendMessageToToken.

### DIFF
--- a/testing/test_framework/src/firebase_test_framework.h
+++ b/testing/test_framework/src/firebase_test_framework.h
@@ -183,20 +183,20 @@ namespace firebase_test_framework {
 // Helper macros to assist with RunFlakyBlock.
 // Unlike EXPECT_*, FLAKY_EXPECT_* just prints out the error and returns
 // false, which will cause RunFlakyBlock to retry.
-#define FLAKY_EXPECT_EQ(a, b)                                 \
-  {                                                           \
-    auto a_result = (a);                                      \
-    auto b_result = (b);                                      \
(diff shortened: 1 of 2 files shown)


testing/test_framework/src/firebase_test_framework.h
```diff
@@ -215,6 +215,27 @@ namespace firebase_test_framework {
...
+      std::stringstream a_str;                                         \
+      a_str << a_result;                                               \
+      app_framework::LogError(                                         \
+          "Expected %s to be null, but it is not null. value(%s)", #a, \
+          a_str.str().c_str());                                        \
+      return false;                                                    \
+    }                                                                  \
+  }
+#define FLAKY_EXPECT_NOTNULL(a)                                              \
+  {                                                                          \
```
DellaBitta, line 231: technically this doesn't need a scope, but it looks nicer this way. :)

DellaBitta (dismissed):

testing/test_framework/src/firebase_test_framework.h
jonsimantov: Without the scope, the trailing ; will confuse the compiler - it ignores a ; after a {}.

(Although a lot of people like defining macros as "do { /* code_goes_here */ } while(0)" to give the semicolon something to do.)

DellaBitta: Ah right right.

DellaBitta (approved):

PR #527, opened 2021-07-12
Add retry for brew update in integration test.

jonsimantov:


sunmou99 (changes requested):
Homebrew are used at 4 different places. 
This is another place:
https://github.com/firebase/firebase-cpp-sdk/blob/a39bd30dd346bef6772803c53f6a8a1cffc87dbb/.github/workflows/integration_tests.yml#L400

jonsimantov:
> Homebrew are used at 4 different places.
> This is another place:
> https://github.com/firebase/firebase-cpp-sdk/blob/a39bd30dd346bef6772803c53f6a8a1cffc87dbb/.github/workflows/integration_tests.yml#L400

Thanks, will update all of them. Wanted to see if this would work first - will also add to the other workflows.

sunmou99 (approved):
LGTM

PR #529, opened 2021-07-13
Update FTL test devices

sunmou99:
Updated mobile devices setting. Removed deprecated FTL devices.
```
gcloud firebase test ios models list
gcloud firebase test android models list
```


scripts/gha/print_matrix_configuration.py
```diff
@@ -151,8 +151,8 @@
   "emulator_target": {"type": "virtual", "image":"system-images;android-28;google_apis;x86_64"},
   "emulator_latest": {"type": "virtual", "image":"system-images;android-29;default;x86"},
   "ios_min": {"type": "real", "model":"iphone8", "version":"11.4"},
-  "ios_target": {"type": "real", "model":"iphone6s", "version":"12.0"},
-  "ios_latest": {"type": "real", "model":"iphone11pro", "version":"14.1"},
+  "ios_target": {"type": "real", "model":"iphone8plus", "version":"12.0"},
+  "ios_latest": {"type": "real", "model":"iphone11", "version":"13.6"},
```
jonsimantov, line 155: Just a note that we should remember to add back an ios14 device once there is a non-deprecated one.

jonsimantov (approved):

scripts/gha/print_matrix_configuration.py
sunmou99: This requires manual maintenance. I feel like we need a plan to check the device list. Like 1) quarterly, 2) or during Tech Debt Fixit, 3) or whenever we find we are using a deprecated device

PR #533, opened 2021-07-13
GitHub REST API: add timeout and retry 

sunmou99:
Existing http request flakiness in step "Update PR label and comment": https://github.com/firebase/firebase-cpp-sdk/runs/2945372323?check_suite_focus=true#step:18:23
Should be fixed now.

Add timeout (5s) to all GitHub REST API requests.
Add retry (3 times) to all query and update requests.

### DIFF
--- a/scripts/gha/github.py
+++ b/scripts/gha/github.py
@@ -23,6 +23,13 @@ import json
 import shutil
 
 from absl import logging
+from requests.adapters import HTTPAdapter
+from requests.packages.urllib3.util.retry import Retry
+
+RETRIES = 3
+BACKOFF = 3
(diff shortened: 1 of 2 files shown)


sunmou99:
add comment test

sunmou99:
add comment test

vimanyu (commented):
One clarification, do we not want to use retries on "put" and "patch" requests?

sunmou99:
> One clarification, do we not want to use retries on "put" and "patch" requests?

We retries on "patch".
I don't retry on "post", because retry may add duplicate comments.

vimanyu (approved):
LGTM!

PR #534, opened 2021-07-13
Clean up flaky tests a bit by adding std::function support to RunWithRetry/RunFlakyBlock.

jonsimantov:
By adding std::function overloads to the flaky retry functions (possible now that stlport is removed), we can clean up a lot of code that uses RunFlakyBlock and RunWithRetry.

Refactor flaky test handling in Storage to use RunWithRetry instead of RunFlakyBlock where possible - it's much cleaner.

Also add some flaky test handling in Auth and Analytics.


auth/integration_test/src/integration_test.cc
```diff
@@ -719,62 +719,75 @@ TEST_F(FirebaseAuthTest, TestWithCustomEmailAndPassword) {
...
+        FLAKY_EXPECT_TRUE(loaded_provider_data_ids == prev_provider_data_ids);
+
+        // Cleanup, ensure we are signed in as the user so we can delete it.
+        FLAKY_WAIT_FOR_COMPLETION(
+            auth_->SignInWithEmailAndPassword(email.c_str(), kTestPassword),
+            "SignInWithEmailAndPassword");
+        FLAKY_EXPECT_NONNULL(auth_->current_user());
+        DeleteUser();
+
+        return true;
```
alexames, line 786: Consider replacing this with something like:

    return FLAKEY_SUCCESS;

or something similar.

auth/integration_test/src/integration_test.cc
```diff
@@ -860,57 +873,67 @@ TEST_F(FirebaseAuthTest, TestPhoneAuth) {
...
-      EXPECT_NE(listener.verification_id(), "");
-      LogDebug("Signing in with verification code.");
-      const firebase::auth::Credential phone_credential =
-          phone_provider.GetCredential(listener.verification_id().c_str(),
-                                       kPhoneAuthTestVerificationCode);
-
-      WaitForCompletion(auth_->SignInWithCredential(phone_credential),
-                        "SignInWithCredential(PhoneCredential)");
-    }
+  if (!RunFlakyBlock([&]() {
```
alexames, line 876: If you wanted to be really fancy, I bet you could create a FLAKEY_TEST_F macro, that was defined as something like:

    enum class FlakeResult { FLAKEY_FAILURE, FLAKEY_SUCCESS };

    #define FLAKEY_TEST_F(TestHarness, TestName)                      \
                                                                      \
    class TestHarness ## _ ## TestName ## _ ## FlakeHarness {         \
      FlakeResult RunTest();                                          \
    }                                                                 \
                                                                      \
    TEST_F(TestHarness, TestName) {                                   \
      if (FLAKEY_FAILURE == RunFlakyBlock([&]() {                     \
        TestHarness ## _ ## TestName ## _ ## FlakeHarness().RunTest() \
      })) {                                                           \
        FAIL() << "Your flakey test failed.";                         \
      }                                                               \
    }                                                                 \
                                                                      \
    FlakeResult TestHarness ## _ ## TestName ## _ ## FlakeHarness::RunTest()

(This is just a proof of concept, and I haven't tested or even tried to compile it - but I think something of this general form would work)

Then you could just write the test like this:

    FLAKEY_TEST_F(FirebaseAuthTest, TestPhoneAuth) {
      // Do the test
      return FLAKEY_SUCCESS;
    }

If there is work you only want to do once, you could throw it in the Setup and Teardown functions, and It hink that would work for you.

alexames (changes requested):
Left two suggestions, feel free to file bugs and defer either one

auth/integration_test/src/integration_test.cc
jonsimantov: This is a very interesting idea - I'm going to attempt this in a subsequent PR.

auth/integration_test/src/integration_test.cc
jonsimantov: I just made it FLAKY_SUCCESS() to match the syntax of the others.

auth/integration_test/src/integration_test.cc
alexames: sgtm

alexames (dismissed):
lgtm

alexames (approved):

PR #536, opened 2021-07-14
Feature/update versions improvements

vimanyu:
Improvements to script that updates Android and iOS versions across the repo.

- Renamed to `update_android_ios_versions.py` for consistency with g3 script names.
- Added `--ignore_android_packages` to specify a list of substrings that will be matched against android package names. Android package names with these substrings will be skipped for updates.
- Added `--ignore_ios_pods` to specify a list of substrings that will be matched against pod names. iOS pods with these substrings will be skipped for updates.
- Added `--ignore_directories` to specify list of directory names to not recurse into while searching for files to update.
- Added `--allow_experimental` to allow versions containing "-alpha", "-beta" or any other trailing qualifiers. ("-cppsdk" is always accepted as a special case)
- Fixed trailing space at the end of an android package substitution in readme files.
- Fixed parsing of android packages of type foo.bar.version-<something>. Eg: foo.bar.1.0.0-alpha

For the time being, the ideal command for release is
```
python3 scripts/update_android_ios_versions.py --ignore_ios_pods Google-Mobile-Ads-SDK --ignore_android_packages firebase-ads --ignore_directories gameloop_android
```


jonsimantov:
What about a setting to avoid using alpha versions of Android packages?

scripts/update_android_ios_dependencies.py
```diff
@@ -71,7 +79,7 @@
 
 
 def get_files_from_directory(dirpath, file_extension, file_name=None,
-                             absolute_paths=True):
+                             absolute_paths=True, ignore_directories=None):
```
sunmou99, line 82: What about `ignore_directories=[]` ?

vimanyu:
> What about a setting to avoid using alpha versions of Android packages?

Am allowing only versions numbers with digits (and versions with -cppsdk as a special case) in the updates and have added a flag `--allow_experimental` to allow all versions (eg: 1.2.3-alpha, 12.3-beta etc) if the user wants to just use the latest and greatest. This affects iOS and Android both.

scripts/update_android_ios_dependencies.py
vimanyu: Ah... Using mutable values as defaults for arguments is a known python gotcha and can lead to unexpected behavior.
https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

This pattern (default to None and set it to empty list in function) is used to avoid this.

scripts/update_android_ios_dependencies.py
```diff
@@ -34,18 +34,30 @@
...
-python3 scripts/update_ios_android_dependencies.py --podfiles foo/Podfile
+python3 scripts/update_android_ios_dependencies.py --podfiles foo/Podfile
                                                               dir_with_podfiles
 
+# Update all Android packages except any names containing androidx and auth
+python3 scripts/update_android_ios_dependencies.py --skip_ios
+  --ignore_android_packages auth androidx
+
+# Ignore updating any files in gameloop_android and test directories
+python3 scripts/update_android_ios_dependencies.py --ignore_directories test
```
sunmou99, line 58: I feel it's safe to ignore the  /scripts folder

vimanyu: Yup. Makes sense. I want to avoid hard coding as much as possible. This docstring is an example of the flags that users can use to ignore specific directories.

scripts/update_android_ios_dependencies.py
sunmou99: Interesting😔

scripts/update_android_ios_dependencies.py
```diff
@@ -516,7 +596,7 @@ def replace_module_line(m):
 # Regex to match lines like:
 # implementation 'com.google.firebase:firebase-auth:1.2.3'
```
sunmou99, line 597: It will be good to update the example here.

sunmou99 (approved):
The script LGTM

A few other things: 
1) I feel it's good to provide a default value to`--ignore_directories`, `--ignore_android_packages`, and `--ignore_ios_pods`, even if the default value is None.  Because other use may not have the knowledge.
2) The name `update_android_ios_dependencies` may be confusing, because tvos is also affected

vimanyu:
> The script LGTM
> 
> A few other things:
> 
> 1. I feel it's good to provide a default value to`--ignore_directories`, `--ignore_android_packages`, and `--ignore_ios_pods`, even if the default value is None.  Because other use may not have the knowledge.
> 2. The name `update_android_ios_dependencies` may be confusing, because tvos is also affected

Great points.
1. Do you mean to make these required flags? I feel that these flags are not going to be used for the long term (only until we have pinned admob). As far as directories are concerned, I think the issue with gameloop_android updating was it updated to the latest (alpha) version of androidx core and maybe that wasn't compatible. Ideally, if we choose to update androidx.core across the repository, it might be beneficial to check if the new version also works for gameloop_android. 

2. I agree that it is not completely descriptive but the name will get unnecessarily long. And this script updates the cocoapods which contain both ios and tvos libraries. I think we will keep the name as this for now as it is the same as another similar script in g3.

sunmou99:
> > The script LGTM
> > A few other things:
> > 
> > 1. I feel it's good to provide a default value to`--ignore_directories`, `--ignore_android_packages`, and `--ignore_ios_pods`, even if the default value is None.  Because other use may not have the knowledge.
> > 2. The name `update_android_ios_dependencies` may be confusing, because tvos is also affected
> 
> Great points.
> 
> 1. Do you mean to make these required flags? I feel that these flags are not going to be used for the long term (only until we have pinned admob). As far as directories are concerned, I think the issue with gameloop_android updating was it updated to the latest (alpha) version of androidx core and maybe that wasn't compatible. Ideally, if we choose to update androidx.core across the repository, it might be beneficial to check if the new version also works for gameloop_android.
> 2. I agree that it is not completely descriptive but the name will get unnecessarily long. And this script updates the cocoapods which contain both ios and tvos libraries. I think we will keep the name as this for now as it is the same as another similar script in g3.

1. I mean provide default value like this:
https://github.com/firebase/firebase-cpp-sdk/blob/main/scripts/gha/print_matrix_configuration.py#L361
But as long as whoever using this script knows what to exclude, I'm good.

2. I'm fine with the name.

jonsimantov (commented):
If you don't want to add default values for --ignore_android_packages and --ignore_ios_packages, then what do you think about including a shell script to run the script with the correct parameters for our releases?

Otherwise, make sure this is documented in both our workflow doc and the releasing doc.

vimanyu:
> If you don't want to add default values for --ignore_android_packages and --ignore_ios_packages, then what do you think about including a shell script to run the script with the correct parameters for our releases?
> 
> Otherwise, make sure this is documented in both our workflow doc and the releasing doc.

Yes, I think it makes sense to keep this command as simple as possible for the release. I have added default values with a TODO note to remove them when we do not need them anymore. Updating the release and workflow docs.

scripts/update_android_ios_dependencies.py
vimanyu: Done!

sunmou99 (approved):

PR #538, opened 2021-07-14
Add a "wrapper" header that includes `firestore_exceptions.h` from the iOS repo

var-const:
This simplifies the Unity build (the SWIG part of the build has issues with directly including a header that comes from the iOS repo).


var-const:
@cynthiajoan I can merge this PR if it helps with the Unity build.

cynthiajoan (approved):

PR #539, opened 2021-07-14
Fix detection of which tests to run when "quick" tests are requested.

jonsimantov:
Previously, the code was attempting to auto-diff with ${{github.event.pull_request.base.sha}}.

As it turns out, this is incorrect - the base sha in this case points to the current head of main, not the head of main when the branch was created (which is closer to what we want).

Instead, we now do the actual correct thing: query git to find the __common ancestor commit__ of _origin/main_ and _origin/[pr branch name]_, and diff the PR's branch (again from origin) against that common ancestor.

This makes it much more robust to detect which files were actually edited by a PR, even if the base main branch has since been modified (and even if those modifications were partially merged into this branch).

Also in this PR: change to print_matrix_configuration to allow us to treat certain files as being within a given subdirectory when looking through the file list in auto-diff mode. For example, cmake/external/firestore.cmake should be considered under the "firestore" directory.


scripts/gha/print_matrix_configuration.py
```diff
@@ -243,16 +243,32 @@ def filter_values_on_diff(parm_key, value, auto_diff):
       # Any top-level directories set to None are completely ignored.
       "external": None,
       "release_build_files": None,
+      # Uncomment the two below lines when debugging this script, or GitHub
+      # actions related to auto-diff mode.
+      # ".github": None,
+      # "scripts": None,
```
sunmou99, line 249: Will these code be removed?

scripts/gha/print_matrix_configuration.py
```diff
@@ -243,16 +243,32 @@ def filter_values_on_diff(parm_key, value, auto_diff):
...
       # For example, if auth is touched by a PR, we also need to test functions,
       # database, firestore, and storage.
       "auth": "auth,functions,database,firestore,storage",
     }
+    file_redirects = {
+      # Custom handling for specific files, to be treated as a different path or
+      # ignored completely (set to None).
+      "cmake/external/firestore.cmake": "firestore",
+      "cmake/external/libuv.cmake": "database",
+      "cmake/external/uWebSockets.cmake": "database",
```
sunmou99, line 260: Do we have plan for expanding this redirect list?

scripts/gha/print_matrix_configuration.py
```diff
@@ -243,16 +243,32 @@ def filter_values_on_diff(parm_key, value, auto_diff):
...
+    }
     requested_api_list = set(value.split(','))
     filtered_api_list = set()
 
     for path in file_list:
       if len(path) == 0: continue
+      if path in file_redirects:
+        if file_redirects[path] is None:
```
sunmou99, line 268: It seems "if path in file_redirects", then file_redirects[path] cannot be None.

.github/workflows/integration_tests.yml
```diff
@@ -166,11 +166,18 @@ jobs:
...
+            git diff --name-only HEAD..${MERGE_BASE}
+            echo "AUTO_DIFF_PARAM=--auto_diff HEAD..${MERGE_BASE}" >> $GITHUB_ENV
           else
             # If running in a PR, diff against the PR's base.
-            echo "AUTO_DIFF_PARAM=--auto_diff ${{github.event.pull_request.base.sha}}" >> $GITHUB_ENV
+            # "git merge-base main branch_name" will give the common ancestor of both branches.
+            MERGE_BASE=$(git merge-base origin/${{github.event.pull_request.head.ref}} origin/${{github.event.pull_request.base.ref}})
+            echo "::warning ::Auto-diff origin/${{github.event.pull_request.head.ref}}..${MERGE_BASE}"
+            git diff --name-only origin/${{github.event.pull_request.head.ref}}..${MERGE_BASE}
+            echo "AUTO_DIFF_PARAM=--auto_diff origin/${{github.event.pull_request.head.ref}}..${MERGE_BASE}" >> $GITHUB_ENV
```
sunmou99, line 180: I maybe wrong.
But isn't origin/${{github.event.pull_request.head.ref}} equal to HEAD, and ${MERGE_BASE} is equal to origin/${{github.event.pull_request.base.ref}?

scripts/gha/print_matrix_configuration.py
jonsimantov: Can't it though? e.g.:
file_redirects = {
  "ignore_this_file.txt": None
}

scripts/gha/print_matrix_configuration.py
jonsimantov: *shrug* do you want me to?

scripts/gha/print_matrix_configuration.py
jonsimantov: Eventually.

.github/workflows/integration_tests.yml
jonsimantov: No, see, this is actually what the issue was.

HEAD is actually a merge commit that's created just for the purpose of the workflow run. For example, HEAD might point to https://github.com/firebase/firebase-cpp-sdk/commit/b069bd7d8edb80d0b3d9137120ab486973f7d463, whereas the latest commit is actually one prior to that - https://github.com/firebase/firebase-cpp-sdk/commit/494ee219290440e3ced7556df989186ffc0a4388.

Meanwhile, origin/main points to the latest head of main in the repo, not the main when the PR was created.

The whole point is that neither of these points to the ref that you'd expect.

scripts/gha/print_matrix_configuration.py
sunmou99: I maybe wrong, but then it is not redirects and shouldn't put here.

.github/workflows/integration_tests.yml
jonsimantov: Here's an example, let me know if this makes sense.

- main: contains commits 111, 222, and 333.
- git checkout main # you are at commit 333
- git checkout -b my_feature_branch
- Then you make changes and commit new commits AAA, BBB, and CCC.
- The commit history for my_feature_branch is now (in reverse order) CCC, BBB, AAA, 333, 222, 111.
- Someone else checks in more commits into main: 444, 555, 666
- Now main is 666, 555, 444, 333, 222, 111 and your branch is CCC, BBB, AAA, 333, 222, 111.
- In order to auto-diff, you need to diff your branch against the common ancestor (in this case 333). This will show the changes that you made in your commits AAA, BBB, and CCC only.
- Now, let's say you merge main into your branch. Your branch now looks like 666, 555, 444, CCC, BBB, AAA, 333, 222, 111.
- But because of the way that merge commits work, the common ancestor is now 666.
- If someone else checks in commits 777 and 888 into main at this point, your common ancestor with main is still 666.
- This will still correctly determine the diff from your branch, even if you merge commits from main into your branch as you go, because diffing your branch against 666 still correctly shows the changes in AAA, BBB, and CCC only.

I realize that was probably entirely unclear - sorry!

sunmou99 (approved):
LGTM. 
Thanks for the explanation.

jonsimantov:
> ### ⏳  Integration test in progress...
> Requested by @jonsimantov on commit [1b31f93](https://github.com/firebase/firebase-cpp-sdk/commit/1b31f93c164633b56181b484569cb0f00c1a2c69)
> Last updated: Wed Jul 14 20:47 PDT 2021
> **[View integration test log & download artifacts](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1032577800)**

Merging anyway, as it's a flake.

Failure log:
BundleTest.LoadBundlesForASecondTimeSkips (1079 ms)
[ RUN      ] BundleTest.LoadInvalidBundlesShouldFail
DEBUG: Using Firestore Prod for testing.
/Users/runner/work/firebase-cpp-sdk/firebase-cpp-sdk/ta/firestore/iti/src/bundle_test.cc:230: Failure
Expected equality of these values:
  progresses.size()
    Which is: 0
  1

PR #540, opened 2021-07-14
Enable expanded matrix for integration test

sunmou99:
Add expanded matrix for integration test workflow.
Nightly Test Report starts to use expanded matrix to build & test.


.github/workflows/integration_tests.yml
```diff
@@ -158,7 +158,7 @@ jobs:
           fetch-depth: 0
           submodules: false
       - name: Use expanded matrix
-        if: github.event.inputs.use_expanded_matrix == '1' || needs.check_trigger.outputs.requested_tests == 'full'
+        if: github.event_name == 'schedule' || github.event.inputs.use_expanded_matrix == '1' || needs.check_trigger.outputs.requested_tests == 'full'
         run: |
           echo "EXPANDED_MATRIX_PARAM=-e=1" >> $GITHUB_ENV
```
jonsimantov, line 163: Could you actually log a ::warning here (similar to what we do for auto-diff mode) so we can be sure it's running on the expanded matrix?

scripts/gha/print_matrix_configuration.py
```diff
@@ -113,6 +113,17 @@
       "ndk_version": ["r22b"],
       "platform_version": ["28"],
       "build_tools_version": ["28.0.3"],
+
+      EXPANDED_KEY: {
+        # architecture_windows_linux, ndk_version, msvc_runtime: not enabled yet
+        "architecture_windows_linux": ["x64", "x86"],
```
jonsimantov, line 119: Since I think we're not currently testing on x86, if this fails, feel free to remove it until we can fix it.

.github/workflows/integration_tests.yml
sunmou99: log added
https://github.com/firebase/firebase-cpp-sdk/runs/3071396257?check_suite_focus=true#step:3:15

jonsimantov (approved):

PR #548, opened 2021-07-16
Disable test on android min devices

sunmou99:


jonsimantov (approved):

jonsimantov:
With regards to the minimum device, we probably should find a later device to use as a minimum: the error we got is mainly "Unable to initialize due to missing Google Play services dependency"

sunmou99:
> "Unable to initialize due to missing Google Play services dependency"

Sure. Created ticket for it. b/193938198

PR #550, opened 2021-07-16
Add flaky retry logic for waiting for token in Messaging.

jonsimantov:
Sometimes it can take a while to get the token, and the first attempt fails.


jonsimantov:
Ignoring the failure as it's due to an unrelated issue - that there is no tvOS messaging test, so there's nothing to run and the test scripts can't handle that.

vimanyu (approved):
LGTM!

PR #551, opened 2021-07-16
After a branch is deleted, force quick tests to run the full suite.

jonsimantov:
Fix for bug in https://github.com/firebase/firebase-cpp-sdk/runs/3089608121


sunmou99 (approved):
LGTM

PR #553, opened 2021-07-16
Integration test workflow exclude tvOS build & test for certain apis

sunmou99:
This is a quick fix that integration test workflow exclude tvOS build & test for certain apis (e.g. admob, analytics, messaging, etc.)
"admob,analytics" not trigger tvos: https://github.com/firebase/firebase-cpp-sdk/actions/runs/1046072350
"analytics,auth" trigger tvos: https://github.com/firebase/firebase-cpp-sdk/actions/runs/1046073504

Note that the `build_testapps` script also need to handle this type of exclusion. But I will fix it in a separate PR.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -186,8 +186,9 @@ jobs:
       - id: export-result
         # e.g. 'ubuntu-latest,macos-latest' -> '["ubuntu-latest","macos-latest"]'
         run: |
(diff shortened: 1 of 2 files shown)


vimanyu (dismissed):
LGTM! 

Couldn't think of a better name for the new flag (-p) but at some point, we should make our flags consistent across all scripts. -p is used for platforms in build_testapps.py. 
Maybe use "-i" for apis?

vimanyu (approved):

PR #554, opened 2021-07-19
Remove unnecessary guard around FieldValue::Increment.

ehsannas:
Firestore has been building with INTERNAL_EXPERIMENTAL=1 for quite some time. These methods should always be exposed. Test coverage for them is available in field_value_test.cc and numeric_transforms_test.cc.

### DIFF
--- a/firestore/CMakeLists.txt
+++ b/firestore/CMakeLists.txt
@@ -342,6 +342,8 @@ add_custom_target(
   COMMENT "Copying internal Firestore headers"
 )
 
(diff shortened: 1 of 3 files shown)


var-const (changes requested):
I think it's worth calling out in the release notes, something like
```
`FieldValue::Increment` functions are no longer guarded by the `INTERNAL_EXPERIMENTAL` macro.
```

var-const (approved):

PR #555, opened 2021-07-19
Bugfix/Android emulator test failures

sunmou99:
1. Use Android Emulators with Google play Service
2. Accept all licenses
3. Set virtual device test timeout: 60 minutes


scripts/gha/print_matrix_configuration.py
```diff
@@ -156,7 +156,7 @@
   "android_latest": {"type": "real", "model":"flame", "version":"29"},
   "emulator_min": {"type": "virtual", "image":"system-images;android-18;google_apis;x86"},
   "emulator_target": {"type": "virtual", "image":"system-images;android-28;google_apis;x86_64"},
-  "emulator_latest": {"type": "virtual", "image":"system-images;android-29;google_apis;x86_64"},
+  "emulator_latest": {"type": "virtual", "image":"system-images;android-29;google_apis;x86"},
```
jonsimantov, line 159: Why 32-bit?

sunmou99: firestore has errors on latest x86_64 devices. They are look on it: b/192351260

jonsimantov: It looks like Firestore is fixing these errors (or maybe has fixed them already). Assuming that change goes in, we should keep this to 64-bit.

HOWEVER, I actually think it might make sense to check on at least one 32-bit device as part of the expanded matrix. Maybe we add an emulator_32bit?

scripts/gha/test_simulator.py
```diff
@@ -493,14 +493,21 @@ def _setup_android(platform_version, build_tool_version, sdk_id):
...
   args = ["sdkmanager", 
     "emulator", "platform-tools", 
     "platforms;%s" % platform_version, 
     "build-tools;%s" % build_tool_version]
   logging.info("Install packages: %s", " ".join(args))
   subprocess.run(args=args, check=True)
 
+  args = ["sdkmanager", "--licenses"]
+  logging.info("Accept all licenses: %s", " ".join(args))
+  p_yes = subprocess.Popen(["echo", "yes"], stdout=subprocess.PIPE)
```
jonsimantov, line 506: Wow is this really the right way to do this?

Shouldn't it be more like ... create a file handle, pass it into the below subprocess.Popen, and then write "yes" to it in Python?

jonsimantov (changes requested):

scripts/gha/test_simulator.py
sunmou99: I'm no expert. This is a headache for me. It definitely works here: https://github.com/firebase/firebase-cpp-sdk/blob/9fce79ac551c04c352c8d511ef340c13928d1697/scripts/gha/test_simulator.py#L531

What's your solution? Could you provide an example?

jonsimantov: I'm not really an expert either - let's save this for a future cleanup. :)

jonsimantov (approved):

PR #556, opened 2021-07-20
Feature/tvos messaging support

vimanyu:
- Added integration_test_tvos target to XCode project for messaging
- Using the newer recommended way of requesting permissions for messaging (ios10, tvos10) Instead of entirely disabling requests for tvos.
- Added messaging to the config file used by integration tests github workflows.

### DIFF
--- a/messaging/integration_test/Podfile
+++ b/messaging/integration_test/Podfile
@@ -1,9 +1,14 @@
 
 source 'https://github.com/CocoaPods/Specs.git'
-platform :ios, '10.0'
 # Firebase Cloud Messaging test application.
 
 target 'integration_test' do
+  platform :ios, '10.0'
+  pod 'Firebase/Messaging', '8.3.0'
+end
+
+target 'integration_test_tvos' do
+  platform :tvos, '10.0'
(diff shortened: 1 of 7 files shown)


sunmou99 (commented):
The changes looks good to me. And I tried it locally. Tests passed.
But as you mentioned during today's stand up: UI tests were excluded in the workflow, and UI tests have weird errors on Simulators (I saw errors on my Simulator too). It's worth mentioning what's the UI test result on your real device.

alexames (approved):

vimanyu:
> The changes looks good to me. And I tried it locally. Tests passed.
> But as you mentioned during today's stand up: UI tests were excluded in the workflow, and UI tests have weird errors on Simulators (I saw errors on my Simulator too). It's worth mentioning what's the UI test result on your real device.

Thanks for testing :). 
I see a permissions dialog appear on Apple TV but the tests fail after that step. My guess is that I need to setup something else. Will continue to test and possibly incorporate the new way of setting up notifications as pointed out by Jon today in the future. Going to merge this in for now so I can finally toggle tvos in our cpp packaging workflow.

PR #560, opened 2021-07-21
Add `is_valid` methods to the public API classes that can be in an invalid state

var-const:
Also improve the error message contained in the invalid futures returned from invalid objects.

Addresses b/172570071 and b/157227167.

### DIFF
--- a/firestore/src/common/futures.h
+++ b/firestore/src/common/futures.h
@@ -56,8 +56,8 @@ Future<T> FailedFuture() {
   static auto* future = new Future<T>(FailedFuture<T>(
       Error::kErrorFailedPrecondition,
       "The object that issued this future is in an invalid state. This can be "
-      "because:\n- the object was default-constructed and never reassigned;\n"
-      "- the object was moved from;\n- the Firestore instance with which the "
+      "because the object was default-constructed and never reassigned, "
+      "the object was moved from, or the Firestore instance with which the "
       "object was associated has been destroyed."));
   return *future;
 }
--- a/firestore/src/common/futures.h
+++ b/firestore/src/common/futures.h
@@ -55,8 +55,10 @@ template <typename T>
 Future<T> FailedFuture() {
   static auto* future = new Future<T>(FailedFuture<T>(
       Error::kErrorFailedPrecondition,
-      "This instance is in an invalid state. This is because the underlying "
-      "Firestore instance has been destructed."));
+      "The object that issued this future is in an invalid state. This can be "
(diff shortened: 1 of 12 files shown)


var-const:
@dconeybe @schmidt-sebastian Feel free to tweak/nitpick the error message and the comments.

dconeybe (dismissed):

firestore/src/common/futures.h
```diff
@@ -55,8 +55,10 @@ template <typename T>
 Future<T> FailedFuture() {
   static auto* future = new Future<T>(FailedFuture<T>(
       Error::kErrorFailedPrecondition,
-      "This instance is in an invalid state. This is because the underlying "
-      "Firestore instance has been destructed."));
+      "The object that issued this future is in an invalid state. This can be "
+      "because:\n- the object was default-constructed and never reassigned;\n"
+      "- the object was moved from;\n- the Firestore instance with which the "
+      "object was associated has been destroyed."));
```
schmidt-sebastian, line 61: Is it common to have error messages like this in C++? I know that in Java a line break in an error message would be pretty annoying and make the overall message (which includes the stacktrace) less readable.

firestore/src/include/firebase/firestore/document_change.h
```diff
@@ -159,6 +159,19 @@ class DocumentChange {
    */
   virtual std::size_t new_index() const;
 
+  /**
+   * @brief Returns true if this `DocumentChange` is valid, false if it is
```
schmidt-sebastian, line 163: The last part (false if it is not valid) does not seem to be strictly needed.

firestore/src/include/firebase/firestore/query.h
```diff
@@ -637,6 +637,18 @@ class Query {
       std::function<void(const QuerySnapshot&, Error, const std::string&)>
           callback);
 
+  /**
+   * @brief Returns true if this `Query` is valid, false if it is not valid. An
```
schmidt-sebastian, line 641: I am not sure if we need to mention this, but the main usage for this API will likely be via `CollectionReference.is_valid`. I am not sure if all of our users are aware of our type hierarchy.

schmidt-sebastian (approved):
LGTM with nits you can ignore.

dconeybe (approved):

firestore/src/include/firebase/firestore/query.h
schmidt-sebastian: Yeah, I also do not know of a good way to address this.

schmidt-sebastian (approved):

PR #561, opened 2021-07-21
Bugfix/app cmake exceptions private

vimanyu:
Make the "no exceptions" compiler setting for Firebase app a private dependency so it does not affect the consuming projects. 
Some of the unit tests dealing with threads now throw an exception and have been modified to expect such exceptions.


app/tests/thread_test.cc
```diff
@@ -20,6 +20,17 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
+#define EXPECT_SYSTEM_ERROR(statement, error_code) \
```
jonsimantov, line 23: WDYT about naming this something like EXPECT_THROW_WITH_CODE? (just a thought - so it's clearer in the tests that this is specifically testing for a thrown exception.) Or if you think it isn't too wordy, just to use EXPECT_THROW directly with the EXPECT_EQ inside the catch in each of the tests, rather than the #define, since it's all just in this file anyway.

vimanyu: Ah good point. I wanted to avoid repeating the pattern (EXPECT_THROW with EXPECT_EQ)  in multiple places. I could change it to that but it seemed like a lot of code duplication. How about `EXPECT_THROW_WITH_ERROR_CODE`?

DellaBitta (dismissed):

release_build_files/readme.md
```diff
@@ -575,13 +575,13 @@ code.
         iOS and tvOS. The following products are currently included for tvOS:
         Auth, Database, Firestore, Functions, Installations, Messaging, 
         Remote Config, Storage.
+    -   General: Compiler setting of "no exceptions" on app is PRIVATE now
```
jonsimantov, line 578: maybe add a "(open-source build only)" annotation?

jonsimantov (approved):

PR #582, opened 2021-07-28
Add check for release notes update.

jonsimantov:
All PRs merging to `main` must either modify `release_build_files/readme.md` or contain the label `skip-release-notes`.


DellaBitta (approved):
Should only we admins be able to set this label?  Maybe it doesn't matter.

jonsimantov:
> Should only we admins be able to set this label? Maybe it doesn't matter.

Meh.

PR #585, opened 2021-07-29
Fix ReferenceCountedFutureImpl to avoid using FirestoreInternal after its deletion.

dconeybe:
This fixes a crash on Android that occurs if the `Firestore` instance is deleted from an "OnCompletion" handler attached to a `Future` returned from `RunTransaction()`. The issue is that in the bowels on the code that manages `Future` objects it will end up using a deleted `FirestoreInternal` object when the completion completes. The fix is to null out the `FirestoreInternal` pointer when it is deleted and avoid using it thereafter.

This fix also fixes a crash that was occurring in `SourceTests`. Those tests are re-enabled by this PR.

### DIFF
--- a/firestore/src/android/firestore_android.h
+++ b/firestore/src/android/firestore_android.h
@@ -7,6 +7,7 @@
 #include <functional>
 #include <list>
 #include <memory>
+#include <mutex>
 #include <unordered_set>
 
 #include "app/src/cleanup_notifier.h"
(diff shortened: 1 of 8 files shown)


firestore/src/android/firestore_android.h
```diff
@@ -191,6 +192,85 @@ class FirestoreInternal {
...
+  auto Run(CallbackT callback) -> decltype(callback(this->firestore_)) {
+    std::lock_guard<std::recursive_mutex> lock(mutex_);
+    return callback(firestore_);
+  }
+
+  /**
+   * Same as `Run()` except that it only invokes the given callback if the
+   * `FirestoreInternal` pointer is non-null.
+   */
+  void RunIfValid(std::function<void(FirestoreInternal*)> callback) {
```
var-const, line 266: Nit: because this code is not storing the callback, it can be passed by a const reference.

var-const: I'd suggest passing `FirestoreInternal` to the callback by reference, to show that it is guaranteed to be valid (not null) if the callback is invoked.

firestore/src/android/firestore_android.h
```diff
@@ -191,6 +192,85 @@ class FirestoreInternal {
...
+   * Invokes the given function with the referent `FirestoreInternal` object,
+   * specifying `nullptr` if the `FirestoreInternal` object has been deleted,
+   * returning whatever the given callback returns.
+   *
+   * If a non-null `FirestoreInternal` pointer was specified to the method, then
+   * that `FirestoreInternal` object will remain alive (i.e. will not be
+   * deleted) for the duration of the callback's execution.
+   *
+   * If this method is invoked concurrently then calls will be serialized in
+   * an undefined order. Recursive calls of this method are allowed and will not
```
var-const, line 253: Nit: consider `s/are allowed/are allowed on the same thread/`.

var-const (approved):

firestore/src/android/firestore_android.h
dconeybe: Done.

firestore/src/android/firestore_android.h
dconeybe: Done.

var-const (approved):

PR #586, opened 2021-07-29
Add firestore/CPPLINT.cfg

dconeybe:
The new `CPPLINT.cfg` just disables some noisy checks:

* `runtime/references` - This category produces warnings like `Is this a non-const reference? If so, make const or use a pointer: TestData& test_data`
* `build/c++11` - This category produces warnings like `<mutex> is an unapproved C++11 header.`


var-const (approved):
Thanks!

var-const (approved):

PR #592, opened 2021-08-05
Add Firestore to README.md.

ehsannas:


README.md
```diff
@@ -65,9 +66,9 @@ required packages:
 The following prerequisites are required when building the libraries for
 desktop platforms.
 
-- [OpenSSL](https://www.openssl.org/), needed for Realtime Database
+- [OpenSSL](https://www.openssl.org/), needed for Realtime Database and Cloud Firestore
 - [Protobuf](https://github.com/protocolbuffers/protobuf/blob/master/src/README.md),
-  needed for Remote Config
+  needed for Remote Config and Cloud Firestore
```
var-const, line 71: Hmm, I'm not sure if we need `Protobuf`, can you clarify?

ehsannas: I had to run `brew install protobuf` (and `sudo apt-get install protobuf-compiler` on Linux) when I first wanted to compile the firebase-cpp-sdk repo (although, now that I think about it, I built all the targets... perhaps if I had only built the `firebase_firestore` target alone it might have not been needed?)

var-const: We do use `Protobuf` in the iOS repo (in integration tests), maybe that's the reason?

It would be best to temporarily uninstall `protobuf` (or rename the executable, etc.) and try again, can you please try that?

ehsannas: I tried this, and the compilation was successful without it. So I'll update the readme

var-const (approved):
LGTM -- can you please also get an approval from somebody from the Games team before merging?

ehsannas:
@a-maurice Would you please review this PR as well? Thanks

a-maurice (approved):

PR #602, opened 2021-08-12
operator== and operator!= for DocumentSnapshot.

ehsannas:


firestore/src/android/document_snapshot_android.cc
```diff
@@ -114,5 +114,11 @@ FieldValue DocumentSnapshotInternal::Get(const FieldPath& field,
   return FieldValueInternal::Create(env, field_value);
 }
 
+bool operator==(const DocumentSnapshotInternal& lhs,
+                const DocumentSnapshotInternal& rhs) {
+  Env env = FirestoreInternal::GetEnv();
```
var-const, line 119: I presume both this pattern and the pattern in `common` will be repeated a lot. What do you think about creating a helper template function?

```cpp
// Defined elsewhere
template <typename T>
bool CompareJni(const T& lhs, const T& rhs) {
  Env env = FirestoreInternal::GetEnv();
  return Object::Equals(env, lhs.ToJava(), rhs.ToJava());
}

// ...
bool operator==(const DocumentSnapshotInternal& lhs,
                const DocumentSnapshotInternal& rhs) {
  return CompareJni(lhs, rhs);
}
```

And similarly for the the pattern in `document_snapshot.cc`.

ehsannas: Done

firestore/src/jni/compare.h
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+
+/**
+ * Returns the result of `Object::Equals` for the given `lhs` and `rhs` in the
+ * current Firestore `jni::Env`.
+ */
+template <typename T>
+bool CompareJni(const T& lhs, const T& rhs) {
+  Env env = FirestoreInternal::GetEnv();
```
var-const, line 23: Nit: add the appropriate includes.

var-const (changes requested):

firestore/src/jni/compare.h
dconeybe: Drive-by comment: You should just create a new `Env` object and use it here. The `Env` object is tied to the thread that created it and by using the `Env` returned from `FirestoreInternal::GetEnv()` it makes this comparison cause undefined behavior if ever used on a different thread (e.g. the user callback thread).

dconeybe: Alternately, you could require the caller to specify an `Env&` reference as an argument, with is semi-conventional within our code base.

ehsannas: Done.

firestore/src/jni/compare.h
```diff
@@ -0,0 +1,31 @@
...
+#include "firestore/src/android/wrapper.h"
+#include "firestore/src/jni/env.h"
+#include "firestore/src/jni/object.h"
+
+/**
+ * Returns the result of `Object::Equals` for the given `lhs` and `rhs` in the
+ * current Firestore `jni::Env`.
+ */
+template <typename T>
+bool CompareJni(const T& lhs, const T& rhs) {
```
var-const, line 27: Ultranit: I think this function should be named `EqualityCompareJni` for consistency with the other one. Sorry I missed that in my suggestions.

var-const (approved):

firestore/src/jni/compare.h
ehsannas: Done. It also turns out that it would have been better to place this function in the proper namespace (the Android build was not successful as it was looking for `firebase::firestore::*` for Env, Object, and GlobalUnhandledExceptionHandler.)

firestore/src/jni/compare.h
```diff
@@ -0,0 +1,39 @@
...
+namespace firebase {
+namespace firestore {
+namespace jni {
+
+/**
+ * Returns the result of `Object::Equals` for the given `lhs` and `rhs` in the
+ * current Firestore `jni::Env`.
+ */
+template <typename T>
+bool EqualityCompare(const T& lhs, const T& rhs) {
```
var-const, line 31: Nit: I'd name this function differently from the one in `common` (e.g. `EqualityCompareJni`). Having two functions that only differ by one namespace seems unnecessarily brittle.

var-const (approved):
LGTM with one comment.

firestore/src/jni/compare.h
ehsannas: Done

var-const (approved):

PR #605, opened 2021-08-16
setup ccache

wu-hui:


dconeybe:
Is this going to cause the cached ccache artifacts to grow unbounded over time?

wu-hui:
> Is this going to cause the cached ccache artifacts to grow unbounded over time?

My understanding is no. Each job has 5GB, and github will delete some cache to enforce the limit.

I do not intend to merge this as-is though, more of an experiment.

dconeybe:
> My understanding is no. Each job has 5GB, and github will delete some cache to enforce the limit.
> 
> I do not intend to merge this as-is though, more of an experiment.

Ahh got it. Thanks for the clarification!

PR #606, opened 2021-08-16
operator== and operator!= for SetOptions.

ehsannas:


firestore/src/main/set_options_main.h
```diff
@@ -27,7 +27,10 @@ class SetOptionsInternal {
       : options_{std::move(options)} {}
 
   Type type() const { return options_.type_; }
-  const std::vector<FieldPath>& field_mask() const { return options_.fields_; }
+  const std::vector<FieldPath> field_mask() const {
```
var-const, line 30: Can you make this function return a reference to the `unordered_set` instead, i.e. the exact stored type? It would require some changes to user data converter, but should be quite feasible. Let me know if you run into any trouble.

var-const (changes requested):

firestore/src/main/set_options_main.h
ehsannas: Done

var-const (dismissed):

ehsannas:
@var-const I had to resolve the conflict in readme.md, and so GitHub has thrown away your approval :(

var-const (approved):

PR #607, opened 2021-08-17
operator== and operator!= for DocumentChange

ehsannas:


ehsannas:
This has to be rebased once #602 has landed.

firestore/integration_test_internal/src/document_change_test.cc
```diff
@@ -78,5 +78,67 @@ TEST_F(DocumentChangeTest, TestDocumentChanges) {
...
+  Await(listener);
+  QuerySnapshot snapshot = listener.last_result();
+  EXPECT_EQ(snapshot.size(), 0);
+
+  WriteDocument(doc1, MapFieldValue{{"a", FieldValue::Integer(1)}});
+  Await(listener);
+  snapshot = listener.last_result();
+
+  std::vector<DocumentChange> changes = snapshot.DocumentChanges();
+  EXPECT_EQ(changes.size(), 1);
```
dconeybe, line 105: nit: Use `ASSERT_EQ` instead of `EXPECT_EQ` here. By using `EXPECT_EQ`, if the assertion fails then the test will continue. And if the assertion failed because the actual size of the vector was zero then accessing element 0 on line 107 will crash. By using `ASSERT_EQ` the test will be aborted and avoid the undefined behavior of accessing a non-existent element of a vector.

Also on lines 118 and 132.

firestore/integration_test_internal/src/document_change_test.cc
```diff
@@ -78,5 +78,67 @@ TEST_F(DocumentChangeTest, TestDocumentChanges) {
...
+  // first_change: Added doc1.
+  auto first_change = changes[0];
```
dconeybe, line 107: Optional: Consider using the names `change1` and `change2` instead of `first_change` and `second_change`, respectively. IMO it's a bit more readable that way and consistent with other naming conventions like `doc1` and `doc2`.

dconeybe (commented):
A couple of drive-by comments. Enjoy :)

var-const:
@ehsannas Sorry I'm slow with these reviews, will definitely get to them tomorrow.

firestore/src/common/document_change.cc
```diff
@@ -104,5 +104,13 @@ std::size_t DocumentChange::new_index() const {
   return internal_->new_index();
 }
 
+bool operator==(const DocumentChange& lhs, const DocumentChange& rhs) {
+  if (!lhs.internal_ || !rhs.internal_) {
```
var-const, line 108: Can you merge #602 first and use the helper functions defined in that PR here?

firestore/integration_test_internal/src/document_change_test.cc
var-const: +1.

firestore/integration_test_internal/src/document_change_test.cc
var-const: +1.

var-const (changes requested):

var-const (commented):
@ehsannas Also, I think readme needs to be updated as well.

var-const (approved):

PR #618, opened 2021-08-25
Replace ThrowInvalidArgument with SimpleThrowInvalidArgument to fix iOS absl linker error in Unity SDK

dconeybe:
This a follow-up fix to #616 which additionally replaces `ThrowInvalidArgument` with `SimpleThrowInvalidArgument` to avoid absl-related linker errors in the Unity SDK for iOS.

The "style" of the changes made in this PR were based on those made in other files, such as:

https://github.com/firebase/firebase-cpp-sdk/blob/cebf363f7b25a7c8f200d1381159784df7ee52b3/firestore/src/main/query_main.cc#L196-L206

Googlers can see cl/289114361 for the introduction of `SimpleThrowInvalidArgument` for context.

### DIFF
--- a/firestore/src/main/firestore_main.cc
+++ b/firestore/src/main/firestore_main.cc
@@ -35,6 +35,7 @@
 #include "app/src/include/firebase/future.h"
 #include "app/src/reference_counted_future_impl.h"
 #include "firebase/firestore/firestore_version.h"
+#include "firestore/src/common/exception_common.h"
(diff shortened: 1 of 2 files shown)


ehsannas:
Do we expect the absl linker errors to be resolved anytime soon? If not, should we consider removing `ThrowInvalidArgument`? It may inadvertently be used again and break releases.

PR #622, opened 2021-08-27
Initial Admob 2021 pinned to 20.1.0

DellaBitta:
Create a baseline branch of Admob's 2021 development.

This change removes features that are no longer present in admob: RewardedAds, NativeExpressAds, etc.  Additionally some work has been done to reduce the need for BannerViewHelper.  Further work will be done here, and in the InterstiailAdHelper, when we rework the listener architecture.

### DIFF
--- a/admob/integration_test/src/integration_test.cc
+++ b/admob/integration_test/src/integration_test.cc
@@ -354,6 +354,8 @@ TEST_F(FirebaseAdMobTest, TestBannerView) {
 }
 
 TEST_F(FirebaseAdMobTest, TestBannerViewAlreadyInitialized) {
+  SKIP_TEST_ON_DESKTOP;
+
   static const int kBannerWidth = 320;
   static const int kBannerHeight = 50;
 


Atom14011985:
feature/admob_2021_baseline

Atom14011985:
admob_2021_baseline.diff

Atom14011985:
@Atom14011985 #622

admob/src/android/banner_view_internal_android.cc
```diff
@@ -66,37 +99,131 @@ void DestroyOnDeleteCallback(const Future<void>& result, void* sem_data) {
...
 Future<void> BannerViewInternalAndroid::Initialize(AdParent parent,
                                                    const char* ad_unit_id,
                                                    AdSize size) {
   FutureCallbackData* callback_data =
       CreateFutureCallbackData(&future_data_, kBannerViewFnInitialize);
 
+  if (initialized_) {
+    CompleteFuture(kAdMobErrorAlreadyInitialized, "Ad is already initialized.",
+                   callback_data->future_handle, callback_data->future_data);
+    return GetLastResult(kBannerViewFnInitialize);
```
jonsimantov, line 211: Note that this pattern introduces a race condition that we should try to avoid in new code. See b/110082243 and cl/217546936 for details.

jonsimantov (changes requested):

admob/src/android/banner_view_internal_android.cc
DellaBitta: I believe I fixed this. Please take a look.

admob/src/android/ad_request_converter.cc
```diff
@@ -194,5 +137,38 @@ AdRequestConverter::~AdRequestConverter() {
 
 jobject AdRequestConverter::GetJavaRequestObject() { return java_request_; }
 
+void AdRequestConverter::ConvertRequestConfiguration(AdRequest request) const {
+  JNIEnv* env = ::firebase::admob::GetJNI();
+  jobject builder = env->NewObject(request_config_builder::GetClass(),
```
a-maurice, line 142: If the if isn't true, builder leaks

admob/src/android/ad_request_converter.cc
```diff
@@ -194,5 +137,38 @@ AdRequestConverter::~AdRequestConverter() {
...
+    builder = util::ContinueBuilder(
+        env, builder,
+        env->CallObjectMethod(builder,
+                              request_config_builder::GetMethodId(
+                                  request_config_builder::kSetTestDeviceIds),
+                              test_device_list));
+    env->DeleteLocalRef(test_device_list);
+
+    // Build request configuration.
+    jobject request_configuration = env->CallObjectMethod(
```
a-maurice, line 162: Does request_configuration need to be deleted?

DellaBitta: Good catch.  Please Re-review.

admob/src/android/ad_request_converter.cc
DellaBitta: Fixed, PTAL.

admob/src/android/ad_request_converter.cc
DellaBitta: FWIW this method is removed in the [next PR](https://github.com/firebase/firebase-cpp-sdk/pull/665/files).  Not that we should ever check-in code that leaks memory...

admob/src/ios/banner_view_internal_ios.mm
```diff
@@ -39,15 +40,24 @@
...
                                                AdSize size) {
-  const firebase::FutureHandle handle = CreateFuture(kBannerViewFnInitialize, &future_data_);
-  dispatch_async(dispatch_get_main_queue(), ^{
-    banner_view_ = [[FADBannerView alloc] initWithView:parent
+  FutureCallbackData* callback_data =
+      CreateFutureCallbackData(&future_data_, kBannerViewFnInitialize);
+  if(initialized_) {
+    future_data_.future_impl.Complete(callback_data->future_handle,
+                                      kAdMobErrorAlreadyInitialized,
+                                      "Ad is already initialized.");
```
jonsimantov, line 48: nit: AdMob is already initialized

jonsimantov (approved):

a-maurice (approved):

admob/src/ios/banner_view_internal_ios.mm
DellaBitta: In this case it's the ad that's already initialized, not admob.

jonsimantov (approved):

PR #623, opened 2021-09-01
Specify `targets` in `build.gradle` to improve build times of Firestore's integration tests on Android

dconeybe:
Add the following line to the `build.gradle` files for Firestore's integration tests:

```
targets 'firebase_firestore', 'firebase_auth', 'android_integration_test_main'
```

This configures cmake to only build the necessary targets, instead of building all of the targets. This produces a meaningful build time improvement.


dconeybe:
Nevermind, it's not actually yielding any performance improvements. I think it only helps if you comment out the unnecessary directories in the root `settings.gradle`.

PR #625, opened 2021-09-03
iOS Baseline to build upon for the Admob 2021 SDK

DellaBitta:


DellaBitta:
* removed nativeAdExpress and rewardedVideo iOS impls
* removed extra comma in admob testapp project

PR #626, opened 2021-09-03
Add presubmit trigger with ccache in linux firestore

sunmou99:
Combined "check_trigger" job and "prepare_matrix" job into "check_and_prepare" job.
Added "trigger" env variable. Apply "expanded_matrix, minimal_matrix, auto_diff" with different triggers accordingly.
Removed "auto_diff" from manual trigger, because the logic doesn't work. (Always override by inputs.)
Add presubmit(push) trigger, which run linux firestore with ccache. 
presubmit check: [without ccache](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1236217489) 35 mins vs [with ccache](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1236567562) 18 mins
To understand different [triggers](https://g3doc.corp.google.com/firebase/testing/g3doc/sdk/games-cpp.md?cl=head#ci-triggers).

TODO: Will add branch_protection_rules "build-desktop-ubuntu-latest-openssl" and "test-desktop-ubuntu-latest-openssl" after this PR get merged.
Will have a cache system which enable ccache in more cases.

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -105,17 +105,17 @@ jobs:
         elif [[ "${{ github.event_name }}" == "pull_request" ]]; then
           echo "::set-output name=github_ref::$GITHUB_SHA"
           echo "::set-output name=pr_number::${{ github.event.pull_request.number }}"
-          if [[ "${{ github.event.action }}" == "labeled" && "${{ github.event.action }}" && "${{ github.event.label.name }}" == "${{ env.triggerLabelPrefix }}"* ]]; then
+          if [[ "${{ github.event.action }}" == "labeled" && "${{ github.event.action }}" && "${{ github.event.label.name }}" == ${{ env.triggerLabelPrefix }}* ]]; then
             echo "::set-output name=trigger::label_trigger"
             if [[ "${{ github.event.label.name }}" == "${{ env.triggerLabelQuick }}" ]]; then
               echo "::set-output name=requested_tests::auto"
             else
               echo "::set-output name=requested_tests::expanded"
             fi
-          elif [[ "${{ github.event.action }}" == "closed" && "${{ github.event.pull_request.merged }}"]]; then
+          elif [[ "${{ github.event.action }}" == "closed" && ${{ github.event.pull_request.merged }}]]; then
             echo "::set-output name=trigger::postsubmit_trigger"
             echo "::set-output name=requested_tests::auto"
-          elif [[ "opened,reopened,synchronize" == *"${{ github.event.action }}"*  ]]; then
+          elif [[ "opened,reopened,synchronize" == *${{ github.event.action }}*  ]]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/integration_tests.yml
```diff
@@ -49,189 +49,171 @@ env:
...
+          echo "::set-output name=github_ref::$GITHUB_SHA"
+          if [[ "${{ github.event.action }}" == "labeled" && "${{ github.event.label.name }}" == "${{ env.triggerLabelPrefix }}"* ]]; then
+            echo "::set-output name=trigger::label_trigger"
+            echo "::set-output name=pr_number::${{ github.event.pull_request.number }}"
+            if [[ "${{ github.event.label.name }}" == "${{ env.triggerLabelQuick }}" ]]; then
+              echo "::set-output name=requested_tests::auto"
+            else
+              echo "::set-output name=requested_tests::expanded"
+            fi
+          elif [[ "${{ github.event.action }}" == "closed" && "${{ github.event.pull_request.merged }}" == "true" ]]; then
```
jonsimantov, line 116: Just a nit, isn't it slightly safer to say:
("${{ github.event.action }}" == "closed") && "${{ github.event.pull_request.merged }}"

(in case github.event.pull_request.merged happens to be "1" instead of "true" or some other non-zero value.

.github/workflows/integration_tests.yml
```diff
@@ -49,189 +49,171 @@ env:
...
+            elif [[ "${{github.event.inputs.test_pull_request}}" == *:* ]]; then
+              # If specified as pr:commit_hash, split them.
+              echo "::set-output name=github_ref::$(echo ${{ github.event.inputs.test_pull_request }} | cut -d: -f2)"
+              echo "::set-output name=pr_number::$(echo ${{ github.event.inputs.test_pull_request }} | cut -d: -f1)"
+            else
+              # Just the PR specified, use refs/pull/<number>/merge as the ref.
+              echo "::set-output name=github_ref::refs/pull/${{github.event.inputs.test_pull_request}}/merge"
+              echo "::set-output name=pr_number::${{ github.event.inputs.test_pull_request }}"
+            fi
+          elif [[ "${{ github.event.inputs.test_pull_request }}" == "sdk" ]]; then
```
jonsimantov, line 96: Remind me, did this value already exist? Why does it need to be set if we have the test_packaged_sdk input?

jonsimantov (approved):
This seems mostly good, just a couple quick questions.

.github/workflows/integration_tests.yml
sunmou99: This value already exist. https://github.com/firebase/firebase-cpp-sdk/blob/main/.github/workflows/cpp-packaging.yml#L725
Because it means it's from **scheduled** packaging workflow, thus it can generate nightly report. test_packaged_sdk only indicates that it's from packaging workflow.

PR #654, opened 2021-09-13
Add some licenses to `NOTICES`

var-const:


var-const:
@jonsimantov  It looks like the diffing algorithm doesn't produce a great delta. If this makes reviewing difficult and you would prefer me to just add all the newer licenses to the end, just let me know.

jonsimantov (commented):
Looks fine (you're right, diff is silly) but is the vcpkg update part of it?

var-const:
Thanks for noticing! Fixed.

For completeness' sake, I didn't change any of the existing text except removing trailing whitespace in a couple of places (never in the license text) and fixing one typo in a header (`zlib crompession` -> `zlib compression`).

jonsimantov (approved):

PR #656, opened 2021-09-13
Fix log file for when both Android and iOS are present.

jonsimantov:
Was accidentally including both iOS and Android logs in both sections - this filters out the correct one for each section.


a-maurice (approved):

jonsimantov:
Fixed run: https://github.com/firebase/firebase-cpp-sdk/pull/657

PR #667, opened 2021-09-17
Change the way to achieve token while init the app

cynthiajoan:
The current way to grab token is starting a new IntentService, and tried to get IID token. Since we already have the new API to get installations token with a Future function call, it's better to use the new way and deprecate the IID related code.

This fix will address the issue https://github.com/firebase/quickstart-unity/issues/1088


DellaBitta (approved):

release_build_files/readme.md
```diff
@@ -567,10 +567,16 @@ code.
 
 ## Release Notes
 
+### Future Release
+-   Changes
+    -   Messaging (Android): Fixes an issue to receive token when
```
DellaBitta, line 572: `Fixed` instad of `Fixes`

release_build_files/readme.md
```diff
@@ -567,10 +567,16 @@ code.
 
 ## Release Notes
 
+### Future Release
```
jonsimantov, line 570: Please change this to the actual version number of the upcoming release at the next opportunity, per http://go/firebase-cpp-release.

PR #674, opened 2021-09-21
M1 Mac (arm64) open source build

jonsimantov:
This enables cross-compiling for arm64 Mac, for open-source builds only. (This changes the default Xcode version to 12.4, as Xcode 11 doesn't support arm64. It'll still build with 11.7 in the expanded matrix.)

Unit tests are built, but not run, on arm64. (They can't be run until GitHub has hosted ARM Mac runners, or we add our own.)
  
Packaged builds will be done in a subsequent PR, as they require switching to an alternate version of objcopy.


.github/workflows/desktop.yml
```diff
@@ -114,21 +122,25 @@ jobs:
...
+            echo VCPKG_EXTRA_HASH-xcode${{matrix.xcode_version}} >> $GITHUB_ENV
+          fi
 
       - name: Cache vcpkg C++ dependencies
         id: cache_vcpkg
         uses: actions/cache@v2
         with:
           path: external/vcpkg/installed
-          key: dev-vcpkg-${{ matrix.architecture }}-${{ env.VCPKG_SUFFIX }}-${{ matrix.msvc_runtime }}-${{ hashFiles(format('{0}', env.VCPKG_RESPONSE_FILE)) }}-${{ hashFiles('.git/modules/external/vcpkg/HEAD') }}
+          key: dev-vcpkg-${{ matrix.architecture }}-${{ env.VCPKG_SUFFIX }}-${{ matrix.msvc_runtime }}-${{ hashFiles(format('{0}', env.VCPKG_RESPONSE_FILE)) }}-${{ hashFiles('.git/modules/external/vcpkg/HEAD') }}${{ env.VCPKG_EXTRA_HASH }}
```
DellaBitta, line 135: I guess this is up to opinion, but based on the previous pattern this is missing an `-` before the new suffix.

.github/workflows/desktop.yml
```diff
@@ -114,21 +122,25 @@ jobs:
...
 
       - name: Cache ccache files
         if: startsWith(matrix.os, 'ubuntu') || startsWith(matrix.os, 'macos')
         id: cache_ccache
         uses: actions/cache@v2
         with:
           path: ccache_dir
-          key: dev-test-ccache-${{ env.MATRIX_UNIQUE_NAME }}
+          key: dev-test-ccache-${{ env.MATRIX_UNIQUE_NAME }}${{ env.VCPKG_EXTRA_HASH }}
```
DellaBitta, line 143: and here.

CMakeLists.txt
```diff
@@ -402,13 +425,15 @@ if(DESKTOP)
         ${UWEBSOCKETS_SOURCE_DIR}/src/WebSocket.cpp)
     if(MSVC)
       set(websockets_additional_defines
+        ${websockets_additional_defines}
```
DellaBitta, line 428: you might be able to use [list append](https://cmake.org/cmake/help/v3.12/command/list.html#append) here.  Just a thought.

.github/workflows/desktop.yml
jonsimantov: The new suffix contains the "-"

CMakeLists.txt
jonsimantov: Hmm, do we tend to prefer one or the other in our cmake files? I think I see more instances of set(${thing} ${thing} additional items) rather than list append

.github/workflows/desktop.yml
DellaBitta: Got it.

CMakeLists.txt
DellaBitta: Meh.  Thought it might be easier to read.  I don't have a strong opinion about it, though.

DellaBitta (approved):

PR #676, opened 2021-09-23
Change Git tag of googletest to the 1.11.0 release.

jonsimantov:
This is already the version set in download_googletest.py, and is the
version that the integration tests have been built/tested with.

(Because the integration tests do some internal GoogleTest stuff, it's
risky to pin to the main branch.)


sunmou99 (approved):
LGTM. 
Two things that may be helpful. 1. add comments and let others know why we use "1.11.0" not "master". 2. If the git tag is likely to change in the future, make it a variable in CMake.

jonsimantov:
> 1. add comments and let others know why we use "1.11.0" not "master". 
Done.

> 2. If the git tag is likely to change in the future, make it a variable in CMake.
I don't think it will change any time soon - the only reason it changed from 1.10 to 1.11 is because Firestore started using an exception throwing test.

PR #677, opened 2021-09-23
ARM Mac open-source and packaged build

jonsimantov:
This enables cross-compiling for arm64 Mac. This changes the default Xcode version to 12.2, as Xcode 11 doesn't support arm64, and 12.0 fails with vcpkg. It'll still build with 11.7 in the expanded matrix and omit arm64.

Unit tests are built, but not run, on arm64. (They can't be run until GitHub has hosted ARM Mac runners, or we add our own.)

This PR adds a --arch option to the testapp builder but does not yet build for arm64 in the integration_tests workflow.

For packaged builds, we now build LLVM binutils on Mac, in addition to GNU binutils. LLVM binutils are required for the merge_libraries step on Mac, as GNU binutils does not support outputting ARM64 Mach-O files properly. (We still need GNU binutils on Mac for c++filt). Because LLVM binutils can take a long time to build, we cache it.


jonsimantov:
Successful packaging run: https://github.com/firebase/firebase-cpp-sdk/actions/runs/1268032776

Integration test failures in both source and packaged runs are flakes only.

.github/workflows/cpp-packaging.yml
```diff
@@ -105,13 +107,47 @@ jobs:
...
+
+      - name: Build LLVM (Mac only) 
+        # Only build LLVM if we don't already have binaries from the cache.
+        if: ${{ matrix.tools_platform == 'darwin' && !steps.cache_llvm.outputs.cache-hit }}
+        run: |
+          mkdir llvm-src/llvm/build
+          cd llvm-src/llvm/build
+          cmake -DLLVM_ENABLE_PROJECTS=clang -DCMAKE_BUILD_TYPE=Release ..
+          cmake --build . -j 3 --target llvm-objcopy --target llvm-nm --target llvm-ar
+          cd -
```
DellaBitta, line 144: is this `cd` needed?

release_build_files/readme.md
```diff
@@ -337,40 +337,41 @@ information).
...
-|                               | firebase.xcframework
-Firebase Dynamic Links (stub)   | firebase_dynamic_links.xcframework
-|                               | firebase.xcframework
-Firebase Installations (stub)   | firebase_installations.xcframework
-|                               | firebase.xcframework
-Firebase Cloud Messaging (stub) | firebase_messaging.xcframework
-|                               | firebase.xcframework
-
-The provided libraries have been tested using Xcode 12.0. When building C++
+Firebase Authentication         | firebase_auth.framework
```
DellaBitta, line 347: We're not providing xcframeworks anymore?  I thought they were the new thing, not the old thing..?

DellaBitta (approved):
Approved with some questions.

release_build_files/readme.md
jonsimantov: We are for iOS. We aren't for OSX, and never were. This was in error.

.github/workflows/cpp-packaging.yml
jonsimantov: Ah I guess not, it's leftover from before the caching when it was 1 step

release_build_files/readme.md
DellaBitta: Ahhh, I see.

jonsimantov:
This enables cross-compiling for arm64 Mac. This changes the default Xcode version to 12.2, as Xcode 11 doesn't support arm64, and 12.0 fails with vcpkg. It'll still build with 11.7 in the expanded matrix and omit arm64.

Unit tests are built, but not run, on arm64. (They can't be run until GitHub has hosted ARM Mac runners, or we add our own.)

This PR adds a --arch option to the testapp builder but does not yet build for arm64 in the integration_tests workflow.

For packaged builds, we now build LLVM binutils on Mac, in addition to GNU binutils. LLVM binutils are required for the merge_libraries step on Mac, as GNU binutils does not support outputting ARM64 Mach-O files properly. (We still need GNU binutils on Mac for c++filt). Because LLVM binutils can take a long time to build, we cache it.

PR #679, opened 2021-09-24
Admob - AdSize and AdRequest updates

DellaBitta:
Updated the AdSize and AdRequest structures:
 - Converted them to classes.
 - AdRequest now has methods to manipulate the content url, keywords and network extras.  
 - Extras are are now collections with mediation class names keys.  The class names now dynamically resolved to jclasses on on Android and GADSize objects on iOS, as these objects are required as part of the Admob phone SDK API when setting extras.
 - AdSize now provides methods for Adaptive sizes, and includes constants for common ad sizes.
 
Banner and Interstitial ads were updated to use these new structures.

Updated Itests
 - Some unit testing is done on these structures.
 - Added pauses between banner ad move commands as they were moving too quickly for Android phones to properly render them (they flashed as blank black boxes.)

[iTest execution](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1271413043)
[Packaging execution 1](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1293182830) [Packaging execution 2, expanded](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1295146882) - builds fine, some test failures in unrelated storage and messaging code.

### DIFF
--- a/admob/src/android/ad_request_converter.cc
+++ b/admob/src/android/ad_request_converter.cc
@@ -43,8 +43,9 @@ METHOD_LOOKUP_DEFINITION(ad_request_builder,
                          ADREQUESTBUILDER_METHODS);
 
 jobject GetJavaAdRequestFromCPPAdRequest(const AdRequest& request,
-                                         admob::AdMobError& error) {
-  error = kAdMobErrorNone;
+                                         admob::AdMobError* error) {
+  FIREBASE_ASSERT(error);
+  *error = kAdMobErrorNone;
 
   JNIEnv* env = ::firebase::admob::GetJNI();
   jobject builder = env->NewObject(
@@ -64,7 +65,7 @@ jobject GetJavaAdRequestFromCPPAdRequest(const AdRequest& request,
           "Failed to resolve extras class. Check that \"%s\""
           " is present in your APK.",
           adapter_name.c_str());
-      error = kAdMobErrorAdNetworkClassLoadError;
+      *error = kAdMobErrorAdNetworkClassLoadError;
       env->DeleteLocalRef(builder);
       return nullptr;
     }
--- a/admob/src/android/ad_request_converter.cc
+++ b/admob/src/android/ad_request_converter.cc
@@ -53,6 +53,9 @@ jobject GetJavaAdRequestFromCPPAdRequest(const AdRequest& request,
       ad_request_builder::GetMethodId(ad_request_builder::kConstructor));
 
   // Network Extras.
+  // The map associates class names of mediation adapters with a 
+  // key value pairs, the extras, to send to those medation adapters.
+  // e.g. Mediation_Map < class name, Extras_Map < key, value > >
   const std::map<std::string, std::map<std::string, std::string>>& extras =
       request.extras();
   for (auto adapter_iter = extras.begin(); adapter_iter != extras.end();
--- a/admob/src/android/ad_request_converter.cc
+++ b/admob/src/android/ad_request_converter.cc
@@ -42,24 +42,14 @@ METHOD_LOOKUP_DEFINITION(ad_request_builder,
                          "com/google/android/gms/ads/AdRequest$Builder",
                          ADREQUESTBUILDER_METHODS);
 
-AdRequestConverter::AdRequestConverter(const AdRequest& request) {
+jobject GetJavaAdRequestFromCPPAdRequest(const AdRequest& request,
+                                         admob::AdMobError& error) {
+  error = kAdMobErrorNone;
+
   JNIEnv* env = ::firebase::admob::GetJNI();
   jobject builder = env->NewObject(
       ad_request_builder::GetClass(),
       ad_request_builder::GetMethodId(ad_request_builder::kConstructor));
-
-  // Keywords.
-  const std::unordered_set<std::string>& keywords = request.keywords();
-  for (auto keyword = keywords.begin(); keyword != keywords.end(); ++keyword) {
-    jstring keyword_str = env->NewStringUTF((*keyword).c_str());
-    builder = util::ContinueBuilder(
-        env, builder,
-        env->CallObjectMethod(
-            builder,
-            ad_request_builder::GetMethodId(ad_request_builder::kAddKeyword),
-            keyword_str));
-    env->DeleteLocalRef(keyword_str);
-  }
 
   // Network Extras.
   const std::map<std::string, std::map<std::string, std::string>>& extras =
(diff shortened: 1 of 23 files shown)


admob/src/ios/banner_view_internal_ios.mm
```diff
@@ -79,11 +79,19 @@
...
+    AdMobError error = kAdMobErrorNone;
+    GADRequest *ad_request = GADRequestFromCppAdRequest(*request_copy, &error);
     delete request_copy;
-    // Make the banner view ad request.
-    [banner_view_ loadRequest:ad_request];
+    if(ad_request==nullptr) {
+      if(error==kAdMobErrorNone) {
+        error = kAdMobErrorInternalError;
+      }
+      CompleteLoadFuture(error, "");
```
jonsimantov, line 90: Might be nice to also have an error string passed back from GADRequestFromCppAdRequest too, rather than it just LogError-ing. Same in interstitial.

jonsimantov: (also a default error message for Internal Error)

admob/src/android/ad_request_converter.cc
```diff
@@ -35,70 +37,88 @@
...
-        env->CallObjectMethod(
-            builder,
-            ad_request_builder::GetMethodId(ad_request_builder::kAddKeyword),
-            keyword_str));
-    env->DeleteLocalRef(keyword_str);
-  }
-
   // Network Extras.
-  if (request.extras_count > 0) {
+  const std::map<std::string, std::map<std::string, std::string>>& extras =
```
jonsimantov, line 56: Add a comment of what is in this map?

admob/src/android/ad_request_converter.cc
```diff
@@ -35,70 +37,88 @@
...
     }
 
-    jclass admob_adapter_class =
-        util::FindClass(env, admob::GetActivity(), kAdMobAdapterClassName);
-
-    // If the adapter wasn't found correctly, the app has a serious problem.
-    FIREBASE_ASSERT_MESSAGE(
-        admob_adapter_class,
-        "Failed to locate the AdMobAdapter class for extras. Check that "
-        "com.google.ads.mediation.admob.AdMobAdapter is present in your APK.");
```
jonsimantov, line 87: minor nit: Not sure our assert macro works this way, but can you re-use kAdMobAdapterClassName in this string? (if it's complicated at all, don't bother.)

admob/src/android/banner_view_internal_android.cc
```diff
@@ -120,7 +117,8 @@ struct BannerViewInternalInitializeData {
   BannerViewInternalInitializeData()
       : activity_global(nullptr),
         ad_view(nullptr),
-        banner_view_helper(nullptr) {}
```
jonsimantov, line 123: Just putting this out there, since we don't need to support STLPort any more, you could de-void the internal initialize function and use std::function I think.

admob/src/common/admob_common.cc
```diff
@@ -50,6 +51,77 @@ DEFINE_FIREBASE_VERSION_STRING(FirebaseAdMob);
 static CleanupNotifier* g_cleanup_notifier = nullptr;
 const char kAdMobModuleName[] = "admob";
 
+const AdSize AdSize::kBanner(/*width=*/320, /*height=*/50);
```
jonsimantov, line 54: Can you reference where these come from?

admob/src/include/firebase/admob/types.h
```diff
@@ -17,7 +17,9 @@
 #ifndef FIREBASE_ADMOB_SRC_INCLUDE_FIREBASE_ADMOB_TYPES_H_
 #define FIREBASE_ADMOB_SRC_INCLUDE_FIREBASE_ADMOB_TYPES_H_
 
+#include <map>
 #include <string>
+#include <unordered_set>
```
jonsimantov, line 22: I'm sure it's fine, but just make sure both Android stlport variants that we still support have std::unordered_set.

DellaBitta: I didn't realize that we still support stlport.  Our [release notes say that we don't](https://firebase.google.com/support/release-notes/cpp-relnotes#version_800_-_june_18_2021).

Which versions should I be checking?

DellaBitta: Or did you mean that we should ensure that it's in libc++?  Looks like it was [part of the intiial port](https://github.com/llvm/llvm-project/blame/d5db71d19f11d7c31257066aea6bd41ef04f28b7/libcxx/include/unordered_set).

admob/src/android/ad_request_converter.cc
DellaBitta: This code and the defaulting to kAdMobAdapterClassName was removed.  We now return errors if the user-supplied class names could not be resolved, and we log the name.   See `if (adapter_class == nullptr) {` above.

admob/src/ios/banner_view_internal_ios.mm
DellaBitta: Done.

admob/src/common/admob_common.cc
DellaBitta: Added a comment!

admob/src/android/banner_view_internal_android.cc
DellaBitta: Sorry, I'm not quite sure I understand. Could we talk about this offline and have it part of the Banner View rewrite (which has it's own story.)

admob/src/common/admob_common.cc
DellaBitta: Which doesn't show up in the code snippet above.  I guess you have to go into review mode to see it.

admob/src/include/firebase/admob/types.h
jonsimantov: gnustl, I think.

admob/src/android/banner_view_internal_android.cc
jonsimantov: SG, let's do this offline. (basically, we can stop using void* context pointers in callbacks, and just use std::function instead.)

admob/src/android/ad_request_converter.cc
jonsimantov: Got it, thanks.

admob/src/include/firebase/admob/types.h
jonsimantov: (basically, just run cpp packaging on this and see if both Android STL versions build.)

jonsimantov (approved):
LGTM assuming packaging succeeds (at least both Android STL variants).

PR #682, opened 2021-09-27
LoadAd operations return LoadAdResults

DellaBitta:
LoadAd operations on BannerView and Interstitial Ads now return a LoadAdResult.  On error this contains information from the AdMob service as to why a failure occurred.  The C++ SDK also uses it to return errors that may have occurred when attempting to marshal the Android and iOS SDKs.

The implementation obfuscates Objective C pointers and Android references via Internal structs. These are forward declared in our public API and defined internally.  These structs are passed to AdapterResponseInfo, AdResult, LoadAdResult and ResponseInfo class constructors.

Other changes include:
 - Updated the use of Futures in BannerView and InterstitialAd to use SafeFutures.
 - Removed the use GetLastResult to construct future objects, and its use opens up the potential issue of race conditions. Replaced this implementation with MakeFuture.
 - Moved LoadAd error code mappings from Android and iOS to C++ code.
 - Updated integration tests to include the following scenarios for both Interstitial and Banner Ads
   - Duplicate initialization invocations on the same object.
   - Duplicate loadAd requests on the same object.
   - The use of an errant AdUnitId string
 - Reenabled a multithreaded deletion test that was previously disabled due to a now obsolete Android bug
 - Added a LoadAd integration test that doesn't require user interaction, so that we can run it on FTL.
 - Updated iOS Admob SDK to 8+, and reworked the InterstitialDelegates and Loading process to match the new 8.0 flow.

[Integration Test run](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1408627388)


admob/integration_test/src/integration_test.cc
```diff
@@ -420,7 +431,20 @@ TEST_F(FirebaseAdMobTest, TestBannerView) {
 
   // Load the banner ad.
   firebase::admob::AdRequest request = GetAdRequest();
-  WaitForCompletion(banner->LoadAd(request), "LoadAd");
+  firebase::Future<firebase::admob::LoadAdResult> load_ad_future =
+      banner->LoadAd(request);
+  WaitForCompletion(load_ad_future, "LoadAd");
+
+  const firebase::admob::LoadAdResult* result_ptr = load_ad_future.result();
```
jonsimantov, line 438: assert result_ptr isn't null

DellaBitta: Done.

admob/src/include/firebase/admob/types.h
```diff
@@ -82,13 +92,142 @@ enum AdMobError {
...
+  const std::string& message() const;
+
+  /// Returns a log friendly string version of this object.
+  const std::string& ToString() const;
+
+  /// A domain string which represents an undefined error domain.
+  ///
+  /// The Admob SDK returns this domain for domain() method invocations when
+  /// converting error information from legacy mediation adapter callbacks.
+  static const char* kUndefinedDomain;
```
jonsimantov, line 173: nit: I think this needs to be a static const char* const, or a static const char[], otherwise the pointer can be reassigned.

admob/src/ios/load_ad_result_ios.mm
```diff
@@ -0,0 +1,51 @@
...
+#include "admob/src/include/firebase/admob.h"
+#include "admob/src/ios/response_info_ios.h"
+#include "app/src/util_ios.h"
+
+namespace firebase {
+namespace admob {
+
+LoadAdResult::LoadAdResult() : AdResult(), response_info_() {}
+
+LoadAdResult::LoadAdResult(const LoadAdResultInternal& load_ad_result_internal)
```
jonsimantov, line 30: Let me make sure I've got this clear, because it seems odd to me...

- AdResult contains an AdResultInternal - this is similar to the model we use for the other modern SDKs.
- LoadAdResult is a subclass of AdResult, and does *not* contain a LoadAdResultInternal.
  - Instead, LoadAdResultInternal contains an AdResultInternal
  - When a LoadAdResult is created, the LoadAdResultInternal's AdResultInternal is set in the base AdResult class.

Does that sound like what's happening? If so, LoadAdResultInternal seems weird and extraneous, no?

admob/src_java/com/google/firebase/admob/internal/cpp/BannerViewHelper.java
```diff
@@ -175,9 +173,9 @@ public void run() {
             notifyStateChanged(
                 mBannerViewInternalPtr, ConstantsHelper.AD_VIEW_CHANGED_BOUNDING_BOX);
             completeBannerViewFutureCallback(
-                callbackDataPtr,
-                ConstantsHelper.CALLBACK_ERROR_NONE,
-                ConstantsHelper.CALLBACK_ERROR_MESSAGE_NONE);
+               callbackDataPtr,
```
jonsimantov, line 176: Can we run these through https://github.com/google/google-java-format ?

For bonus points let's file a ticket to add google-java-format to our code formatting script. :)

app/src/reference_counted_future_impl.h
```diff
@@ -25,7 +25,6 @@
 #include "app/src/include/firebase/internal/common.h"
 #include "app/src/mutex.h"
 
-
```
jonsimantov, line 28: nit: remove, unless it's important that this PR touches this file

admob/src/ios/admob_ios.mm
```diff
@@ -156,5 +164,72 @@ void Terminate() {
...
+  CompleteLoadAdResult(callback_data, /*error=*/nullptr, error_code, error_message);
+}
+
+void CompleteLoadAdIOSResult(FutureCallbackData<LoadAdResult>* callback_data,
+                             NSError *gad_error) {
+  FIREBASE_ASSERT(callback_data);
+
+  AdMobError error_code = MapADErrorCodeToCPPErrorCode(gad_error.code);
+  CompleteLoadAdResult(callback_data, gad_error, error_code,
+                       /*error_message=*/"");
```
jonsimantov, line 231: Should we use gad_error.localizedDescription here?

admob/src/android/admob_android.cc
```diff
@@ -441,33 +453,151 @@ JNIEnv* GetJNI() {
...
+    adr->is_successful = false;
+    adr->message = error_message;
+    adr->domain = "SDK";
+    adr->to_string = std::string("Internal error: ") + adr->message;
+    future_error_message = adr->message;
+  }
+
+  // Invoke a friend of LoadAdResult to have it invoke the LoadAdResult
+  // protected constructor with the LoadAdResultInternal data.
+  AdmobInternal::CompleteLoadAdFuture(
```
jonsimantov, line 516: Why doesn't CompleteLoadAdFuture just take an AdResultInternal instead of wrapping it in a LoadAdResultInternal?

admob/src/android/admob_android.cc
```diff
@@ -441,33 +453,151 @@ JNIEnv* GetJNI() {
...
+  const AdMobError error_code =
+      MapAndroidAdRequestErrorCodeToCPPErrorCode(j_error_code);
+
+  CompleteLoadAdAndroidErrorResult(env, data_ptr, j_load_ad_error, error_code,
+                                   j_error_message);
+}
+
+// Internal Errors use AdMobError codes.
 extern "C" JNIEXPORT void JNICALL
-Java_com_google_firebase_admob_internal_cpp_BannerViewHelper_completeBannerViewFutureCallback(
```
jonsimantov, line 463: FYI this is an opportunity to rename these native methods. Since we have to call RegisterNatives on them anyway (due to the way we load the Android classes, they aren't bound automatically), we could shorten / simplify the names for readability. Up to you.

admob/integration_test/src/integration_test.cc
```diff
@@ -612,15 +636,80 @@ TEST_F(FirebaseAdMobTest, TestBannerViewAlreadyInitialized) {
...
+  firebase::admob::BannerView* banner = new firebase::admob::BannerView();
+  WaitForCompletion(banner->Initialize(app_framework::GetWindowContext(),
+                                       kBannerAdUnit, banner_ad_size),
+                    "Initialize");
+
+  // Load the banner ad.
+  firebase::admob::AdRequest request = GetAdRequest();
+  firebase::Future<firebase::admob::LoadAdResult> first_load_ad =
+      banner->LoadAd(request);
+  firebase::Future<firebase::admob::LoadAdResult> second_load_ad =
```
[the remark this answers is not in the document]
jonsimantov: If it's possible that it might flake, should we use the flaky test macros?

admob/src/ios/adapter_response_info_ios.mm
```diff
@@ -0,0 +1,58 @@
...
+  ad_result_ = AdResult(ad_result_internal);
+  
+  adapter_class_name_ = util::NSStringToString(
+    internal.ad_network_response_info.adNetworkClassName);
+  
+  // ObjC latency is an NSInterval, which is in seconds.  Convert to millis.
+  latency_ = (int64_t)NSInteger(internal.ad_network_response_info.latency) *
+    (int64_t)1000;
+
+  NSString *to_string = [[NSString alloc]initWithFormat:@"AdapterResponseInfo"
```
jonsimantov, line 49: nit: I think a better way to do this is: [NSString stringWithFormat:...]

jonsimantov (approved):
Mostly minor nits, though I'm not clear on why LoadAdResultInternal exists. Approving anyway (assuming nits are addressed), but I wouldn't mind some more comments explaining the flow.

admob/integration_test/src/integration_test.cc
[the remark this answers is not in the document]
DellaBitta: I don't think we need them at this point.  I've run this test a lot, especially with some fixes in the next PR, and it hasn't flaked yet.

As of now, the internals of loadAd will defer operations to a function on the main thread.  This redirection forces this invocation to return before the add load attempt is ever made.  Still, I think the note is a good idea in case that ever changes.

admob/src/include/firebase/admob/types.h
DellaBitta: Done.

admob/src/ios/load_ad_result_ios.mm
DellaBitta: Yes, you're right.  It once had more information but I have since removed the other fields.  I'll move `LoadAdResultInternal` and just use `AdResultInternal` in all cases.

admob/src/include/firebase/admob/types.h
DellaBitta: Nice, thanks. Done!

admob/src/common/banner_view_internal.cc
```diff
@@ -74,6 +74,11 @@ Future<void> BannerViewInternal::GetLastResult(BannerViewFn fn) {
       future_data_.future_impl.LastResult(fn));
```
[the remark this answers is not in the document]
DellaBitta: Added an assertion to prevent its misuse.

DellaBitta:
> Mostly minor nits, though I'm not clear on why LoadAdResultInternal exists. Approving anyway (assuming nits are addressed), but I wouldn't mind some more comments explaining the flow.

I believe I fixed this.  PTAL.

admob/src_java/com/google/firebase/admob/internal/cpp/BannerViewHelper.java
DellaBitta: Done.

admob/src/ios/load_ad_result_ios.mm
DellaBitta: Done.

app/src/reference_counted_future_impl.h
DellaBitta: Done.

admob/src/android/admob_android.cc
DellaBitta: Fixed.

admob/src/android/admob_android.cc
DellaBitta: I will do it in a followup PR since I have another change based on this branch and I'd like to prevent as many merge conflicts as possible.

admob/src/ios/admob_ios.mm
DellaBitta: Done.

admob/src/ios/adapter_response_info_ios.mm
DellaBitta: Nice, thanks.  Done!

admob/src/include/firebase/admob/types.h
```diff
@@ -309,6 +447,66 @@ struct BoundingBox {
...
+  explicit ResponseInfo(const ResponseInfoInternal& internal);
+
+  std::vector<AdapterResponseInfo> adapter_responses_;
+  std::string mediation_adapter_class_name_;
+  std::string response_id_;
+  std::string to_string_;
+};
+
+/// @brief Information about why an ad load operation failed.
+class LoadAdResult : public AdResult {
```
jonsimantov, line 489: One more question - the only thing this adds on top of AdResult is the ResponseInfo, right? Is there any reason that shouldn't just be in the base AdResult and avoid subclassing at all? Seems like a lot of mental overhead to subclass AdResult just to add one field.

PR #708, opened 2021-10-25
Fix Firestore build resulting from the rename of auth to credentials in the iOS SDK

dconeybe:
Note that this PR will not build until the iOS dependencies are updated to 8.9.0 (currently they are 8.8.0). This PR will not be merged; instead, it will be merged into the branch that updates the iOS dependencies to 8.9.0.

I tested this locally on Mac and Linux desktop and on iOS simulator.


dconeybe:
These changes have been merged in #714. This PR is now obsolete.

PR #723, opened 2021-11-02
Remove LoadAdResult from new Admob.

DellaBitta:
### Description
Remove LoadAdResult and fold it's data into AdResult.

The ResponseInfo member of LoadAdResult now resides in AdResult and will only be populated in Results that stem from `loadAd` requests.

**Note:** With this new class / data member topography, there isn't an order in which I could declare AdResult in types.h without the compiler complaining that it needed a definition of its ResponseInfo member.  I solved this by making the member a _pointer_ to a ResponseInfo and this made the compiler very happy.  We're friends now.

***
### Testing
Local testing on iOS and Android phones.
**[iTestRun](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1413601990)**
***

### Type of Change
***

#### Notes
N/A.


admob/src/include/firebase/admob/types.h
```diff
@@ -186,6 +190,10 @@ class AdResult {
   /// Gets the message describing the error.
   const std::string& message() const;
 
+  /// Gets the ResponseInfo if an loadAd error occurred, with a collection of
```
jonsimantov, line 193: Will there always be a responseinfo? What if it's the situation where it wasn't, will it just be empty?

jonsimantov (approved):

admob/src/include/firebase/admob/types.h
DellaBitta: Yes, the call to this method will return an emtpy ResponseInfo object.

DellaBitta: Updated the comments.

PR #733, opened 2021-11-03
If simulator  install ios app failed, reset simulator and try again

sunmou99:
### Description
Test on Simulator sometimes failed when install iOS app. https://github.com/firebase/firebase-cpp-sdk/runs/4086443522?check_suite_focus=true#step:9:3602
```
Install testapp: xcrun simctl install 3629D0C4-107A-472E-A79F-A62F60B396DD /Users/runner/work/firebase-cpp-sdk/firebase-cpp-sdk/testapps/auth/integration_test.app
An error was encountered processing the command (domain=IXUserPresentableErrorDomain, code=1):
This app could not be installed at this time.
Failed to load Info.plist from bundle at path /Users/runner/Library/Developer/CoreSimulator/Devices/3629D0C4-107A-472E-A79F-A62F60B396DD/data/Library/Caches/com.apple.mobile.installd.staging/temp.upi3zw/extracted/Payload/integration_test.app
Failed to load Info.plist from bundle at path /Users/runner/Library/Developer/CoreSimulator/Devices/3629D0C4-107A-472E-A79F-A62F60B396DD/data/Library/Caches/com.apple.mobile.installd.staging/temp.upi3zw/extracted/Payload/integration_test.app
Underlying error (domain=MIInstallerErrorDomain, code=35):
	Failed to load Info.plist from bundle at path /Users/runner/Library/Developer/CoreSimulator/Devices/3629D0C4-107A-472E-A79F-A62F60B396DD/data/Library/Caches/com.apple.mobile.installd.staging/temp.upi3zw/extracted/Payload/integration_test.app
```

Looks like a flakiness in iOS Simulator.  Reset and reboot the simulator could possibly solve the issue:
https://stackoverflow.com/questions/40327935/failed-to-load-info-plist-from-bundle

***
### Testing
See the test report in the comments
***

### Type of Change
Place an `x` the applicable box:
- [x] Bug fix. Add the issue # below if applicable.
- [ ] New feature. A non-breaking change which adds functionality.
- [ ] Other, such as a build process or documentation change.
***

PR #734, opened 2021-11-04
Fix test on emulator workflow failures

sunmou99:
### Description
Test on Emualtor sometimes failed when install Android app. https://github.com/firebase/firebase-cpp-sdk/runs/4078767834?check_suite_focus=true
```
Install testapp: adb install /Users/runner/work/firebase-cpp-sdk/firebase-cpp-sdk/testapps/firestore/iti-debug.apk
cmd: Failure calling service package: Broken pipe (32)
adb: failed to install /Users/runner/work/firebase-cpp-sdk/firebase-cpp-sdk/testapps/firestore/iti-debug.apk: cmd: Failure calling service package: Broken pipe (32)
Performing Streamed Install
```
This article suggests delete the old emulator and create a new one :
https://medium.com/@aedwin905/solved-failure-calling-service-package-broken-pipe-32-3b860c7e04bb

Also saw some errors when downloading emulator, uninstalling apps.

Thus, I made the following changes:
1. For critical steps, I added 5 minutes timeout and 3 times retry. 
2. Before retry the step, I also reset and reboot the emulator. 
3. Test on emulator timeout in 90 minutes.
4. Updated the build-tool version to 29.0.2 

***
### Testing
See the testing result in the comment
***

### Type of Change
***

### DIFF
--- a/scripts/gha/test_simulator.py
+++ b/scripts/gha/test_simulator.py
@@ -513,7 +513,14 @@ def _setup_android(platform_version, build_tool_version, sdk_id, retry=1):
 
   command = "yes | sdkmanager --licenses"
   logging.info("Accept all licenses: %s", command)
-  subprocess.run(command, shell=True, check=False)
+  if retry > 1:
+    try:
+      subprocess.run(command, shell=True, check=False, timeout=_CMD_TIMEOUT)
+    except:
+      _setup_android(platform_version, build_tool_version, sdk_id, retry-1)
+      return
+  else:
+    subprocess.run(command, shell=True, check=False, timeout=_CMD_TIMEOUT)
 
   args = ["sdkmanager", sdk_id]
   logging.info("Download an emulator: %s", " ".join(args))
@@ -524,12 +531,18 @@ def _setup_android(platform_version, build_tool_version, sdk_id, retry=1):
       _setup_android(platform_version, build_tool_version, sdk_id, retry-1)
       return
   else:
-      subprocess.run(args=args, check=True, timeout=_CMD_TIMEOUT)
+    subprocess.run(args=args, check=True, timeout=_CMD_TIMEOUT)
 
   args = ["sdkmanager", "--update"]
   logging.info("Update all installed packages: %s", " ".join(args))
-  subprocess.run(args=args, check=False)
-
+  if retry > 1:
+    try:
+      subprocess.run(args=args, check=False, timeout=_CMD_TIMEOUT)
+    except:
+      _setup_android(platform_version, build_tool_version, sdk_id, retry-1)
+      return
+  else:
+    subprocess.run(args=args, check=False, timeout=_CMD_TIMEOUT)
 
 def _shutdown_emulator():
   command = "adb devices | grep emulator | cut -f1 | while read line; do adb -s $line emu kill; done"
--- a/scripts/gha/test_simulator.py
+++ b/scripts/gha/test_simulator.py
@@ -657,7 +657,16 @@ def _install_android_gameloop_app(gameloop_project, retry=1):
     subprocess.run(args=args, check=False, timeout=_CMD_TIMEOUT)
   args = ["./gradlew", "clean"]
   logging.info("Clean game-loop cache: %s", " ".join(args))
-  subprocess.run(args=args, check=False)
+  if retry > 1:
+    try:
+      subprocess.run(args=args, check=False, timeout=_CMD_TIMEOUT)
+    except:
+      logging.info("Clean game-loop cache timeout, Reset Emualtor now... Remaining retry: %s", retry-1)
+      _reset_emulator_on_error()
+      _install_android_gameloop_app(gameloop_project, retry-1)
+      return
+  else:
+    subprocess.run(args=args, check=False, timeout=_CMD_TIMEOUT)
   args = ["./gradlew", "installDebug", "installDebugAndroidTest"]
   logging.info("Installing game-loop app and test: %s", " ".join(args))
   if retry > 1:
--- a/scripts/gha/test_simulator.py
+++ b/scripts/gha/test_simulator.py
@@ -549,8 +549,8 @@ def _create_and_boot_emulator(sdk_id):
   logging.info("Wait for emulator to boot: %s", " ".join(args))
   subprocess.run(args=args, check=True)
   if FLAGS.ci: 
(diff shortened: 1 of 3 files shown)


scripts/gha/test_simulator.py
```diff
@@ -595,30 +621,76 @@ def _run_android_gameloop_test(package_name, app_path, gameloop_project, retry=1
...
 
-def _install_android_app(app_path):
+def _install_android_app(app_path, retry=1):
   """Install integration_test app into the emulator."""
   args = ["adb", "install", app_path]
   logging.info("Install testapp: %s", " ".join(args))
-  subprocess.run(args=args, check=False)
+  if retry > 1:
+    try:
+      subprocess.run(args=args, check=True, timeout=_CMD_TIMEOUT)
```
jonsimantov, line 630: I don't really love this construction - repeating the subprocess.run on multiple lines. Is there a cleaner way to do this?

(also, like I said on iOS, WDYT about a more generic "run with retry" function?)

sunmou99: I don't like this construction too. Just, for different retry, we need to run different cleaning functions. Do you have a better idea?

sunmou99: What about have a set of retry functions:
run_with_retry // retry on except
run_with_retry_reboot // retry on except, reboot before retry
run_with_retry_wipe_reboot
run_test_with_retry // retry on log error info

jonsimantov: Hmm, are they similar enough to have a single function with a parameter that controls its behavior? (or maybe one for reboot and one for not reboot)

sunmou99: All those are some tiny differences. For those in common, I created the func `_reset_emulator_on_error()`.
And, yeah. I think I could add  parameters (more than one) to control the behavior.

scripts/gha/test_simulator.py
```diff
@@ -59,12 +59,12 @@
 Device Information is stored in TEST_DEVICES in print_matrix_configuration.py
 Example:
 sdk id "system-images;android-29;google_apis;x86":
-  --android_sdk "system-images;android-29;google_apis;x86" --build_tools_version "28.0.3"
+  --android_sdk "system-images;android-29;google_apis;x86" --build_tools_version "29.0.2"
```
chkuang-g, line 62: Is there a need to update the version here since it is just an example?

Similarly to line 67

scripts/gha/test_simulator.py
```diff
@@ -549,27 +557,43 @@ def _create_and_boot_emulator(sdk_id):
...
+    # wait extra 210 seconds to ensure emulator fully booted.
+    time.sleep(210)
   else:
     time.sleep(45)
 
+def _run_with_retry(args, shell=False, check=True, timeout=_CMD_TIMEOUT, retry_time=_TEST_RETRY, device=_DEVICE_NONE, type=_RESET_TYPE_REBOOT):
+  if retry_time > 1:
+    try:
+      subprocess.run(args, shell=shell, check=check, timeout=timeout)
+    except:
```
chkuang-g, line 569: We should log the exception here so that we know what is triggering the retry.

scripts/gha/test_simulator.py
```diff
@@ -549,27 +557,43 @@ def _create_and_boot_emulator(sdk_id):
...
   subprocess.run(args=args, check=True)
   if FLAGS.ci: 
-    # wait extra 90 seconds to ensure emulator fully booted.
-    time.sleep(180)
+    # wait extra 210 seconds to ensure emulator fully booted.
+    time.sleep(210)
   else:
     time.sleep(45)
 
+def _run_with_retry(args, shell=False, check=True, timeout=_CMD_TIMEOUT, retry_time=_TEST_RETRY, device=_DEVICE_NONE, type=_RESET_TYPE_REBOOT):
```
chkuang-g, line 565: We should log when all the retry attempts failed.

scripts/gha/test_simulator.py
```diff
@@ -501,19 +509,19 @@ def _setup_android(platform_version, build_tool_version, sdk_id):
...
 
   args = ["sdkmanager", sdk_id]
   logging.info("Download an emulator: %s", " ".join(args))
-  subprocess.run(args=args, check=True)
+  _run_with_retry(args)
 
   args = ["sdkmanager", "--update"]
   logging.info("Update all installed packages: %s", " ".join(args))
-  subprocess.run(args=args, check=True)
+  _run_with_retry(args, check=False)
```
chkuang-g, line 524: Why changing check from True to False?

scripts/gha/test_simulator.py
```diff
@@ -129,7 +137,7 @@
     "android_sdk", "system-images;android-29;google_apis;x86",
     "See module docstring for details on how to set and get this id.")
 flags.DEFINE_string(
-    "build_tools_version", "28.0.3",
+    "build_tools_version", "29.0.2",
```
chkuang-g, line 140: Isn't this value in `build.gradle`? Is this the right place to update the config value like this?

scripts/gha/test_simulator.py
sunmou99: No need, but I feel there is not harm too.

scripts/gha/test_simulator.py
sunmou99: This is a non-critical step.
`check=True` only when this step block next steps.

scripts/gha/test_simulator.py
sunmou99: We are using tools under `$ANDROID_HOME/build-tools/${build_tools_version}`

scripts/gha/test_simulator.py
sunmou99: `subprocess.run` itself print out logs and exceptions.
log example: https://github.com/firebase/firebase-cpp-sdk/runs/4145793172?check_suite_focus=true#step:9:78

scripts/gha/test_simulator.py
sunmou99: Thanks for the suggestion. I just added log in another PR (due to branch dependency): https://github.com/firebase/firebase-cpp-sdk/pull/733/files#diff-51ade566b474eb1faa5dfc7cfdcc3d943f850740c053a79ba0e4faa9f6f89315R679

chkuang-g (approved):

Issue #737, opened 2021-11-05
[Bug] Crash in AuthResponse::error_code


mcasper:
### [REQUIRED] Please fill in the following fields:

  * Pre-built SDK from the [website](https://firebase.google.com/download/cpp) or open-source from this repo: open-source from this repo
  * Firebase C++ SDK version: v8.6.0
  * Problematic Firebase Component: Auth
  * Other Firebase Components in use: Database
  * Platform you are using the C++ SDK on: Mac
  * Platform you are targeting: Desktop

### [REQUIRED] Please describe the issue here:

If an AuthRequest is cancelled or times out, calling `error_code()` on it causes a crash.

Example stacktraces:

```
Crashed: Thread
0  Tuple                          0xef351e firebase::auth::AuthResponse::error_code() const + 2608 (memory:2608)
1  Tuple                          0xed6999 firebase::auth::AuthenticationResult firebase::auth::AuthenticationResult::FromResponse<firebase::auth::VerifyCustomTokenResponse>(firebase::auth::VerifyCustomTokenResponse const&) + 39 (authentication_result.h:39)
2  Tuple                          0xed609a firebase::auth::AuthenticationResult firebase::auth::CompleteSignInFlow<firebase::auth::VerifyCustomTokenResponse>(firebase::auth::AuthData*, firebase::auth::VerifyCustomTokenResponse const&) + 62 (authentication_result.h:62)
3  Tuple                          0xec7ca4 void firebase::auth::PerformSignInFlow<firebase::auth::VerifyCustomTokenResponse, firebase::auth::User*, firebase::auth::VerifyCustomTokenRequest>(firebase::auth::AuthDataHandle<firebase::auth::User*, firebase::auth::VerifyCustomTokenRequest>*) + 62 (authentication_result.h:62)
[...]

Issue #615, opened 2021-08-24
[Bug] Refresh token


awemvfx:
### [REQUIRED] Please fill in the following fields:

  * Pre-built SDK open-source from this repo
  * Firebase C++ SDK version: 8.3.0
  * Problematic Firebase Component: Auth
  * Other Firebase Components in use: App, Firestore
  * Platform you are using the C++ SDK on: Mac
  * Platform you are targeting: desktop

### [REQUIRED] Please describe the issue here:

App start normally. User is logged in and watching his collections. After few hours some users had crash on refresh token.

```Crashed Thread: 18606

Application Specific Information:
Fatal Error: EXC_BAD_ACCESS / KERN_INVALID_ADDRESS

Thread 18606 Crashed:
0   mInstaller                      0x105439db5         bool
1   mInstaller                      0x105439c54         firebase::auth::AuthResponse::error_code
2   mInstaller                      0x1054033c9         firebase::auth::(anonymous namespace)::EnsureFreshToken
3   mInstaller                      0x1054030d6         firebase::auth::(anonymous namespace)::EnsureFreshToken
4   mInstaller                      0x105402fc4         firebase::auth::User::GetTokenInternal::lambda::operator()
5   mInstaller                      0x105402f94         firebase::auth::User::GetTokenInternal::lambda::__invoke
6   mInstaller                      0x10541c706         firebase::auth::CallAsync<T>::lambda::operator()
7   mInstaller                      0x10541c6b4         firebase::auth::CallAsync<T>::lambda::__invoke
8   mInstaller                      0x10541cc76         firebase::callback::CallbackVariadic<T>::RunInternal<T>
9   mInstaller                      0x10541cb34         firebase::callback::CallbackVariadic<T>::Run
10  mInstaller                      0x10528a6b7         firebase::scheduler::Scheduler::TriggerCallback
11  mInstaller                      0x10528a20a         firebase::scheduler::Scheduler::WorkerThreadRoutine
12  mInstaller                      0x10528e621         std::__1::__invoke<T>
13  mInstaller                      0x10528e56d         std::__1::__thread_execute<T>
14  mInstaller                      0x10528dd7a         std::__1::__thread_proxy<T>
15  libsystem_pthread.dylib         0x7fff203c58fb      _pthread_start
16  libsystem_pthread.dylib         0x7fff203c1442      thread_start

Thread 17239
0   libsystem_kernel.dylib          0x7fff203902ba      mach_msg_trap
[...]

PR #738, opened 2021-11-05
Override ResponseJson::MarkFailed to set application_data_.

jonsimantov:
Setting application_data_ to an empty flatbuffer will prevent application_data_ from being nullptr. This was causing an issue in #737 when AuthResponse::error_code and AuthResponse::error_message were accessing a null application_data_ if the response failed.

***

Integration tests to be run.
***

### Type of Change
***

#### Notes

### DIFF
--- a/release_build_files/readme.md
+++ b/release_build_files/readme.md
@@ -567,6 +567,12 @@ workflow use only during the development of your app, not for publicly shipping
 code.
 
 ## Release Notes
+### Next Release
+-   Changes
+    -   Auth (Desktop): Fixed a crash in `error_code()` when a request
+        is cancelled or times out.
+	([#737](https://github.com/firebase/firebase-cpp-sdk/issues/737))
+
 ### 8.7.0
 -   Changes
     -   Firestore: Released to general availability for Android and iOS (desktop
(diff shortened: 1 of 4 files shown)


dconeybe:
FWIW I don't think that this is the right fix. I think the right fix is to override `MarkFailed()` in `ResponseJson`. I considered implementing the same fix (check for application_data_ being empty) in #615 but ended up fixing a lower-level problem in `ResponseJson::MarkCompleted()`.

jonsimantov:
> FWIW I don't think that this is the right fix. I think the right fix is to override `MarkFailed()` in `ResponseJson`. I considered implementing the same fix (check for application_data_ being empty) in #615 but ended up fixing a lower-level problem in `ResponseJson::MarkCompleted()`.

What should MarkFailed put in application_data_, just an empty flatbuffer?

dconeybe (approved):
LGTM. Just needs a release notes entry.

dconeybe (approved):

release_build_files/readme.md
```diff
@@ -666,7 +672,7 @@ code.
         ([#429](https://github.com/firebase/firebase-cpp-sdk/pull/429)).
     -   AdMob (iOS): Temporarily pinned AdMob dependency to a special version of the
         Google-Mobile-Ads-SDK Cocoapod, "7.69.0-cppsdk", to maintain compatibility
-	with version 8.x of the Firebase iOS SDK.
+        with version 8.x of the Firebase iOS SDK.
```
jonsimantov, line 675: Untabified the file.

PR #739, opened 2021-11-05
Fix crash in Messaging on Android when Terminate is called while tasks are still pending.

TBarendt:
When firebase::messaging::Terminate is called while there are still pending tasks left a null pointer exception is causing an ANR on Andriod.

FutureData is destroyed before calling util::Terminate which in turn will try to cancel all pending callbacks. This will call cancel() in JniResultCallback.java which will lock a mutex and callback into C++ using the JniResultCallback_nativeResult function. Depending on the task canceled it will call into different callbacks in messaging.cc and some of them as for example CompleteStringCallback are using FutureData which has already been destroyed. This will cause the thread to stop and the mutex in JniResultCallbak.java to remain locked. The next message that either completes or fails will now end up in a dead lock causing an ANR.

The proposed fix is to not destroy FutureData until after the tasks have been canceled.

***

***

### Type of Change
***

#### Notes

### DIFF
--- a/messaging/src/android/cpp/messaging.cc
+++ b/messaging/src/android/cpp/messaging.cc
@@ -704,8 +704,8 @@ void Terminate() {
   g_firebase_messaging = nullptr;
   SetListener(nullptr);
   ReleaseClasses(env);
-  FutureData::Destroy();
   util::Terminate(env);
+  FutureData::Destroy();
 }
 
 // Start a service which will communicate with the Firebase Cloud Messaging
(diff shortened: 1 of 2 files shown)


kevinsquid (dismissed):
Resolves a common ANR event in firebase when shutting down Messaging on Android

jonsimantov:
Nice find! I kicked off an integration test run and will approve the PR after that finishes.

Because we currently have some flaky tests, don't worry if the integration test fails - I'll manually remove the failure tag as long as the failure isn't caused by this change.

Thanks for your contribution!

jonsimantov (approved):

PR #741, opened 2021-11-09
RewardedAds for Android

DellaBitta:
- Implemented RewardedAd on Android, and created stubs for iOS and Desktop builds.
- Revamped C++ JNI function signatures to have much shorter names, and to reuse functions common to interstitial and rewarded ads.
- Created a new internal base class for full screen event listeners to share its implementation across RewardedAds and InsterstitialAds.  This allows us to reuse some JNI callback functions.
- Grouped those integration tests which require user interaction so that they execute one after the next, and front load them to the start of the itest flow (after a few quick sanity checks).  This should make running the manual tests easier.

***
### Testing
**[iTest Run](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1467851369)**
Expanded AdMob integration tests.
Executed tests manually on Android  phones.
Validated (via logs) that the SetServerSideVerificationOption strings are properly reaching the Android SDK space.
***

### Type of Change
***

### DIFF
--- a/admob/integration_test/src/integration_test.cc
+++ b/admob/integration_test/src/integration_test.cc
@@ -937,6 +937,9 @@ TEST_F(FirebaseAdMobTest, TestBannerView) {
             bounding_box_listener.bounding_box_changes_.size());
 
 #if defined(ANDROID) || TARGET_OS_IPHONE
+  // Simulators have varying aspect ratios.
+  TEST_REQUIRES_USER_INTERACTION;
+
   // As an extra check, all bounding boxes except the last should have the same
   // size aspect ratio that we requested. For example if you requested a 320x50
   // banner, you can get one with the size 960x150. Use EXPECT_NEAR because the
--- a/admob/integration_test/src/integration_test.cc
+++ b/admob/integration_test/src/integration_test.cc
@@ -1233,6 +1233,7 @@ TEST_F(FirebaseAdMobTest, TestRewardedAdErrorNotInitialized) {
 
 TEST_F(FirebaseAdMobTest, TesRewardedAdErrorAlreadyInitialized) {
   SKIP_TEST_ON_DESKTOP;
+  SKIP_TEST_ON_IOS;
 
   {
     firebase::admob::RewardedAd* rewarded = new firebase::admob::RewardedAd();
--- a/admob/integration_test/src/integration_test.cc
+++ b/admob/integration_test/src/integration_test.cc
@@ -607,8 +607,8 @@ TEST_F(FirebaseAdMobTest, TestRewardedAdLoad) {
 // Interactive test section.  These have been placed up front so that the
 // tester doesn't get bored waiting for them.
 TEST_F(FirebaseAdMobTest, TestBannerViewAdOpenedAdClosed) {
-  SKIP_TEST_ON_DESKTOP;
   TEST_REQUIRES_USER_INTERACTION;
+  SKIP_TEST_ON_DESKTOP;
 
   const firebase::admob::AdSize banner_ad_size(kBannerWidth, kBannerHeight);
   firebase::admob::BannerView* banner = new firebase::admob::BannerView();
@@ -755,8 +755,8 @@ TEST_F(FirebaseAdMobTest, TestRewardedAdLoadAndShow) {
   WaitForCompletion(rewarded->Show(&user_earned_reward_listener), "Show");
 
   LogDebug(
-      "Click the Install button in the ad, "
-      "return to the ad, and close the ad to continue");
+      "Wait for the Ad to finish playing, click the ad, return to the ad, "
+      "then close the ad to continue.");
 
   while (
       full_screen_content_listener.num_on_ad_dismissed_full_screen_content_ ==
@@ -785,8 +785,6 @@ TEST_F(FirebaseAdMobTest, TestRewardedAdLoadAndShow) {
 // Other Banner View Tests
 
 TEST_F(FirebaseAdMobTest, TestBannerView) {
-  // AdMob cannot be tested on Firebase Test Lab, so disable tests on FTL.
-  TEST_REQUIRES_USER_INTERACTION;
   SKIP_TEST_ON_DESKTOP;
 
   const firebase::admob::AdSize banner_ad_size(kBannerWidth, kBannerHeight);
@@ -1406,8 +1404,6 @@ TEST_F(FirebaseAdMobTest, TestRewardedAdErrorBadExtrasClassName) {
 
 // Stress tests.  These take a while so run them near the end.
 TEST_F(FirebaseAdMobTest, TestBannerViewStress) {
-  // AdMob cannot be tested on Firebase Test Lab, so disable tests on FTL.
-  TEST_REQUIRES_USER_INTERACTION;
   SKIP_TEST_ON_DESKTOP;
 
   for (int i = 0; i < 10; ++i) {
@@ -1426,7 +1422,6 @@ TEST_F(FirebaseAdMobTest, TestBannerViewStress) {
 }
 
 TEST_F(FirebaseAdMobTest, TestInterstitialAdStress) {
-  TEST_REQUIRES_USER_INTERACTION;
   SKIP_TEST_ON_DESKTOP;
 
   for (int i = 0; i < 10; ++i) {
@@ -1446,7 +1441,6 @@ TEST_F(FirebaseAdMobTest, TestInterstitialAdStress) {
 }
 
 TEST_F(FirebaseAdMobTest, TestRewardedAdStress) {
-  TEST_REQUIRES_USER_INTERACTION;
   SKIP_TEST_ON_DESKTOP;
   SKIP_TEST_ON_IOS;
 
--- a/admob/integration_test/src/integration_test.cc
+++ b/admob/integration_test/src/integration_test.cc
@@ -1405,6 +1405,7 @@ TEST_F(FirebaseAdMobTest, TestRewardedAdErrorBadExtrasClassName) {
 
 // Stress tests.  These take a while so run them near the end.
 TEST_F(FirebaseAdMobTest, TestBannerViewStress) {
+  TEST_REQUIRES_USER_INTERACTION;
   SKIP_TEST_ON_DESKTOP;
 
   for (int i = 0; i < 10; ++i) {
@@ -1423,6 +1424,7 @@ TEST_F(FirebaseAdMobTest, TestBannerViewStress) {
 }
 
 TEST_F(FirebaseAdMobTest, TestInterstitialAdStress) {
+  TEST_REQUIRES_USER_INTERACTION;
   SKIP_TEST_ON_DESKTOP;
 
   for (int i = 0; i < 10; ++i) {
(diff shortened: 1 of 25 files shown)


admob/integration_test/src/integration_test.cc
```diff
@@ -489,6 +506,229 @@ TEST_F(FirebaseAdMobTest, TestRequestConfigurationSetGet) {
                          retrieved_configuration.test_device_ids.end(), "3"));
 }
 
+// Simple Load Tests as a sanity check. These don't show the ad, just
```
DellaBitta, line 509: Moved some BannerView and Interstitial tests around.  Will call out the new tests via comments below. so that you needed waste your time reviewing unchanged tests.

admob/integration_test/src/integration_test.cc
```diff
@@ -489,6 +506,229 @@ TEST_F(FirebaseAdMobTest, TestRequestConfigurationSetGet) {
...
+                    "Initialize");
+
+  // When the InterstitialAd is initialized, load an ad.
+  firebase::admob::AdRequest request = GetAdRequest();
+  WaitForCompletion(interstitial->LoadAd(kInterstitialAdUnit, request),
+                    "LoadAd");
+  delete interstitial;
+}
+
+TEST_F(FirebaseAdMobTest, TestRewardedAdLoad) {
```
DellaBitta, line 542: New test.

admob/integration_test/src/integration_test.cc
```diff
@@ -489,6 +506,229 @@ TEST_F(FirebaseAdMobTest, TestRequestConfigurationSetGet) {
...
+  EXPECT_EQ(
+      full_screen_content_listener.num_on_ad_dismissed_full_screen_content_, 1);
+
+  interstitial->SetFullScreenContentListener(nullptr);
+  interstitial->SetPaidEventListener(nullptr);
+
+  delete interstitial;
+}
+
+TEST_F(FirebaseAdMobTest, TestRewardedAdLoadAndShow) {
```
DellaBitta, line 679: New test.

admob/integration_test/src/integration_test.cc
```diff
@@ -1148,6 +1214,200 @@ TEST_F(FirebaseAdMobTest, TestInterstitialAdErrorBadExtrasClassName) {
   delete interstitial_ad;
 }
 
+// Other RewardedAd Tests.
+
```
DellaBitta, line 1218: New tests from here downwards.

admob/src/common/interstitial_ad_internal.cc
```diff
@@ -53,61 +50,6 @@ InterstitialAdInternal* InterstitialAdInternal::CreateInstance(
         // FIREBASE_PLATFORM_TVOS
 }
 
-void InterstitialAdInternal::SetFullScreenContentListener(
```
DellaBitta, line 56: Moved to a new internal full screen event listener base class which is shared with RewardedAds.

admob/src/ios/rewarded_ad_internal_ios.h
```diff
@@ -0,0 +1,70 @@
+/*
```
DellaBitta, line 1: Stub implementation for now.

admob/integration_test/src/integration_test.cc
```diff
@@ -536,6 +553,237 @@ TEST_F(FirebaseAdMobTest, TestRequestConfigurationSetGet) {
...
+  // these values properly (there are no get methods). At least invoke the
+  // method to ensure least we can set them without any exceptions occurring.
+  options.custom_data = "custom data";
+  options.user_id = "123456";
+  rewarded->SetServerSideVerificationOptions(options);
+
+  TestUserEarnedRewardListener user_earned_reward_listener;
+  WaitForCompletion(rewarded->Show(&user_earned_reward_listener), "Show");
+
+  LogDebug(
```
jonsimantov, line 757: I think you get credit for the ad even if you just watch it through, you don't have to install the thing.

admob/integration_test/src/integration_test.cc
```diff
@@ -1195,6 +1269,201 @@ TEST_F(FirebaseAdMobTest, TestInterstitialAdErrorBadExtrasClassName) {
...
+  request.add_extra(kAdNetworkExtrasInvalidClassName, "shouldnot", "work");
+  WaitForCompletion(rewarded->LoadAd(kRewardedAdUnit, request), "LoadAd",
+                    firebase::admob::kAdMobErrorAdNetworkClassLoadError);
+  delete rewarded;
+}
+
+// Stress tests.  These take a while so run them near the end.
+TEST_F(FirebaseAdMobTest, TestBannerViewStress) {
+  // AdMob cannot be tested on Firebase Test Lab, so disable tests on FTL.
+  TEST_REQUIRES_USER_INTERACTION;
```
jonsimantov, line 1410: > AdMob cannot be tested on Firebase Test Lab, so disable tests on FTL.

What's the deal with this?

Also, even if we can't run this test on FTL, it would be better here to just ensure that you are specifically not running on FTL - the automated tests should still run on simulator, no?

Let's talk to @sunmou99 about seeing if there's a way to add a SKIP_TEST_ON_FIREBASE_TEST_LAB macro.

admob/src/android/admob_android.cc
```diff
@@ -980,39 +876,90 @@ Java_com_google_firebase_admob_internal_cpp_InterstitialAdHelper_notifyAdFailedT
   // Invoke AdMobInternal, a friend of AdResult, to have it access its
   // protected constructor with the AdError data.
   const AdResult& ad_result = AdMobInternal::CreateAdResult(ad_result_internal);
-  internal->NotifyListenerOfAdFailedToShowFullScreenContent(ad_result);
+  listener->NotifyListenerOfAdFailedToShowFullScreenContent(ad_result);
 }
 
-extern "C" JNIEXPORT void JNICALL
-Java_com_google_firebase_admob_internal_cpp_InterstitialAdHelper_notifyAdImpressionEvent(  // NOLINT
+extern "C" JNIEXPORT void JNICALL JNI_notifyAdImpressionEvent(JNIEnv* env,
```
jonsimantov, line 882: I believe you don't need any of the "extern "C" JNIEXPORT JNICALL stuff since we are registering natives directly - also it would probably be best to put these in an unnamed namespace. (along with the one I added!)

admob/src/common/full_screen_ad_event_listener.cc
```diff
@@ -0,0 +1,83 @@
...
+void FullScreenAdEventListener::SetPaidEventListener(
+    PaidEventListener* listener) {
+  MutexLock lock(listener_mutex_);
+  paid_event_listener_ = listener;
+}
+
+void FullScreenAdEventListener::NotifyListenerOfAdClickedFullScreenContent() {
+  MutexLock lock(listener_mutex_);
+  if (full_screen_content_listener_ != nullptr) {
+    full_screen_content_listener_->OnAdClicked();
```
jonsimantov, line 40: Just a question: is it safe to keep the mutex locked the whole time the user's callback is running? This could potentially be a long time. Typically if there's any way to avoid locking a mutex while a user callback runs we try not to lock it. Not really sure how you'd safely do this here though, to avoid the listener itself being deleted, so maybe never mind,.

admob/src/common/rewarded_ad.cc
```diff
@@ -0,0 +1,117 @@
...
+
+const char kUninitializedError[] =
+    "Initialize() must be called before this method.";
+
+RewardedAd::RewardedAd() {
+  FIREBASE_ASSERT(admob::IsInitialized());
+  internal_ = internal::RewardedAdInternal::CreateInstance(this);
+
+  GetOrCreateCleanupNotifier()->RegisterObject(this, [](void* object) {
+    FIREBASE_ASSERT_MESSAGE(
```
jonsimantov, line 35: The idea of CleanupNotifier is to invalidate the objects and WARN the user they did the wrong thing, not assert.

admob/src_java/com/google/firebase/admob/internal/cpp/RewardedAdHelper.java
```diff
@@ -0,0 +1,371 @@
...
+
+  // Pointer to the RewardedAdInternalAndroid object that created this
+  // object.
+  private long mRewardedAdInternalPtr;
+
+  // The GMA SDK {@link RewardedAd} associated with this helper.
+  private RewardedAd mRewarded;
+
+  // Synchronization object for thread safe access to:
+  // * mRewarded
```
jonsimantov, line 53: It doesn't actually seem to be locking when mRewarded is used, is this because mRewarded is only ever used in the UI thread? If so you can mention that

jonsimantov (approved):

admob/integration_test/src/integration_test.cc
DellaBitta: I've enabled this test successfully in testing another PR, so I'll just remove it here.

I think that this TEST_REQUIRES_USER_INTERACTION was originally for the main BannerView tests that required you to click things but I was confused by the comment that was there, which seems to make it sound like the tests don't run at all.  I hadn't tested this until recentlly and it works fine.

admob/src/common/full_screen_ad_event_listener.cc
DellaBitta: Yes, I see you're point.

I feel like tackling it this way:  If this mutex lock is a problem then we'll have safe customer reports saying that it's in their way.  But if we take the mutex out there, then we _could_ have crazy crash reports due to corrupted memory somewhere later in the runtime. For now I'd like to play it safe and open up to dangers later, but only if we need to.  WDYT?

admob/src/common/rewarded_ad.cc
DellaBitta: Noted. This was from the older implementation. Will fix it in banner and interstitial, too.

admob/src_java/com/google/firebase/admob/internal/cpp/RewardedAdHelper.java
DellaBitta: Adding it for other operations...

admob/integration_test/src/integration_test.cc
DellaBitta: While you get the reward the ad doesn't produce the ad clicked event that we're also looking for.

DellaBitta: Done.

admob/src_java/com/google/firebase/admob/internal/cpp/RewardedAdHelper.java
DellaBitta: Done.

admob/integration_test/src/integration_test.cc
DellaBitta: Done.

admob/src/android/admob_android.cc
DellaBitta: Done.

admob/src/common/rewarded_ad.cc
DellaBitta: Done.

jonsimantov (approved):

admob/integration_test/src/integration_test.cc
```diff
@@ -937,6 +937,9 @@ TEST_F(FirebaseAdMobTest, TestBannerView) {
             bounding_box_listener.bounding_box_changes_.size());
 
 #if defined(ANDROID) || TARGET_OS_IPHONE
```
jonsimantov, line 939: Instead of disabling this test, is it possible to just add some wiggle room to the aspect ratio check on simulator?

PR #744, opened 2021-11-10
Guard against Banner/Interstitial use of nullptr callbacks

DellaBitta:
Check to see if Interstitial and BannerView callback pointers have been nullified before invoking them in Java.

***

Ran iTests on Android phone.
**[iTest Run](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1446187252)**
***

### Type of Change
***


admob/src_java/com/google/firebase/admob/internal/cpp/BannerViewHelper.java
```diff
@@ -177,7 +177,6 @@ public void destroy(final long callbackDataPtr) {
     completeBannerViewFutureCallback(callbackDataPtr,
         ConstantsHelper.CALLBACK_ERROR_NONE,
         ConstantsHelper.CALLBACK_ERROR_MESSAGE_NONE);
-    mNotifyBoundingBoxListenerOnNextDraw.set(true);
```
DellaBitta, line 180: This wouldn't have done anything, anyway, as the mAdView was already set to null.  Also there's a synchronous bounding box changed event immediately prior to this.

sunmou99 (commented):
Do we need to add some logs or throw errors when it's a nullptr?

DellaBitta:
> Do we need to add some logs or throw errors when it's a nullptr?

I don't think so.  The most common use case is launching an ad operation and then destroying C++ object before a callback occurs.  

Or more commonly, start the process of destroying the C++ object when suddenly a listener is invoked from Android and iOS about movement, clicks, etc.

sunmou99 (commented):
Check no pointer erros. LGTM.

sunmou99 (approved):

PR #762, opened 2021-11-29
Feature/admob 2021 stability in future usage

DellaBitta:
Fixes stability issues regarding race conditions in the AdMob SDK:
 - Ensure that future objects are fully created before launching async operations which use their handles.
 - Remove the blocking destructor of BannerView on Android. 
   - There was a race condition where the the future might be destroyed before the destroy method returns from Android causing an assertion in the Future implementation.
 - Remove the blocking destructors of Interstitial and Rewarded Ads on iOS.
   - These implementations don't invoke AdMob objects and so don't need to be run on the main thread.
 
***

FTL testing.  Local Testing. [Integration CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1516389819)
***

### Type of Change
***

### DIFF
--- a/admob/integration_test/src/integration_test.cc
+++ b/admob/integration_test/src/integration_test.cc
@@ -795,6 +795,8 @@ TEST_F(FirebaseAdMobTest, TestBannerView) {
   // Set the listener.
   TestBoundingBoxListener bounding_box_listener;
   banner->SetBoundingBoxListener(&bounding_box_listener);
+  PauseForVisualInspectionAndCallbacks();
+
   int expected_num_bounding_box_changes = 0;
   EXPECT_EQ(expected_num_bounding_box_changes,
             bounding_box_listener.bounding_box_changes_.size());
--- a/admob/src/android/banner_view_internal_android.cc
+++ b/admob/src/android/banner_view_internal_android.cc
@@ -150,7 +150,7 @@ BannerViewInternalAndroid::~BannerViewInternalAndroid() {
   // we can prevent leaking memory.
   env->CallVoidMethod(
       helper_, banner_view_helper::GetMethodId(banner_view_helper::kDestroy),
-      nullptr);
+      /*callbackDataPtr=*/nullptr, /*destructor_invocation=*/true);
 
   env->DeleteGlobalRef(ad_view_);
   ad_view_ = nullptr;
@@ -386,7 +386,14 @@ Future<void> BannerViewInternalAndroid::Resume() {
 
 Future<void> BannerViewInternalAndroid::Destroy() {
   firebase::MutexLock lock(mutex_);
-  return InvokeNullary(kBannerViewFnDestroy, banner_view_helper::kDestroy);
+  FutureCallbackData<void>* callback_data =
+      CreateVoidFutureCallbackData(kBannerViewFnDestroy, &future_data_);
+  Future<void> future =
+      MakeFuture(&future_data_.future_impl, callback_data->future_handle);
+  ::firebase::admob::GetJNI()->CallVoidMethod(
+      helper_, banner_view_helper::GetMethodId(banner_view_helper::kDestroy),
+      reinterpret_cast<jlong>(callback_data), /*destructor_invocation=*/false);
+  return future;
 }
 
 Future<void> BannerViewInternalAndroid::SetPosition(int x, int y) {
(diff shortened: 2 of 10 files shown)


admob/integration_test/src/integration_test.cc
```diff
@@ -795,6 +795,8 @@ TEST_F(FirebaseAdMobTest, TestBannerView) {
   // Set the listener.
   TestBoundingBoxListener bounding_box_listener;
   banner->SetBoundingBoxListener(&bounding_box_listener);
+  PauseForVisualInspectionAndCallbacks();
```
DellaBitta, line 798: Wait to ensure that any callback attempts have time to complete before we start testing the listener.

admob/src/android/banner_view_internal_android.cc
```diff
@@ -390,21 +386,29 @@ Future<void> BannerViewInternalAndroid::Resume() {
 
 Future<void> BannerViewInternalAndroid::Destroy() {
   firebase::MutexLock lock(mutex_);
-  return InvokeNullary(kBannerViewFnDestroy, banner_view_helper::kDestroy);
+  FutureCallbackData<void>* callback_data =
```
DellaBitta, line 389: No longer use futures to track the destruction as they may be destroyed before the're completed. This had caused a crash on some Android runs.

TBH, the Applicatoin _should already have invoked_ `Destroy()` prior to this object's destruction.  Invoking it again here isn't required.  I guess it was put here to protect our users from accidentally leaking memory. 

There doesn't seem to be a good reason to block the destructor though.  Instead I'll start an asynchronous attempt to destroy the Java resources, if they haven't been destroyed already, and then return immediately here.

admob/src/common/admob_common.cc
```diff
@@ -227,7 +227,9 @@ const char* GetRequestAgentString() {
 // Mark a future as complete.
 void CompleteFuture(int error, const char* error_msg,
                     SafeFutureHandle<void> handle, FutureData* future_data) {
-  future_data->future_impl.Complete(handle, error, error_msg);
+  if (future_data->future_impl.ValidFuture(handle)) {
+    future_data->future_impl.Complete(handle, error, error_msg);
```
DellaBitta, line 231: Just an extra safety check.

admob/src/ios/interstitial_ad_internal_ios.mm
```diff
@@ -32,29 +32,20 @@
 
 InterstitialAdInternalIOS::~InterstitialAdInternalIOS() {
   firebase::MutexLock lock(mutex_);
-  // Clean up any resources created in InterstitialAdInternalIOS.
```
DellaBitta, line 35: This code was waiting to run on the main thread but all it was doing was setting values to null.  While all AdMob operations should be conducted on the main thread, we're not actually performing any AdMob operations here, so let's simplify this code and just assign the null values here.

admob/src/ios/rewarded_ad_internal_ios.mm
```diff
@@ -33,22 +33,13 @@
 RewardedAdInternalIOS::~RewardedAdInternalIOS() {
   firebase::MutexLock lock(mutex_);
   // Clean up any resources created in RewardedAdInternalIOS.
-  Mutex mutex(Mutex::kModeNonRecursive);
```
DellaBitta, line 36: Simplify in the same way that we did for Interstitials, above.

admob/src_java/com/google/firebase/admob/internal/cpp/BannerViewHelper.java
```diff
@@ -142,41 +142,57 @@ public void initialize(Activity activity) {
...
-        mPopUp = null;
+              synchronized (mPopUpLock) {
+                if (mPopUp != null) {
+                  mPopUp.dismiss();
+                  mPopUp = null;
+                }
+              }
+              synchronized (mBannerViewLock) {
+                if (destructor_invocation == false) {
+                  notifyBoundingBoxChanged(mBannerViewInternalPtr);
```
DellaBitta, line 173: Don't invoke the bounding box listener if this destroy() invocation originated from the C++ object destructor.  The bounding box listener won't be there anymore.

admob/src/android/banner_view_internal_android.cc
jonsimantov: Yeah, I think there's no reason to actually wait for the Java destruction to complete.

jonsimantov (approved):

PR #764, opened 2021-11-29
Workaround for Linux x86 build:  downgrade libraries on GitHub runners

jonsimantov:
When installing 32-bit Linux dependencies on GitHub runners, downgrade `libpcre2-8-0` to an earlier version to ensure compatibility with the i386 version of the package. This is something that should be fixed in a subsequent Ubuntu release and so is a temporary workaround.

This PR also adds checks to the various prerequisite commands run by build_desktop.py, which was previously just silently ignoring errors (making this much harder to track down). Now it will error out as soon as a command fails.
***

Running both relevant workflows: desktop (see latest Checks) and [cpp-packaging](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1518617354) to verify the workaround.
***

### Type of Change
***

#### Notes

### DIFF
--- a/scripts/gha/build_desktop.py
+++ b/scripts/gha/build_desktop.py
@@ -75,7 +75,7 @@ def install_x86_support_libraries(gha_build=False):
       # to get around this issue. This will presumably be fixed in a future
       # Ubuntu update.
       packages = (['--allow-downgrades'] + packages +
-                  ['libpcre2-8-0=10.34-7', 'libsecret-1-dev=0.20.4-0ubuntu1'])
+                  ['libpcre2-8-0=10.34-7'])
 
     # First check if these packages exist on the machine already
     devnull = open(os.devnull, "w")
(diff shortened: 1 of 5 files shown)


scripts/gha/build_desktop.py
```diff
@@ -56,12 +56,27 @@ def append_line_to_file(path, line):
...
                 'libssl-dev:i386']
+    if gha_build:
+      # Workaround for GitHub runners, which have an incompatibility between the
+      # 64-bit and 32-bit versions of libpcre2-8-0. Downgrade the installed
+      # 64-bit version of the library to get around this issue. This will
+      # presumably be fixed in a future Ubuntu update. (Also include
+      # libsecret-1-dev (64-bit) here to ensure it stays installed, otherwise
+      # the downgrade can remove it.)
+      packages = ['--allow-downgrades'] + packages + ['libpcre2-8-0=10.34-7',
+                                                      'libsecret-1-dev']
```
a-maurice: Any problem with libsecret being here without the i386, and above with?

a-maurice (approved):

PR #766, opened 2021-11-30
Revert "Trigger workflow move github api cod to github.py"

jonsimantov:
Reverts firebase/firebase-cpp-sdk#746, which causes the cpp-packaging workflow to be unable to trigger integration tests.


jonsimantov:
@sunmou99 - if you don't have time for a forward fix, let's revert this for now and fix it later.

sunmou99:
The cpp-packaging workflow is able to trigger integration tests, except the branch got deleted. This is an excepted behavior.

PR #772, opened 2021-12-01
Functions to Disable AdMob Automatic Services on iOS

DellaBitta:
Add iOS-specific functions to disable some automatic services:
  - `DisableSDKCrashReporting()`
  - `DisableMediationInitialization()`

These functions have no behavior on Android.
Fulfills b/206022469

***
### Testing
[Integration Test CI #1](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1526705401)
[And again](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1541417047) after merging feature branch to this one.
[And again](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1588778170) after adding pre-init testing.
***

### Type of Change
***


app/src/util_ios.mm
```diff
@@ -161,21 +161,20 @@ void ForEachAppDelegateClass(void (^block)(Class)) {
   return array;
 }
 
-void NSArrayOfNSStringToVectorOfString(NSArray* array, std::vector<std::string>* string_vector) {
+void NSArrayOfNSStringToVectorOfString(NSArray *array, std::vector<std::string> *string_vector) {
```
DellaBitta, line 164: Apparently I branched too soon..?  The only changes to these files is their format.

app/src/util_ios.h
```diff
@@ -195,8 +195,8 @@ NSMutableArray *StringVectorToNSMutableArray(
 
 // Convert a NSArray into a vector of strings.  Asserts if a non NSString
 // object is found in the array.
-void NSArrayOfNSStringToVectorOfString(NSArray* array,
-    std::vector<std::string>* string_vector);
+void NSArrayOfNSStringToVectorOfString(NSArray *array,
+                                       std::vector<std::string> *string_vector);
```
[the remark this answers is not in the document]
DellaBitta: `#include <vector>` is indeed at the top of this file.

app/src/util_ios.mm
jonsimantov: clang-format will generally respect the first "pointer gravity" (left or right) it sees in the file, and enforce it throughout.

app/src/util_ios.h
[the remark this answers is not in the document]
jonsimantov: Probably the linter either doesn't know that OBJC will be defined, you could move the c++-specific includes outside of that ifdef. Or just ignore.

app/src/util_ios.mm
DellaBitta: Got it.

app/src/util_ios.h
DellaBitta: Fixed.

admob/integration_test/src/integration_test.cc
```diff
@@ -287,6 +287,14 @@ TEST_F(FirebaseAdMobTest, TestInitializationStatus) {
       << "Expected adapter class '" << kAdMobClassName << "' is not loaded.";
 }
 
+TEST_F(FirebaseAdMobTest, TestDisableSDKCrashReporting) {
+  firebase::admob::DisableSDKCrashReporting();
+}
+
+TEST_F(FirebaseAdMobTest, TestDisableMediationInitialization) {
+  firebase::admob::DisableMediationInitialization();
```
a-maurice, line 295: It seems like these tests aren't really testing the functionality, since it needs to be called before initialization.  I think a comment regarding what this is actually testing (that calling the function won't have issues, and not necessarily the functionality) would be good.

DellaBitta: Done!

jonsimantov: You could also implement a second test fixture class that doesn't automatically initialize AdMob to test what happens when you call these before initialization.

jonsimantov (approved):

PR #784, opened 2021-12-02
In the test report, if test failed on all devices. Mark [All x test devices] 

sunmou99:
### Description
In the report, if test failed on all devices. Use [All test devices] instead of listing each devices
e.g.
[emulator_target, emulator_latest, android_target, android_latest, emulator_32bit]
-> [All 5 android_device]

[emulator_target,  android_target, android_latest, emulator_32bit]
-> [All 2 FTL Devices, emulator_target, emulator_32bit]

***

***

### Type of Change
***

#### Notes

### DIFF
--- a/scripts/gha/summarize_test_results.py
+++ b/scripts/gha/summarize_test_results.py
@@ -361,24 +361,24 @@ def combine_config(config, platform, k):
   else:
     config_value = PARAMETERS["integration_tests"]["matrix"][config_name]
   if len(config_value) > 1 and len(config) == len(config_value):
-    config = ["All %s" % config_name]
+    config = ["All %d %s" % (len(config_value), config_name)]
   elif config_name == "ios_device":
     ftl_devices = set(filter(lambda device: TEST_DEVICES.get(device).get("type") in "real", config_value))
     simulators = set(filter(lambda device: TEST_DEVICES.get(device).get("type") in "virtual", config_value))
     if ftl_devices.issubset(set(config)):
-      config.insert(0, "All FTL Devices")
+      config.insert(0, "All %d FTL Devices" % len(ftl_devices))
       config = [x for x in config if (x not in ftl_devices)]
     elif simulators.issubset(set(config)):
-      config.insert(0, "All Simulators")
+      config.insert(0, "All %d Simulators" % len(simulators))
       config = [x for x in config if (x not in simulators)]
   elif config_name == "android_device":
     ftl_devices = set(filter(lambda device: TEST_DEVICES.get(device).get("type") in "real", config_value))
     emulators = set(filter(lambda device: TEST_DEVICES.get(device).get("type") in "virtual", config_value))
     if ftl_devices.issubset(set(config)):
-      config.insert(0, "All FTL Devices")
+      config.insert(0, "All %d FTL Devices" % len(ftl_devices))
       config = [x for x in config if (x not in ftl_devices)]
     elif emulators.issubset(set(config)):
-      config.insert(0, "All Emulators")
+      config.insert(0, "All %d Emulators" % len(emulators))
       config = [x for x in config if (x not in emulators)]
 
   return config
--- a/scripts/gha/summarize_test_results.py
+++ b/scripts/gha/summarize_test_results.py
@@ -49,6 +49,7 @@ from absl import flags
 from absl import logging
 from print_matrix_configuration import PARAMETERS
 from print_matrix_configuration import BUILD_CONFIGS
+from print_matrix_configuration import TEST_DEVICES
 
 import glob
 import re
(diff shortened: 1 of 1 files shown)


jonsimantov:
Would it be possible to put the count of devices there? It would mean something different to me if it failed on ALL 3 DEVICES vs ALL 1 DEVICES. Maybe phrase that better, too - ALL DEVICES (3) vs ALL DEVICES (1)

sunmou99:
> Would it be possible to put the count of devices there? It would mean something different to me if it failed on ALL 3 DEVICES vs ALL 1 DEVICES. Maybe phrase that better, too - ALL DEVICES (3) vs ALL DEVICES (1)

Count added.
However, we only [All *] when the len(devices) > 1. If there is only one device, then it just use the device name.

jonsimantov:
> > Would it be possible to put the count of devices there? It would mean something different to me if it failed on ALL 3 DEVICES vs ALL 1 DEVICES. Maybe phrase that better, too - ALL DEVICES (3) vs ALL DEVICES (1)
> 
> Count added. However, we only use [All *] when the len(devices) > 1. If there is only one device, then it just use the device name.

It would still be good to know if the ONLY device failed, as opposed to if 1 device out of of many failed.

sunmou99:
> > > Would it be possible to put the count of devices there? It would mean something different to me if it failed on ALL 3 DEVICES vs ALL 1 DEVICES. Maybe phrase that better, too - ALL DEVICES (3) vs ALL DEVICES (1)
> > 
> > 
> > Count added. However, we only use [All *] when the len(devices) > 1. If there is only one device, then it just use the device name.
> 
> It would still be good to know if the ONLY device failed, as opposed to if 1 device out of of many failed.

I see where you're coming from. If we could provide number like "x/y", which means "x" out of "y" configs has errors, it may be even better?
But enabling this feature requires some more effort:
Currently, the [All *] logic read config info from `print_matrix_configuration.py`, thus would work correctly for nightly test.
However, to simplify the logic, we ignores all succeeded configs. As a result, the report may not apply [All *] to **customized** matrix; instead, it just list all the failed configs.

I feel it's fine. Because we don't generate report in **PR comment** for manually triggered customized matrix. And for label trigger, the author knows what he/she triggers.

Please let me know what you think. And I can address your concern in a new PR.

jonsimantov:
> > > > Would it be possible to put the count of devices there? It would mean something different to me if it failed on ALL 3 DEVICES vs ALL 1 DEVICES. Maybe phrase that better, too - ALL DEVICES (3) vs ALL DEVICES (1)
> > > 
> > > 
> > > Count added. However, we only use [All *] when the len(devices) > 1. If there is only one device, then it just use the device name.
> > 
> > 
> > It would still be good to know if the ONLY device failed, as opposed to if 1 device out of of many failed.
> 
> I see where you're coming from. If we could provide number like "x/y", which means "x" out of "y" configs has errors, it may be even better? But enabling this feature requires some more effort: Currently, the [All *] logic read config info from `print_matrix_configuration.py`, thus would work correctly for nightly test. However, to simplify the logic, we ignores all succeeded configs. As a result, the report may not apply [All *] to **customized** matrix; instead, it just list all the failed configs.
> 
> I feel it's fine. Because we don't generate report in **PR comment** for manually triggered customized matrix. And for label trigger, the author knows what he/she triggers.
> 
> Please let me know what you think. And I can address your concern in a new PR.

I like the idea of doing x/y. It would be fine to do that in a followup. So this version will just do ALL DEVICES for 2+ devices, and just name the device for = 1 device?

sunmou99:
> > > > > Would it be possible to put the count of devices there? It would mean something different to me if it failed on ALL 3 DEVICES vs ALL 1 DEVICES. Maybe phrase that better, too - ALL DEVICES (3) vs ALL DEVICES (1)
> > > > 
> > > > 
> > > > Count added. However, we only use [All *] when the len(devices) > 1. If there is only one device, then it just use the device name.
> > > 
> > > 
> > > It would still be good to know if the ONLY device failed, as opposed to if 1 device out of of many failed.
> > 
> > 
> > I see where you're coming from. If we could provide number like "x/y", which means "x" out of "y" configs has errors, it may be even better? But enabling this feature requires some more effort: Currently, the [All *] logic read config info from `print_matrix_configuration.py`, thus would work correctly for nightly test. However, to simplify the logic, we ignores all succeeded configs. As a result, the report may not apply [All *] to **customized** matrix; instead, it just list all the failed configs.
> > I feel it's fine. Because we don't generate report in **PR comment** for manually triggered customized matrix. And for label trigger, the author knows what he/she triggers.
> > Please let me know what you think. And I can address your concern in a new PR.
> 
> I like the idea of doing x/y. It would be fine to do that in a followup. So this version will just do ALL DEVICES for 2+ devices, and just name the device for = 1 device?

Sure, I will apply this x/y to all configs in a followup PR.

And, no. This version will do [ALL *] if one failed config matches all the config values in the `print_matrix_configuration.py`. 
This version cannot do [ALL DEVICES] for any 2+ devices. Because we ignores all succeeded configs(devices in this case). The count of succeeded devices is critical here. I cannot make sure [ALL DEVICES] failed until I know (the count of succeeded devices == 0).

jonsimantov (approved):

PR #785, opened 2021-12-02
Timeout & retry "install integration test deps" step

sunmou99:
### Description
Install deps step can be flaky and break the build. [e.g.](https://github.com/firebase/firebase-cpp-sdk/runs/4392718616?check_suite_focus=true)
Add timeout & retry

***
### Testing
See the comment section

***

### Type of Change
***

#### Notes

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -145,7 +145,12 @@ jobs:
       with:
         python-version: ${{ env.pythonVersion }}
     - name: Install python deps
-      run: pip install -r scripts/gha/requirements.txt
+      uses: nick-invision/retry@v2
+      with:
+        timeout_minutes: 1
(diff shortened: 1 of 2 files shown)


jonsimantov:
I don't think this is the only issue - the Cache NDK step failed, which caused it to try to download the NDK into a partially-filled out directory. I think we also need to ensure that the NDK is unzipped with -y so it doesn't prompt overwriting files.

sunmou99:
> I don't think this is the only issue - the Cache NDK step failed, which caused it to try to download the NDK into a partially-filled out directory. I think we also need to ensure that the NDK is unzipped with -o so it doesn't prompt overwriting files. (I pushed a commit to do this.)

Yes, you are right. And thanks for your fixing commit.

jonsimantov (approved):

PR #798, opened 2021-12-20
future.h: replace std::mutex with firebase::Mutex

dconeybe:
Replace `std::mutex` with `firebase::Mutex` in `future.h`.

This is the fourth, and last, PR (the previous ones being #751, #792, and #795) towards moving `firebase::Mutex` into the public-internal includes directory, so that it can be used by `future.h`. See #747 for the rationale.

Googlers can see b/206520921 for more details.

### DIFF
--- a/app/src/include/firebase/internal/future_impl.h
+++ b/app/src/include/firebase/internal/future_impl.h
@@ -189,11 +189,14 @@ inline FutureBase::CompletionCallbackHandle Future<ResultType>::AddOnCompletion(
 
 #endif  // defined(INTERNAL_EXPERIMENTAL)
 
-inline FutureBase::FutureBase() : api_(NULL), handle_(0) {}  // NOLINT
+inline FutureBase::FutureBase()
+    : mutex_(Mutex::Mode::kModeNonRecursive),
+      api_(NULL),
+      handle_(0) {}  // NOLINT
 
 inline FutureBase::FutureBase(detail::FutureApiInterface* api,
                               const FutureHandle& handle)
-    : api_(api), handle_(handle) {
+    : mutex_(Mutex::Mode::kModeNonRecursive), api_(api), handle_(handle) {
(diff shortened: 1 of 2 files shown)


app/src/include/firebase/internal/future_impl.h
```diff
@@ -189,11 +189,14 @@ inline FutureBase::CompletionCallbackHandle Future<ResultType>::AddOnCompletion(
...
-inline FutureBase::FutureBase() : api_(NULL), handle_(0) {}  // NOLINT
+inline FutureBase::FutureBase()
+    : mutex_(Mutex::Mode::kModeNonRecursive),
+      api_(NULL),
+      handle_(0) {}  // NOLINT
 
 inline FutureBase::FutureBase(detail::FutureApiInterface* api,
                               const FutureHandle& handle)
-    : api_(api), handle_(handle) {
+    : mutex_(Mutex::Mode::kModeNonRecursive), api_(api), handle_(handle) {
```
jonsimantov, line 199: Is this in here just because std::mutex happens to be nonrecursive, so we know it works?

dconeybe: Basically, yes. It will make it less "scary" to convert to std::mutex some day in the future since we know it's works in non-recursive mode.

dconeybe: Should I make a subclass of `Mutex` called `NonRecursiveMutex` which simply specifies `Mutex::Mode::kModeNonRecursive` to the superclass' constructor? That would avoid having to explicitly initialize the `Mutex` member variable and could encourage the use of non-recursive mutexes, making it easier to port to `std::mutex` some day.

jonsimantov: Nah, I think it's fine as-is. Was just curious.

jonsimantov (approved):

PR #800, opened 2021-12-20
When testing on the expanded matrix, run tests on M1 Mac (via custom runner) as well.

jonsimantov:
If running tests against the expanded matrix, it will build an ARM64 version of the integration tests, and run it on an M1 Mac GitHub runner that we host ourselves.

For security purposes, this job does not use any third-party marketplace actions, nor does it pass in secrets on the command line.

The test_desktop_custom_runner step is written as generically as possible to support other custom runners we may add in the future.
***

Run integration-tests workflow on the expanded matrix. This works both on prebuilt SDK and from source.
***

### Type of Change
***

#### Notes

### DIFF
--- a/.github/workflows/integration_tests.yml
+++ b/.github/workflows/integration_tests.yml
@@ -778,7 +778,7 @@ jobs:
     if: |
       ${{
         contains(needs.check_and_prepare.outputs.matrix_platform, 'Desktop') &&
-        contains(needs.check_and_prepare.outputs.matrix_os, matrix.built_on) &&
+        contains(needs.check_and_prepare.outputs.matrix_os, strategy.matrix.built_on) &&
         github.event_name == 'workflow_dispatch' &&
         needs.check_and_prepare.outputs.apis != '' &&
         !cancelled() &&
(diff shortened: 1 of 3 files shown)


jonsimantov:
@sunmou99 I'm still considering this a draft change because I want to make sure that the runner is stable before merging this, but could you take a look at the changes I've made in this PR and give me your comments?

In particular I am wondering if it would be better to separate out architecture in the summarize script rather than the hack I've done. I'm intending to add x86 integration tests soon, so maybe that would be a better time to make that change.

.github/workflows/integration_tests.yml
```diff
@@ -203,7 +221,7 @@ jobs:
           --run_id ${{github.run_id}}
 
   build_desktop:
-    name: build-desktop-${{ matrix.os }}-${{ matrix.ssl_variant }}
+    name: build-desktop-${{ matrix.os }}-${{ matrix.ssl_variant }}-${{ matrix.arch }}
```
sunmou99, line 224: How about reverse the order of the last 2 variables?
build-desktop-${{ matrix.os }}-${{ matrix.arch }}-${{ matrix.ssl_variant }}

scripts/gha/summarize_test_results.py
```diff
@@ -75,6 +75,7 @@
 
 CAPITALIZATIONS = {
     "macos": "MacOS",
+    "macosarm": "MacOS-ARM",
```
sunmou99, line 78: I prefer to have only one MacOS, and make arch as a secondary description. Some more extra work is required here for the result summary report. I'll figure it out.

jonsimantov: Do you want to work in this same branch?

.github/workflows/integration_tests.yml
jonsimantov: Done.

PR #801, opened 2021-12-21
Added legacy AdMob SDK back into GMA feature branch.

DellaBitta:
- Adds the older AdMob SDK back into the source base.  This way we can have a longer deprecation window.
- Updates the way the AdMob C++ SDK for iOS tracks header dependencies when building its framework.  Our build system doesn't allow us to be dependent on two different versions of the GMA iOS SDK pods (1: older Firebase AdMob pinned version and 2: the new Firebase GMA SDK which will track new iOS GMA SDK releases). Therefore, 
  - the `AdMob` SDK is now tied to a pinned version of headers which have been checked-in to this `admob/admob_resources/ios_headers`, and
  - the new Google Mobile Ads (`GMA`) SDK uses the standard`ios_pod/Podfile` to track the latest and greatest iOS GMA SDK.
- removed `tools/ios/tetapp` from both `admob/` and `gma/`.  Not quite sure why it was migrated from G3.

Changed files of note:
- admob/admob_resources/ios_headers/
  - contains the GMA SDK 7.69 header files.
- CMakeLists.txt
  - reinstated `add_subdirectory(admob)`
- app/CMakeLists.txt
  - reinstated `admob_HDRS`
- build_scripts/ios/build.sh
  - reinstated `firebase_admob`
- scripts/gha/build_ios_tvos.py
  - reinstated `firebase_admob`

***

[Integration Test CI for AdMob and GMA targets](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1613495134)
[Packaging CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1613495342)
***

### Type of Change
***


admob/admob_resources/ios_headers/GoogleMobileAds/GADAdLoader.h
```diff
@@ -0,0 +1,47 @@
...
+
+/// Object notified when an ad request succeeds or fails. Must conform to requested ad types'
+/// delegate protocols.
+@property(nonatomic, weak, nullable) id<GADAdLoaderDelegate> delegate;
+
+/// The ad loader's ad unit ID.
+@property(nonatomic, readonly, nonnull) NSString *adUnitID;
+
+/// Indicates whether the ad loader is loading.
+@property(nonatomic, getter=isLoading, readonly) BOOL loading;
```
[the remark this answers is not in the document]
DellaBitta: Linter doesn't like this format, but formatter reformats it back to this style if manually changed.  Same for all of the other Linter warnings.  Resolving them for now.

jonsimantov:
FYI we can add a .clang-format-ignore file and have it ignore that whole directory. This way you can just use the files as-is.

I also think there is a CPPLINT.cfg file we can put into that directory and set it to exclude all the files.

DellaBitta:
> FYI we can add a .clang-format-ignore file and have it ignore that whole directory. This way you can just use the files as-is.
> 
> I also think there is a CPPLINT.cfg file we can put into that directory and set it to exclude all the files.

Ok. Well the files are already formatted so I didn't add a clang format ignore file.  I did add a lint suppression file, however. 

Hopefully these files never change and it's not required, but it's nice not to have to dismiss the linter's comments in this PR at least. :)

admob/CMakeLists.txt
```diff
@@ -0,0 +1,123 @@
...
+target_link_libraries(firebase_admob
+    PUBLIC firebase_app)
+# Public headers all refer to each other relative to the src/include directory,
+# while private headers are relative to the entire C++ SDK directory.
+target_include_directories(firebase_admob
+  PUBLIC
+    ${CMAKE_CURRENT_LIST_DIR}/src/include
+  PRIVATE
+    ${FIREBASE_CPP_SDK_ROOT_DIR}
+    ${CMAKE_CURRENT_LIST_DIR}/admob_resources/ios_headers
```
jonsimantov, line 91: Can you set this to only be used in ios builds?

Also put a small comment here that this replaces the cocoapod.

jonsimantov (approved):
LGTM except my comment about header paths on non-ios.

admob/CMakeLists.txt
DellaBitta: Done!

PR #811, opened 2022-01-10
GMA - Fix Error in JNI method invocation on BannerView destruction

DellaBitta:
There were stability issues when calling a JNI method and then immediately removing the global reference.  This manifested in JNI corruption errors on simulators.  

The previous BannerView destructor attempted to protect developers from accidentally leaking resources if the BannerView's destructor was called before the Destroy() method was invoked.  Our protected caused JNI assertions in simulators where the JNI parameters were corrupted.  I assume this was due to the  JNI object reference being deleted admist an asynchronous operation.

Therefore this change removes the attempt o invoke the JNI Destroy() method in the destructor on Android.  BannerView's header comments are updated to explicitly state that Destroy should be invoked before the C++ object's destructor is called.  Additionally log a warning if this scenario occurs.

***

* Local Simulator Testing
* [Integration CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1680038278)
* [Packaging CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1680041167) and its [Integration Tests](https://github.com/firebase/firebase-cpp-sdk/actions/runs/1680392713)

***

### Type of Change
***

### DIFF
--- a/gma/src/android/banner_view_internal_android.cc
+++ b/gma/src/android/banner_view_internal_android.cc
@@ -145,8 +145,13 @@ BannerViewInternalAndroid::BannerViewInternalAndroid(BannerView* base)
 
 BannerViewInternalAndroid::~BannerViewInternalAndroid() {
   firebase::MutexLock lock(mutex_);
-  JNIEnv* env = ::firebase::gma::GetJNI();
+  if (!destroyed_) {
+    LogWarning(
+        "Warning: AdView destructor invoked before the application "
+        " called Destroy() on the object. A Memory Leak may occur.");
+  }
 
+  JNIEnv* env = ::firebase::gma::GetJNI();
   env->DeleteGlobalRef(ad_view_);
   ad_view_ = nullptr;
 
--- a/gma/src/android/banner_view_internal_android.cc
+++ b/gma/src/android/banner_view_internal_android.cc
@@ -144,13 +144,6 @@ BannerViewInternalAndroid::~BannerViewInternalAndroid() {
   firebase::MutexLock lock(mutex_);
   JNIEnv* env = ::firebase::gma::GetJNI();
 
-  // The application should have invoked Destroy already, but
-  // invoke it now just in case they haven't in the hope that
-  // we can prevent leaking memory.
-  env->CallVoidMethod(
-      helper_, banner_view_helper::GetMethodId(banner_view_helper::kDestroy),
-      /*callbackDataPtr=*/nullptr, /*destructor_invocation=*/true);
-
   env->DeleteGlobalRef(ad_view_);
   ad_view_ = nullptr;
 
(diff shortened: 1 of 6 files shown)


gma/src/android/banner_view_internal_android.cc
```diff
@@ -142,15 +145,13 @@ BannerViewInternalAndroid::BannerViewInternalAndroid(BannerView* base)
...
-  // The application should have invoked Destroy already, but
-  // invoke it now just in case they haven't in the hope that
-  // we can prevent leaking memory.
-  env->CallVoidMethod(
-      helper_, banner_view_helper::GetMethodId(banner_view_helper::kDestroy),
-      /*callbackDataPtr=*/nullptr, /*destructor_invocation=*/true);
+  if (!destroyed_) {
+    LogWarning(
+        "Warning: AdView destructor invoked before the application "
+        " called Destroy() on the object. A Memory Leak may occur.");
```
jonsimantov, line 151: I think it's probably fine to just warn the developer, but as a way to make this more bulletproof, what about using the cleanup notifier to delegate the Destroy steps if the AdView destructor is called?

You could register object A to the cleanup notifier, with a deletion callback that performs the same thing on object A's contents as Destroy() would (but ignoring the result).

DellaBitta:
I think the problem is the ignoring the result part.  The corruption seems to stem from removing the object reference after the asynchronous destruction operation is kicked off but before it completes. 

Maybe in Java I can switch which C++ function callback is invoked. Instead of using a boolean to denote if BannerViewHelper::Destroy was invoked by the Destructor, as it is now, I could pass the global reference to the BannerViewHelper.  If the global reference is NULL then the current C++ Destroy -> Java Destroy() -> C+++ future path is invoked.  If the global reference is non-null, then Java calls back into a different C++ static function with the Global Reference as a parameter.  The C++ static function then releases the global reference of the Java Object that time, and the C++ destructor doesn't even touch it.

I can try it out tomorrow.

gma/src/android/banner_view_internal_android.cc
DellaBitta: I think the problem is the ignoring the result part.  The corruption seems to stem from removing the object reference after the asynchronous destruction operation is kicked off but before it completes. 

Maybe in Java I can switch which C++ function callback is invoked. Instead of using a boolean to denote if BannerViewHelper::Destroy was invoked by the Destructor, as it is now, I could pass the global reference to the BannerViewHelper.  If the global reference is NULL then the current C++ Destroy -> Java Destroy() -> C+++ future path is invoked.  If the global reference is non-null, then Java calls back into a different C++ static function with the Global Reference as a parameter.  The C++ static function then releases the global reference of the Java Object that time, and the C++ destructor doesn't even touch it.

I can try it out tomorrow.

DellaBitta: Or maybe just add the reference as an optional parameter and keep the boolean.

DellaBitta: Done. PTAL.

jonsimantov (approved):

PR #824, opened 2022-01-25
Added logic to retry transient failures for all operations in desktop storage implementation.

AlmostMatt:
For each retryable operation in firebase storage desktop implementation, after sending an initial request a separate thread will be created which will asynchronously retry transient failures with exponential backoff until a maximum retry deadline is reached.
This affects: GetBytes, PutBytes, GetFile, PutFile, GetMetadata, UpdateMetadata, GetDownloadUrl, and Delete
The public API and usage are both unchanged.
This causes the desktop implementation to be more similar to iOS and Android implementations which already have retry logic built into each of these operations.

***

Added integration tests to verify retry functionality.
In order to force failures, Get operations are called when the resource has not yet been Put. For PutFile, failure is caused by trying to upload a nonexistent local file. test-only flags enable retrying for these types of failures even though normally only transient failures would be retried.
In order to make retried requests succeed, an async thread creates the relevant object after a few seconds.
 
***

### Type of Change
- [ x ] New feature. A non-breaking change which adds functionality.
***

#### Notes


storage/integration_test/src/integration_test.cc
```diff
@@ -19,6 +19,7 @@
 #include <cstdlib>
 #include <cstring>
 #include <ctime>
+#include <thread>
```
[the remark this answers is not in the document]
jonsimantov: I think add a NOLINT on these would be good.

storage/integration_test/src/integration_test.cc
```diff
@@ -386,6 +387,102 @@ const std::string kSimpleTestFile =
...
+
+  // Set max-retry time to 10s and create the file in an async thread after 30 seconds
+  // Verify that GetBytes fails
+  {
+    LogDebug("Call PutBytes after the maximum retry deadline.");
+    firebase::storage::StorageReference ref = CreateFolder().Child("File3.txt");
+    LogDebug("Storage URL: gs://%s%s", ref.bucket().c_str(),
+            ref.full_path().c_str());
+    cleanup_files_.push_back(ref);
+
```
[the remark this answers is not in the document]
jonsimantov: You can run "scripts/format_code.py -git_diff" to auto-format the code.

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -279,18 +277,95 @@ Future<size_t> StorageReferenceInternal::GetBytes(void* buffer,
...
+                                   *(buffer_index_future.result()));
+  } else {
+    future_api->Complete(final_handle, internal_future.error(),
+                         internal_future.error_message());
+  }
+}
 
-  return GetBytesLastResult();
+// Returns whether or not an http status represents a failure that should be retried.
+bool StorageReferenceInternal::IsRetryableFailure(int httpStatus) {
```
jonsimantov, line 363: Here is one ugly way to set up the testing override:

You could set up a global variable here:
static bool g_retry_all_futures_for_testing = false;

And then in the integration_test.cc:
#if FIREBASE_PLATFORM_DESKTOP
namespace firebase {
namespace storage {
namespace internal {
extern bool g_retry_all_futures_for_testing;
}
}
}
#endif  // FIREBASE_PLATFORM_DESKTOP


And in the test that forces retrying futures, inside an #if FIREBASE_PLATFORM_DESKTOP:
bool old_value = g_retry_all_futures_for_testing;
g_retry_all_futures_for_testing = true;
// ..... do all the testing here  .....
g_retry_all_futures_for_testing = old_value;

You could also make the thing you extern a function rather than a global.

jonsimantov (commented):
Not a full review, just a suggestion for forcing the retry for testing.

storage/src/desktop/storage_reference_desktop.cc
jonsimantov: (also, feel free to name it better, like retry_all_http_statuses or whatever)

storage/integration_test/src/integration_test.cc
```diff
@@ -386,6 +387,102 @@ const std::string kSimpleTestFile =
...
+    LogDebug("Call GetBytes on a nonexistent ref.");
+    firebase::storage::StorageReference ref =
+        CreateFolder().Child("File1.txt");
+    LogDebug("Storage URL: gs://%s%s", ref.bucket().c_str(),
+            ref.full_path().c_str());
+    cleanup_files_.push_back(ref);
+
+    const size_t kBufferSize = 1024;
+    char buffer[kBufferSize];
+    memset(buffer, 0, sizeof(buffer));
```
jonsimantov, line 408: If we're not checking that buffer stays all 0s, any reason to memset it?

jonsimantov (changes requested):
Also, could you add support for the remaining operations, e.g. PutFile?

storage/integration_test/src/integration_test.cc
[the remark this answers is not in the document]
AlmostMatt: Done. But if we are ok with using unapproved headers, should we try to get them approved or remove the warnings?

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -14,9 +14,12 @@
 
 #include "storage/src/desktop/storage_reference_desktop.h"
 
+#include <chrono>
```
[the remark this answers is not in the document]
AlmostMatt: Added   // NOLINT

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -14,9 +14,12 @@
...
 #include <limits>
 #include <memory>
+#include <thread>
```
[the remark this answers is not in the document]
AlmostMatt: Added  // NOLINT

storage/src/desktop/storage_reference_desktop.cc
AlmostMatt: Done. 

I made the option only retry 404 failures since I don't want it to retry successful statuses like 200 or 204 and so retrying on a single failure status is the simplest solution.

I am not quite sure if there is a best place in each of these files to place the code snippets that you shared. I put the g_retry_file_not_found__for_testing definition right about the IsRetryable method, and the extern line near the top of the integration test.

I wrapped the entire test in an IF FIREBASE_PLATFORM_DESKTOP check.

storage/integration_test/src/integration_test.cc
AlmostMatt: No reason, removed this line.

storage/integration_test/src/integration_test.cc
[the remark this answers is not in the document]
jonsimantov: (Discussing offline.)

storage/integration_test/src/integration_test.cc
```diff
@@ -40,6 +41,17 @@
 #define FIREBASE_CONFIG_STRING ""
 #endif  // FIREBASE_CONFIG
 
+// Allow integration tests to enable retrying of file-not-found errors.
+#if FIREBASE_PLATFORM_DESKTOP
+namespace firebase {
+namespace storage {
+namespace internal {
+extern bool g_retry_file_not_found__for_testing;
```
jonsimantov, line 49: nit: there's an extra underscore between "found" and "for"

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -279,18 +277,96 @@ Future<size_t> StorageReferenceInternal::GetBytes(void* buffer,
...
+    }
+    // Interrupt the loop if the retry deadline has been reached
+    auto current_time = std::chrono::steady_clock::now();
+    if (current_time + current_sleep_time > end_time) {
+      break;
+    }
+    // Sleep for an exponentially increasing duration, then retry the request.
+    std::this_thread::sleep_for(current_sleep_time);
+    if (current_sleep_time < std::chrono::milliseconds(kMaxSleepTimeMillis)) {
+      current_sleep_time *= 2;
```
jonsimantov, line 346: Probably best to cap it at kMaxSleepTimeMillis rather than *=2 if it's below that; otherwise, the max it will reach right now is 32000 rather than the max of 30000.

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -279,18 +277,96 @@ Future<size_t> StorageReferenceInternal::GetBytes(void* buffer,
...
+    }
+    // For any request that succeeds or fails in a non-retryable way, don't
+    // bother retrying.
+    if (internal_future.status() != firebase::kFutureStatusComplete ||
+        !IsRetryableFailure(response->status())) {
+      break;
+    }
+    // Interrupt the loop if the retry deadline has been reached
+    auto current_time = std::chrono::steady_clock::now();
+    if (current_time + current_sleep_time > end_time) {
```
jonsimantov, line 340: It might be nice, if current_time + current_sleep_time > end_time, but current_sleep_time is quite large, to maybe give it one last retry when the maximum time is almost up?

e.g. Right now, current_sleep_time is 30 seconds, and end_time is 25 seconds from now, it won't actually do a last retry before the 25 seconds are up. I'm not sure if this is a major issue, but it might be nice to detect this case and maybe reduce current_sleep_time a bit.

(Just a suggestion, feel free to ignore.)

storage/src/desktop/storage_reference_desktop.cc
jonsimantov: This part looks good to me.

storage/integration_test/src/integration_test.cc
```diff
@@ -386,6 +398,112 @@ const std::string kSimpleTestFile =
...
+          RunWithRetry<firebase::storage::Metadata>([&]() {
+            return ref.PutBytes(&kSimpleTestFile[0], kSimpleTestFile.size());
+          });
+      WaitForCompletion(future, "PutBytes");
+      auto metadata = future.result();
+      EXPECT_EQ(metadata->size_bytes(), kSimpleTestFile.size());
+    });
+    const size_t kBufferSize = 1024;
+    char buffer[kBufferSize];
+    memset(buffer, 0, sizeof(buffer));
```
jonsimantov, line 456: I see a few other extraneous memsets

storage/integration_test/src/integration_test.cc
```diff
@@ -386,6 +398,112 @@ const std::string kSimpleTestFile =
...
+TEST_F(FirebaseStorageTest, TestReadBytesWithMaxRetryDuration) {
+  // Enable retrying of file-not-found errors for testing.
+  bool old_value = firebase::storage::internal::g_retry_file_not_found__for_testing;
+  firebase::storage::internal::g_retry_file_not_found__for_testing = true;
+
+  int shorter_duration_sec = 2;
+  int longer_duration_sec = 6;
+  SignIn();
+
+  // Attempt to read a non-existent file, and verify that GetBytes fails.
```
jonsimantov, line 415: This is just a baseline test, right? We shouldn't actually check to see that it waited some time before failing?

storage/integration_test/src/integration_test.cc
```diff
@@ -386,6 +398,112 @@ const std::string kSimpleTestFile =
...
+  {
+    LogDebug("Call GetBytes on a nonexistent ref.");
+    firebase::storage::StorageReference ref = CreateFolder().Child("File1.txt");
+    LogDebug("Storage URL: gs://%s%s", ref.bucket().c_str(),
+             ref.full_path().c_str());
+    cleanup_files_.push_back(ref);
```
jonsimantov, line 421: You only need to add ref to cleanup_files if a file was uploaded into that ref.

Although I think it might be best to use the same ref for the first test (to ensure the file starts out missing) and the second test (to ensure that it retries on the file that we previously confirmed is missing).

jonsimantov (commented):
Mostly there, although I'd really like to see support added for the other async operations.

storage/src/desktop/storage_reference_desktop.cc
AlmostMatt: Done, I now set to max_sleep_time if it exceeds max_sleep_time.

storage/integration_test/src/integration_test.cc
AlmostMatt: fixed

AlmostMatt:
I responded to the various minor feedback.
I am still making the change to add retry logic to the put operations. The put operations are structured differently since they they put both bytes/file and also update metadata, so I am just thinking a bit about how best to structure this.

I will respond to this merge request again once put operations are supported.

storage/src/desktop/storage_reference_desktop.cc
AlmostMatt: I based this exponential backoff logic on the ExponentialBackoffSender that is used for firebase android downloads. It has the following check: 
    while (clock.elapsedRealtime() + currentSleepTime <= deadLine ...
so I prefer to keep this logic simpler for now and similarly stop if the sleep time would exceed the deadline.

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -279,18 +285,106 @@ Future<size_t> StorageReferenceInternal::GetBytes(void* buffer,
...
+      current_sleep_time = max_sleep_time;
+    }
+    response = send_request_funct();
+  }
+  // Copy from the internal future to the final future.
+  Future<FutureType> typed_future =
+      static_cast<const Future<FutureType>&>(internal_future);
+  if (typed_future.result() != nullptr) {
+    if constexpr (std::is_void<FutureType>::value) {
+      future_api->Complete(final_handle, internal_future.error());
```
AlmostMatt, line 368: Delete() returns a Future<void>. The code for Delete seemed to end up passing the result() != nullptr check but I initially encountered an error about not being able to get a reference to void, so I added an is_void check.

Another alternative would probably be to explicitly define a variation of this function for type void.

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -368,22 +469,34 @@ Future<Metadata> StorageReferenceInternal::PutFileInternal(
...
+        auto* future_api = future();
+        auto handle =
+            future_api->SafeAlloc<Metadata>(kStorageReferenceFnPutFileInternal);
+
+        // Open the file, calculate the length.
+        storage::internal::RequestFile* request(
+            new storage::internal::RequestFile(final_path.c_str(), 0));
+        if (!request->IsFileOpen()) {
+          delete request;
+          future_api->Complete(handle, kErrorUnknown, "Could not read file.");
```
AlmostMatt, line 484: PutFile has the unique failure case of being unable to read the file. Since this completes with an error, it should not have a retryable http status, and so it should not retry the request.

AlmostMatt:
I added retry logic for each of the other operations in the desktop implementation of storage: get and update metadata, delete, putfileinternal, and putbytesinternal.
I will followup to see if I can any other integration tests related to retrying.

AlmostMatt:
I added additional integration test cases to test retry for GetFile, GetMetadata, and Delete. These all passed for me locally.
These are basically copy-pasted from the test that retries GetBytes but with operations changed.

I am in the process of seeing if there is a way that I can cause PutFile to fail in a retryable way, possibly by having the local file not exist.

storage/integration_test/src/integration_test.cc
```diff
@@ -630,6 +642,402 @@ TEST_F(FirebaseStorageTest, TestDeleteFile) {
...
+  {
+    LogDebug("Call PutBytes after the maximum retry deadline.");
+    firebase::storage::StorageReference ref = CreateFolder().Child("File3.txt");
+    LogDebug("Storage URL: gs://%s%s", ref.bucket().c_str(),
+             ref.full_path().c_str());
+    cleanup_files_.push_back(ref);
+
+    std::thread t1([&] {
+      // PutBytes after a long delay.
+      std::this_thread::sleep_for(
```
jonsimantov, line 708: Here you could (and should) actually check the status of the original future to ensure it's completed. You can even call WaitForCompletion in multiple threads at once.

storage/src/desktop/storage_reference_desktop.h
```diff
@@ -28,6 +28,25 @@ namespace firebase {
...
 namespace internal {
 
+enum StorageReferenceFn {
+  kStorageReferenceFnDelete = 0,
+  kStorageReferenceFnDeleteInternal,
+  kStorageReferenceFnGetBytes,
+  kStorageReferenceFnGetBytesInternal,
+  kStorageReferenceFnGetFile,
+  kStorageReferenceFnGetFileInternal,
+  kStorageReferenceFnGetDownloadUrl,
```
jonsimantov, line 38: I think GetDownloadUrl could also fail and should get a retry.

storage/src/desktop/storage_reference_desktop.cc
jonsimantov: I think this is fine, as long as it works on all 3 desktop platforms.

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -279,18 +285,106 @@ Future<size_t> StorageReferenceInternal::GetBytes(void* buffer,
...
+                         internal_future.error_message());
+  }
+}
 
-  storage::internal::Request* request = new storage::internal::Request();
-  PrepareRequest(request, storageUri_.AsHttpUrl().c_str(), rest::util::kGet);
-  RestCall(request, request->notifier(), response, handle.get(), listener,
-           controller_out);
+// Can be set in tests to retry 404s, but is not exposed in the public API.
+bool g_retry_file_not_found_for_testing = false;
```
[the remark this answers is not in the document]
jonsimantov: Yep, sounds good.

storage/src/desktop/storage_reference_desktop.cc
jonsimantov: Makes sense.

jonsimantov:
> I am in the process of seeing if there is a way that I can cause PutFile to fail in a retryable way, possibly by having the local file not exist.

This is a pretty good idea, you can do the same sort of flag to force it to retry PutFile.

storage/src/desktop/storage_reference_desktop.h
AlmostMatt: GetDownloadUrl is currently implemented as just calling GetMetadata and then returning the download url from the metadata, so it effectively gets retry logic for free from the internal call to GetMetadata.

storage/integration_test/src/integration_test.cc
```diff
@@ -47,7 +47,8 @@ namespace firebase {
 namespace storage {
 namespace internal {
 extern bool g_retry_file_not_found_for_testing;
-}
+extern bool g_retry_failed_future_for_testing;
```
jonsimantov, line 50: Could we have a single flag? g_retry_all_errors_for_testing

storage/integration_test/src/integration_test.cc
```diff
@@ -1038,6 +1039,103 @@ TEST_F(FirebaseStorageTest, TestDeleteWithMaxRetryDuration) {
...
+  int longer_duration_sec = 6;
+  SignIn();
+
+  std::string path = PathForResource() + kPutFileTestFile;
+  // Cloud Storage expects a URI, so add file:// in front of local
+  // paths.
+  std::string file_path = kFileUriScheme + path;
+
+  // Delete the local file first to guarantee it does not already exist.
+  std::remove(path.c_str());
```
jonsimantov, line 1060: Make sure this works on Windows (it probably does, but just to be safe)

jonsimantov (approved):

release_build_files/readme.md
```diff
@@ -573,6 +573,7 @@ code.
         initialization of Firebase iOS SDK.
     -   Remote Config: Fixed default `Fetch()` timeout being 1000 times too
         high.
+    -   Added retry logic to desktop implementation of storage operations.
```
jonsimantov, line 576: Make sure to put "Storage (Desktop):" at the beginning.

Suggested phrasing: Storage (Desktop): Added retry logic to PutFile, GetFile, and other operations.

jonsimantov (changes requested):
Looks good except for a minor nit or two (and the readme phrasing)

storage/integration_test/src/integration_test.cc
AlmostMatt: Done. When g_retry_all_errors_for_testing is true, IsRetryableFailure will retry anything that has a http status other than 200-299 since 2xx indicates succees.

release_build_files/readme.md
AlmostMatt: Done.

storage/integration_test/src/integration_test.cc
AlmostMatt: std::remove should be supported on Windows. The only operating-system specific comment I found about the function is related to whether or not it will delete a file that is already opened on Windows
https://www.programiz.com/cpp-programming/library-function/cstdio/remove

If I do an automated integration test run will it run a windows version of the integration test? That seems like a reasonable way to be extra sure.

jonsimantov (approved):

storage/integration_test/src/integration_test.cc
AlmostMatt: Confirmed that this test worked on windows:
TestPutFileWithMaxRetryDuration passed in the following test run
https://github.com/firebase/firebase-cpp-sdk/runs/5419440986?check_suite_focus=true

jonsimantov (approved):

PR #842, opened 2022-02-07
Remove the "unapproved C++ header" lint warning.

jonsimantov:
### Description
Remove the "unapproved C++ header" lint warning.

(Also add a note explaining each lint warning we ignore.)
***
### Testing
Ran lint_commenter.py script to ensure it doesn't fail.
***

### Type of Change
***

### DIFF
--- a/scripts/gha/lint_commenter.py
+++ b/scripts/gha/lint_commenter.py
@@ -47,7 +47,7 @@ IGNORE_LINT_WARNINGS = [
 ]
 # Exclude files within the following paths (specified as regexes).
 EXCLUDE_PATH_REGEX = [
-  # These files are copied from an external repo and our outside our control.
+  # These files are copied from an external repo and are outside our control.
   r'^analytics/ios_headers/'
 ]
 # The linter gives every error a confidence score.
--- a/scripts/gha/lint_commenter.py
+++ b/scripts/gha/lint_commenter.py
@@ -37,15 +37,18 @@ from unidiff import PatchSet
 import urllib.parse
 
 # Put any lint warnings you want to fully ignore into this list.
+# (Include a short explanation of why.)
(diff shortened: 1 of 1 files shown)


scripts/gha/lint_commenter.py
```diff
@@ -37,15 +37,18 @@
...
 
 # Put any lint warnings you want to fully ignore into this list.
+# (Include a short explanation of why.)
 IGNORE_LINT_WARNINGS = [
-  'build/include_subdir',
-  'readability/casting',
-  'whitespace/indent',
-  'whitespace/line_length'
+  'build/include_subdir',   # doesn't know about our include paths
+  'build/c++11',            # ignore "unapproved c++11 header" warning
```
AlmostMatt, line 43: Where can I see what warnings fall into this category? I am wondering if this might cause us to miss out on some other useful lint warnings.

AlmostMatt (commented):
Sorry I didn't respond earlier, I didn't notice that I was marked as the reviewer for this.

scripts/gha/lint_commenter.py
jonsimantov: You can search this file for "build/c++11": https://github.com/google/styleguide/blob/gh-pages/cpplint/cpplint.py#L5854

It appears to only have the "unapproved C++11 header" and "unapproved C++11 class or function" - since we aren't writing code on the internal repo, I think it's probably fine.

AlmostMatt: Ok got it. The check seems to be just for unapproved headers and unapproved classes, and the reason for such things to be unapproved is generally because there is a google3 version of the same concept, but we are not in the google3 codebase. So disabling the check looks good to me.

AlmostMatt (approved):

PR #846, opened 2022-02-10
Heartbeat logging passthrough for Auth on Mobile Platforms

AlmostMatt:
Update c++ GetAuth to always call native auth getters so that any logging internal to the native auth implementations will be triggered.
This is as described in the "How data is collected" section of
https://docs.google.com/document/d/1eVlEmjdxSsXsvnzGiRYYfb0RtfWRLi3yo9vHj0NhqE4/edit#heading=h.ohspo4gdlkj0 

This change is currently only for Auth, but similar changes will be done separately for Database, Firestore, Functions, Installation and Remote Config.
***

This is effectively no-op for now, since native SDKs do not yet log heartbeats.
Once heartbeat logging from getters is supported, manual testing will verify the process end to end.
***

### Type of Change
Place an `x` the applicable box:
- [ ] Bug fix. Add the issue # below if applicable.
- [ ] New feature. A non-breaking change which adds functionality.
- [x] Other, such as a build process or documentation change.

This is a minor change in order to enable a new feature in the future.
***

#### Notes
- Bug fixes and feature changes require an update to the `Release Notes` section of `release_build_files/readme.md`.
- Read the contribution guidelines [CONTRIBUTING.md](https://github.com/firebase/firebase-cpp-sdk/blob/main/CONTRIBUTING.md).
- Changes to the public API require an internal API review. If you'd like to help us make Firebase APIs better, please propose your change in a feature request so that we can discuss it together.

PR #853, opened 2022-02-17
Heartbeat logging passthrough for FIS for mobile platforms.

AlmostMatt:
Update c++ FIS GetInstance to always call native FIS getters so that any logging internal to the native implementations will be triggered.
This is as described in the "How data is collected" section of
https://docs.google.com/document/d/1eVlEmjdxSsXsvnzGiRYYfb0RtfWRLi3yo9vHj0NhqE4/edit#heading=h.ohspo4gdlkj0 

This is effectively the same as https://github.com/firebase/firebase-cpp-sdk/pull/846, but for FIS
***

This is effectively no-op for now, since native SDKs do not yet log heartbeats.
Once heartbeat logging from getters is supported, manual testing will verify the process end to end.
***

### Type of Change

This is a minor change in order to enable a new feature in the future.
***

#### Notes

### DIFF
--- a/installations/src/android/installations_android.cc
+++ b/installations/src/android/installations_android.cc
@@ -197,7 +197,6 @@ void InstallationsInternal::LogHeartbeat(const firebase::App& app) {
   FIREBASE_ASSERT(installations_instance_local);
   env->DeleteLocalRef(installations_instance_local);
   env->DeleteLocalRef(platform_app);
-  LogDebug("%s API Initialized", kApiIdentifier);
 }
 
 Future<std::string> InstallationsInternal::GetId() {
--- a/installations/src/android/installations_android.cc
+++ b/installations/src/android/installations_android.cc
@@ -186,6 +186,20 @@ void InstallationsInternal::Cleanup() {
   }
 }
 
+void InstallationsInternal::LogHeartbeat(const firebase::App& app) {
+  // Calling the native getter is sufficient to cause a Heartbeat to be logged.
+  JNIEnv* env = app.GetJNIEnv();
+  jobject platform_app = app.GetPlatformApp();
+  jclass installations_class = installations::GetClass();
+  jobject installations_instance_local = env->CallStaticObjectMethod(
+      installations_class,
+      installations::GetMethodId(installations::kGetInstance), platform_app);
+  FIREBASE_ASSERT(installations_instance_local);
+  env->DeleteLocalRef(installations_instance_local);
+  env->DeleteLocalRef(platform_app);
+  LogDebug("%s API Initialized", kApiIdentifier);
+}
+
 Future<std::string> InstallationsInternal::GetId() {
   const auto handle =
       future_impl_.SafeAlloc<std::string>(kInstallationsFnGetId);
--- a/installations/src/android/installations_android.cc
+++ b/installations/src/android/installations_android.cc
@@ -15,6 +15,7 @@
 #include "installations/src/android/installations_android.h"
 
 #include "app/src/util.h"
+#include "app/src/util_android.h"
 #include "installations/src/common.h"
 
 namespace firebase {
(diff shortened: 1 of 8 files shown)


installations/src/android/installations_android.cc
```diff
@@ -186,6 +186,19 @@ void InstallationsInternal::Cleanup() {
   }
 }
 
+void InstallationsInternal::LogHeartbeat(const firebase::App& app) {
+  // Calling the native getter is sufficient to cause a Heartbeat to be logged.
+  JNIEnv* env = app.GetJNIEnv();
```
cynthiajoan, line 191: I'm curious if there will be some warning and exception thrown out while running on android. If so, probably need to follow this pattern to make sure the jni exception is cleared.

https://github.com/firebase/firebase-cpp-sdk/blob/dcb0b3b3da62b04dc869b0b40508c82238e2516b/remote_config/src/android/remote_config_android.cc#L590

installations/src/ios/installations_ios.mm
```diff
@@ -40,6 +40,12 @@
   // Destructor is necessary for ARC garbage collection.
 }
 
+void InstallationsInternal::LogHeartbeat(const firebase::App& app) {
+  // Calling the native getter is sufficient to cause a Heartbeat to be logged.
+  FIRApp *platform_app = app.GetPlatformApp();
+  [FIRInstallations installationsWithApp:platform_app];
```
cynthiajoan, line 46: Is the iOS implementation ready? From this code piece I don't think it's actually doing anything. Anything i'm missing here?

chkuang-g: The way to trigger heartbeat logging at iOS/Android level is to call their getter and I think [installationsWithApp](https://firebase.google.com/docs/reference/ios/firebaseinstallations/api/reference/Classes/FIRInstallations#/c:objc(cs)FIRInstallations(cm)installationsWithApp:) is the getter for iOS SDK.

We can pull Nick Cook in for confirmation.

installations/src/android/installations_android.cc
chkuang-g: Good catch.  We probably should do the same for Auth as well

installations/src/ios/installations_ios.mm
AlmostMatt: My understanding is that the iOS implementation of heartbeat logging is currently in a feature branch, not yet in main. So it's correct that this will initially be effectively no-op. 
Nick gave the OK on a similar pull request for auth: https://github.com/firebase/firebase-cpp-sdk/pull/846

installations/src/android/installations_android.cc
AlmostMatt: I added a call to CheckAndClearJniExceptions.
I removed the FIREBASE_ASSERT since I had just copied it literally from the above InstallationsInternal, but I dont think we need such an assertion in the LogHeartbeat method since the installations instance is not actually used by LogHeartbeat. 

I also checked the Auth change and confirmed that I do already call CheckAndClearJniExceptions in the Auth LogHeartbeat method.
https://github.com/firebase/firebase-cpp-sdk/pull/846/files

I find it a bit odd that there were not any existing calls to CheckAndClearJniExceptions in installations_android.cc, but maybe it is just because it was written longer ago.

cynthiajoan (approved):

installations/src/android/installations_android.cc
```diff
@@ -186,6 +187,19 @@ void InstallationsInternal::Cleanup() {
...
+  jobject platform_app = app.GetPlatformApp();
+  jclass installations_class = installations::GetClass();
```
jonsimantov, line 194: Is there a chance that installations_class could be null here? If so, it would be good to check and avoid calling GetInstance.

AlmostMatt: What can cause a jclass to be null? I didn't expect there to be any challenge with getting a reference to a class in order to call a static method.

This logic is roughly the same as InstallationsInternal above, though that one also does some additional cleanup or initialization because it is the first method called.

PR #855, opened 2022-02-18
(Desktop) Firestore C++ Core from TIP by default

wu-hui:


cmake/external/firestore.cmake
```diff
@@ -18,18 +18,54 @@ if(TARGET firestore)
...
+    BUILD_COMMAND ""
+    INSTALL_COMMAND ""
+    TEST_COMMAND ""
+    HTTP_HEADER "${EXTERNAL_PROJECT_HTTP_HEADER}"
+    )
+endfunction()
+
+if(NOT FIRESTORE_DEP_SOURCE)
+  # Get from tip of the tree by default
+  GetTag("origin/master")
```
jonsimantov, line 61: I would really rather not get from the tip of the tree by default - this could allow our build to break, as the build tree will get out of sync with our iOS dependencies. Having the option to get from HEAD is a good idea though, just not the default.

wu-hui: It certainly can break CPP because we did a commit to iOS, but IMHO this is better than having to discover this by the time we start preparing for a release. We want to catch breakages like this as earlier as possible, when the person who make the commit still have good context.

Ideally, we should be able to run the tests as part of iOS PR CI, which stops the PR from committing if it breaks. Right now the long build time of CPP is not helping with this..

jonsimantov: Right now, the iOS dependency version updates all at once via a workflow (example: https://github.com/firebase/firebase-cpp-sdk/pull/845). This allows us to determine at a glance (if the integration tests succeed) that the new version works. If we allow Firestore to update its version out-of-band with that, anyone's random PR could have tests fail due to the iOS SDK changing out from under them. I don't believe that is an acceptable situation.

wu-hui: As discussed offline, the default behavior now remains unchanged.

sunmou99 (approved):
LGTM

PR #897, opened 2022-04-14
Heartbeat desktop - implementation of heartbeat storage

AlmostMatt:
This is an initial implementation of storage for heartbeat desktop implementation

There are a number of TODOs for details that I am still working on, but the rough outline is there

### DIFF
--- a/app/src/heartbeat/heartbeat_storage_desktop.cc
+++ b/app/src/heartbeat/heartbeat_storage_desktop.cc
@@ -57,7 +57,7 @@ std::string GetFilename(std::string app_id, std::string* error) {
   auto app_id_without_symbols =
       std::regex_replace(app_id, std::regex("[/\\?%*:|\"<>.,;=]"), "");
   // Note: fstream will convert / to \ if needed on windows.
-  return app_dir + "/" + kHeartbeatFilenamePrefix + app_id;
+  return app_dir + "/" + kHeartbeatFilenamePrefix + app_id_without_symbols;
 }
 
 }  // namespace
@@ -79,6 +79,7 @@ HeartbeatStorageDesktop::HeartbeatStorageDesktop(std::string app_id)
 static const int kMaxBufferSize = 1024 * 500;
 
 bool HeartbeatStorageDesktop::Read(LoggedHeartbeats* heartbeats_output_ptr) {
+  // TODO(almostmatt): consider taking a reference instead of a ptr as input
   MutexLock lock(FileMutex());
   error_ = "";
 
@@ -94,7 +95,14 @@ bool HeartbeatStorageDesktop::Read(LoggedHeartbeats* heartbeats_output_ptr) {
   }
   // Current position in file is file size. Fail if file is too large.
   std::streamsize buffer_len = file.tellg();
-  if (buffer_len > kMaxBufferSize || buffer_len == -1) return false;
+  if (buffer_len > kMaxBufferSize) {
+    error_ = "'" + filename_ + "' is too large to read.";
+    return false;
+  }
+  if (buffer_len == -1) {
+    error_ = "Failed to determine size of '" + filename_ + "'.";
+    return false;
+  }
   // Rewind to start of the file, then read into a buffer on the heap.
   file.seekg(0, std::ios_base::beg);
   char* buffer = new char[buffer_len];
@@ -103,10 +111,10 @@ bool HeartbeatStorageDesktop::Read(LoggedHeartbeats* heartbeats_output_ptr) {
   ::flatbuffers::Verifier verifier(reinterpret_cast<const uint8_t*>(buffer),
                                    buffer_len);
   if (!VerifyLoggedHeartbeatsBuffer(verifier)) {
-    error_ =
-        "Failed to parse contents of " + filename_ + " as LoggedHeartbeats.";
+    // If the file is empty or contains corrupted data, return a default instance.
+    *heartbeats_output_ptr = LoggedHeartbeats();
     delete[] buffer;
-    return false;
+    return true;
   }
   const LoggedHeartbeatsFbs* heartbeats_fbs = GetLoggedHeartbeats(buffer);
   *heartbeats_output_ptr = LoggedHeartbeatsFromFbs(heartbeats_fbs);
--- a/app/src/heartbeat/heartbeat_storage_desktop.cc
+++ b/app/src/heartbeat/heartbeat_storage_desktop.cc
@@ -22,12 +22,12 @@
 
 #include "app/logged_heartbeats_generated.h"
 #include "app/src/filesystem.h"
-#include "app/src/include/firebase/internal/mutex.h"
+#include "app/src/logger.h"
 
 namespace firebase {
 namespace heartbeat {
 
-using LoggedHeartbeatsFbs =
+using LoggedHeartbeatsFlatbuffer =
     com::google::firebase::cpp::heartbeat::LoggedHeartbeats;
 using com::google::firebase::cpp::heartbeat::CreateLoggedHeartbeats;
 using com::google::firebase::cpp::heartbeat::CreateUserAgentAndDates;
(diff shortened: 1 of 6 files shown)


app/src/heartbeat/heartbeat_storage_desktop.cc
```diff
@@ -0,0 +1,125 @@
...
+
+  // LoggedHeartbeats is just part of a flatbuffer. To get the entire buffer,
+  // Create a new flat buffer builder based on the LoggedHeartbeats instance.
+  // TODO(almostmatt): This does not seem good. Find a better way to write.
+  flatbuffers::FlatBufferBuilder fbb(1024);
+  auto heartbeats_unique_ptr = UnPackLoggedHeartbeats(heartbeats);
+  auto heartbeats_offset =
+      CreateLoggedHeartbeats(fbb, heartbeats_unique_ptr.get());
+  fbb.Finish(heartbeats_offset);
+  file.write((char*)fbb.GetBufferPointer(), fbb.GetSize());
```
AlmostMatt, line 119: The logic here is not good. All of the documentation and examples I could find for writing a flatbuffer to a disk had a pointer to the buffer of bytes. But I could not find a good way to go from the LoggedHeartbeats object to the buffer pointer, only the flat buffer builder has a convenient getter for the buffer pointer.

If there is a more convenient way to get the buffer pointer from the class I would be happy to use that.

chkuang-g: @jonsimantov  Are you familiar about how this is done with Flatbuffer?

AlmostMatt: I spoke a bit with Jon about flatbuffers and serializations. Seems that it is correct that the buffer pointer and size is needed to serialize, and LoggedHeartbeats is just a part of the buffer, not the whole thing.

Since LoggedHeartbeats is not too easy to modify or write, and a byte array is too implementation-specific, I am considering changing these methods to input and output a simple C struct that contains a last-logged field and a std::map from user agent to a list of date strings. This will mean a little bit of extra code to convert between the two, but the controller code would no longer have to deal with flatbuffer builders to modify the in-memory object.
In that case, the benefit of flatbuffers for serialization is mainly just that it allows us to verify the validity of the structure when it reads a byte buffer from a file.

AlmostMatt:
There are three future improvements that I would like to make but I think this change can be reviewed as-is.
- I would like to validate the date strings when reading from file
- I would like to lock the underlying files (using flock)
- I have a test for reading corrupted data, and it currently makes an assumption about the saved filename. It would be nice if there was a cleaner way to expose saved filename in tests.

app/src/heartbeat/heartbeat_storage_desktop.cc
chkuang-g: It is good idea to free controller from handling flatbuffer.

Given that there is not change in this file, I assume your change will come in in the future?

chkuang-g (commented):
Also, did you have time to look into https://github.com/firebase/firebase-cpp-sdk/pull/897#discussion_r853552296?

app/src/heartbeat/heartbeat_storage_desktop.cc
```diff
@@ -0,0 +1,125 @@
...
+
+std::string GetFilename(std::string& error) {
+  std::string app_dir =
+      AppDataDir(kHeartbeatDir, /*should_create=*/true, &error);
+  if (app_dir.empty()) {
+    return "";
+  }
+
+  // TODO(almostmatt): does this need to be different between windows/mac/linux?
+  return app_dir + "/" + kHeartbeatFilename;
```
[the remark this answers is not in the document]
AlmostMatt: When I searched online it seemed like fstream might have internal logic to convert / and \ for windows. I haven't verified this yet, though in theory some of the added unit tests would confirm this if run on windows.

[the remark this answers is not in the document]
AlmostMatt: Ok, I have confirmed that this does indeed work.
I wrote a simple script that uses fstream to write to "C:/Users/username/Documents/filename1" and also use fstream to write to "C:\\Users\\username\\Documents\\filename2", and ran the script on a windows device. Both files were created.

app/src/heartbeat/heartbeat_storage_desktop.cc
```diff
@@ -0,0 +1,178 @@
...
+
+  if (!file) {
+    error_ = "Unable to open '" + filename_ + "'.";
+  }
+}
+
+// Size is arbitrary, just making sure that there is a sane limit.
+static const int kMaxBufferSize = 1024 * 500;
+
+bool HeartbeatStorageDesktop::ReadTo(LoggedHeartbeats* heartbeats_output) {
```
chkuang-g, line 82: http://go/cstyle#Inputs_and_Outputs

If this parameter is required, use `LoggedHeartbeats& heartbeats_output`

app/src/heartbeat/heartbeat_storage_desktop.cc
```diff
@@ -0,0 +1,178 @@
...
+const char kHeartbeatDir[] = "firebase-heartbeat";
+const char kHeartbeatFilenamePrefix[] = "heartbeats-";
+
+// Returns the mutex that protects accesses to the storage file.
+Mutex& FileMutex() {
+  static Mutex* mutex_ = new Mutex(Mutex::kModeNonRecursive);
+  return *mutex_;
+}
+
+std::string GetFilename(std::string app_id, std::string* error) {
```
chkuang-g, line 50: `const std::string& app_id`

app/src/heartbeat/heartbeat_storage_desktop.cc
```diff
@@ -0,0 +1,178 @@
...
+    delete[] buffer;
+    return true;
+  }
+  const LoggedHeartbeatsFbs* heartbeats_fbs = GetLoggedHeartbeats(buffer);
+  *heartbeats_output = LoggedHeartbeatsFromFbs(heartbeats_fbs);
+  delete[] buffer;
+  return true;
+}
+
+bool HeartbeatStorageDesktop::Write(LoggedHeartbeats heartbeats) const {
```
chkuang-g, line 121: `const LoggedHeartbeats& heartbeats` since this is for read only.

app/src/heartbeat/heartbeat_storage_desktop.cc
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  flatbuffers::FlatBufferBuilder fbb = LoggedHeartbeatsToFbs(heartbeats);
+  file.write((char*)fbb.GetBufferPointer(), fbb.GetSize());
+
+  return !file.fail();
+}
+
+LoggedHeartbeats HeartbeatStorageDesktop::LoggedHeartbeatsFromFbs(
+    const LoggedHeartbeatsFbs* heartbeats_fbs) const {
```
chkuang-g, line 139: If it is required, use `const LoggedHeartbeatsFbs&`

app/src/heartbeat/heartbeat_storage_desktop.cc
```diff
@@ -0,0 +1,178 @@
...
+    for (auto date : *(user_agent_and_dates->dates())) {
+      // TODO(almostmatt): verify format of date string
+      heartbeats_struct.heartbeats[user_agent].push_back(date->str());
+    }
+  }
+  return heartbeats_struct;
+}
+
+flatbuffers::FlatBufferBuilder HeartbeatStorageDesktop::LoggedHeartbeatsToFbs(
+    LoggedHeartbeats heartbeats_struct) const {
```
chkuang-g, line 155: `const LoggedHeartbeats&` to prevent class copy

app/src/heartbeat/heartbeat_storage_desktop.h
```diff
@@ -0,0 +1,94 @@
...
+// - If a failure occurred, GetError can be used to get a descriptive error
+// message.
+// TODO(almostmatt): consider returning distinct error codes.
+class HeartbeatStorageDesktop {
+ public:
+  explicit HeartbeatStorageDesktop(std::string app_id);
+
+  // If the previous disk operation failed, contains additional details about
+  // the error; otherwise is empty.
+  const std::string& GetError() const { return error_; }
```
chkuang-g, line 57: I am not so funded to this pattern.

1. If changing this field is guarded by a mutex, you should guard it when getting this property as well, since this can be changed from a different thread. Even this property is guarded, when the called try to access it, the error may not be relevant to what the called did.
2. It is unclear from an API level that `ReadTo()` and `Write()` would change this property and expect the called to retrieve error using a different API. 

I can think of a couple of approaches for this
1. Change API to something like `bool ReadTo(LoggedHeartbeats* heartbeats_output, std::string* error);`. This makes `error` arg optional and should be handled in the implementation.
2. Just log the error.  This can be done in a couple of ways. Each FirebaseApp instance has its own logger. https://github.com/firebase/firebase-cpp-sdk/blob/main/app/src/app_common.cc#L186. You can keep a reference to `App`, or the logger itself. Or you can create your own logger for an instance of Heartbeat. But I think former is better since it is easier to set verbosity level.

app/src/heartbeat/heartbeat_storage_desktop.cc
```diff
@@ -0,0 +1,178 @@
...
+using com::google::firebase::cpp::heartbeat::VerifyLoggedHeartbeatsBuffer;
+
+// Using an anonymous namespace for file mutex and filename
+namespace {
+
+const char kHeartbeatDir[] = "firebase-heartbeat";
+const char kHeartbeatFilenamePrefix[] = "heartbeats-";
+
+// Returns the mutex that protects accesses to the storage file.
+Mutex& FileMutex() {
```
chkuang-g, line 45: I think this mutex is better to be in class level.

That is, heartbeat storages for app1 and app2 are not necessarily writing to the same location. BUT it is also possible that they are write to the same location as well. Rare but possible.

Beside, static mutex is a sources of troubles. Since you did not write anything to destroy it, that is memory leak as well.

On the other hand, if you can guarantee that the access to this API are all serialized to the same thread, like using `schedular` class, the mutex is actually not required.

chkuang-g:
Also, failed test
https://github.com/firebase/firebase-cpp-sdk/runs/6216165938?check_suite_focus=true

app/tests/heartbeat_storage_desktop_test.cc
```diff
@@ -0,0 +1,210 @@
...
+  LoggedHeartbeats read_heartbeats;
+  bool read_ok = storage.ReadTo(&read_heartbeats);
+  ASSERT_TRUE(read_ok) << "Unable to read the heartbeat file:\n  " +
+                              storage.GetError();
+  EXPECT_EQ(read_heartbeats.last_logged_date, "");
+  ASSERT_EQ(read_heartbeats.heartbeats.size(), 0);
+}
+
+TEST_F(HeartbeatStorageDesktopTest, WriteAndReadWithSymbolsInAppId) {
+  std::string app_id = "start/\\?%*:|\"<>.,;=end";
```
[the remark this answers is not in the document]
AlmostMatt: Rather than exposing a static method that constructs a filename, I just provided a getter to the filename that is used by an instance of storage. I also updated the test to check the generated filename when the appid has symbols.

app/src/heartbeat/heartbeat_storage_desktop.h
AlmostMatt: Ok. For now I removed the mutex.
Deciding whether to rely on some new mutex, or scheduler, or just OS file locking is something that I think can be added while working on the controller.

app/src/heartbeat/heartbeat_storage_desktop.cc
AlmostMatt: Ok. For now I removed the mutex.
Deciding whether to rely on some new mutex, or scheduler, or just OS file locking is something that I think can be added while working on the controller.

app/src/heartbeat/heartbeat_storage_desktop.h
AlmostMatt: Ok following your later suggestion, I added Logger* as a parameter in the constructor for the class, and anywhere that it previously set error_, it will now call logger_->LogError.

app/src/heartbeat/heartbeat_storage_desktop.h
```diff
@@ -0,0 +1,70 @@
...
+  std::map<std::string, std::vector<std::string> > heartbeats;
+};
+
+class HeartbeatStorageDesktop {
+ public:
+  explicit HeartbeatStorageDesktop(const std::string& app_id, Logger* logger);
+
+  // Reads an instance of LoggedHeartbeats from disk into the provided struct.
+  // Returns `false` if the read operation fails.
+  bool ReadTo(LoggedHeartbeats& heartbeats_output);
```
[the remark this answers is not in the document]
AlmostMatt: When I make this output variable a non-const reference it gives a lint warning which is annoying, but I agree that the inputs/outputs documentation you linked does seem to suggest using non-const references for required output parameters.

app/src/heartbeat/heartbeat_storage_desktop.h
AlmostMatt: Another pattern for error output that I have seen is to have Read/Write return
absl::StatusOr<Type>

Then if it succeeds the return value is Type, and if it fails the return value is a Status which can have an error message (like absl::InvalidArgumentError("bad mode"))
But I don't see absl::Status being used much in Firebase C++ SDK

AlmostMatt:
I made all of the recommended minor changes like pointer vs const reference.
I added Logger* as a parameter to the storage class. Errors are logged instead of an error_ string being updated.
I added a Getter to return the filename that is used by an instance of the storage class
I removed the Mutex and associated synchronization test. How synchronization will be handled will be decided when implementing the controller.

app/src/heartbeat/heartbeat_storage_desktop.cc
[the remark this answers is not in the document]
AlmostMatt: It looks like the windows implementation of AppDataDir also uses forward slashes. I think that if we were to start using backslashes for windows implementation we would want to be consistent and update the existing AppDataDir function as well, at which point I think it would make sense to be a separate change from this heartbeats implementation.
I'm happy to do so if you would like.

app/src/heartbeat/heartbeat_storage_desktop.cc
```diff
@@ -0,0 +1,174 @@
...
+  // Remove any symbols from app_id that might not be allowed in filenames.
+  auto app_id_without_symbols =
+      std::regex_replace(app_id, std::regex("[/\\\\?%*:|\"<>.,;=]"), "");
+  // Note: fstream will convert / to \ if needed on windows.
+  return app_dir + "/" + kHeartbeatFilenamePrefix + app_id_without_symbols;
+}
+
+}  // namespace
+HeartbeatStorageDesktop::HeartbeatStorageDesktop(const std::string& app_id,
+                                                 Logger* logger)
```
chkuang-g, line 65: If `logger` is optional, you should do null-check whenever using it.

An easy solution is to make it required by using `const Logger& logger` and keep a reference to it.
Or you can implement functions like 
```
void HeartbeatStorageDesktop::LogWarning(const char* format, ...) {
if (logger_) {
  va_list list;
  va_start(list, format);
  logger_->LogWarning(format, list)
  va_end(list);
}
}
```
See  this https://github.com/firebase/firebase-cpp-sdk/blob/main/app/src/logger.cc#L23.

But I think this could be a bit too much trouble.

app/src/heartbeat/heartbeat_storage_desktop.h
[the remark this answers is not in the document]
chkuang-g: agreed.  We should update the lint check here

chkuang-g (approved):

PR #905, opened 2022-04-19
Modify SDK build Cocoapods inclusion to use frameworks instead of libraries

jonsimantov:
Upcoming release 9.0.0 will require that Cocoapods be included as frameworks, not as a libraries. This change modifies the existing CMake build to be able to build via frameworks. This requires adding a few different locations to the include paths, as the paths are different when using frameworks (and in a few cases e.g. GTMSesionFetcher, headers are no longer in the public location).

This PR also modifies the integration tests to add  "use_frameworks! :linkage => :static" to each one; the static linkage is required for tvOS, otherwise a bug in xcodebuild causes the tvOS build to accidentally import the iOS frameworks. If the xcodebuild command-line ever fixes this issue, we can remove the static linkage workaround.

***

Integration test and [cpp packaging runs](https://github.com/firebase/firebase-cpp-sdk/actions/workflows/cpp-packaging.yml?query=branch%3Abugfix%2Fpod_frameworks) will both be verified.
***

### Type of Change
***

#### Notes


jonsimantov:
@dconeybe The Windows desktop build still seems broken: https://github.com/firebase/firebase-cpp-sdk/runs/6083486670?check_suite_focus=true

Error: Command '['D:\\a\\firebase-cpp-sdk\\firebase-cpp-sdk\\ta\\auth\\it\\bin\\external\\src\\firestore-build\\Firestore\\Protos\\pyvenv\\FirestoreProtos\\Scripts\\python3.exe', '-Im', 'ensurepip', '--upgrade', '--default-pip']' returned non-zero exit status 3221225781.

a-maurice (approved):

jonsimantov:
@sunmou99 The build_testapps.py script seems to be messing something up about adding the frameworks to the tvOS builds now, with this Cocoapods change. I tried looking into what might be happening but I haven't been able to figure it out. Can you try to reproduce the issue on this branch?

sunmou99:
> @sunmou99 The build_testapps.py script seems to be messing something up about adding the frameworks to the tvOS builds now, with this Cocoapods change. I tried looking into what might be happening but I haven't been able to figure it out. Can you try to reproduce the issue on this branch?

tvOS platform shouldn't contain `CoreTelephony`:
https://github.com/google/GoogleDataTransport/search?q=coretelephony
But, we included it anyway.
`
error: CoreTelephony is not available when building for tvOS Simulator. (in target 'GoogleDataTransport-iOS' from project 'Pods')
`

jonsimantov:
> > @sunmou99 The build_testapps.py script seems to be messing something up about adding the frameworks to the tvOS builds now, with this Cocoapods change. I tried looking into what might be happening but I haven't been able to figure it out. Can you try to reproduce the issue on this branch?
> 
> tvOS platform shouldn't contain `CoreTelephony`: https://github.com/google/GoogleDataTransport/search?q=coretelephony But, we included it anyway. `error: CoreTelephony is not available when building for tvOS Simulator. (in target 'GoogleDataTransport-iOS' from project 'Pods')`
> 
> We could change the Podfile and remove the unwanted `target` it before we build the testapp

After some more research, it seems that this is a known issue with command-line *xcodebuild* using the new build system. Unfortunately, we can't use the legacy build system as it doesn't support .xcframeworks. So it seems like the best workaround is to have platform-specific Podfiles (edited on the fly by the script).

I am testing one more possible fix, setting the linkage to static - this worked one time I tested it, though it still printed some warnings.

jonsimantov:
Everything looks good except an unrelated Firestore error, gonna go ahead and merge this.

PR #889, opened 2022-04-11
Remove gnustl support and update NDK

jonsimantov:
For next major revision, remove gnustl support and update to NDK r21e.

In the SDK binary distribution, Android libraries are now in simply libs/android/ARCH, with no STL path component.
***

Running [integration tests](https://github.com/firebase/firebase-cpp-sdk/pull/889#issuecomment-1095348828) in the PR.
Running [c++ packaging workflow](https://github.com/firebase/firebase-cpp-sdk/actions/workflows/cpp-packaging.yml?query=branch%3Adeprecation%2Fremove-gnustl) on the branch and confirming Android libraries are placed in the correct location.
***

### Type of Change
Place an `x` the applicable box:
- [ ] Bug fix. Add the issue # below if applicable.
- [ ] New feature. A non-breaking change which adds functionality.
- [X] Other, such as a build process or documentation change.
***

#### Notes
- Bug fixes and feature changes require an update to the `Release Notes` section of `release_build_files/readme.md`.
- Read the contribution guidelines [CONTRIBUTING.md](https://github.com/firebase/firebase-cpp-sdk/blob/main/CONTRIBUTING.md).
- Changes to the public API require an internal API review. If you'd like to help us make Firebase APIs better, please propose your change in a feature request so that we can discuss it together.

PR #908, opened 2022-04-20
Create individual Python virtualenv's in cmake builds

dconeybe:
Use a Python virtualenv when running Python scripts during a cmake build.

This PR is a copy of https://github.com/firebase/firebase-ios-sdk/pull/9662.

This PR adds a new cmake module, `python_setup`, that defines a single function: `FirebaseSetupPythonInterpreter()`. This function will find a Python interpreter installed on the system, create a virtualenv with it, pip-install dependencies into it, and set a variable to the path of the Python executable. Then, when cmake wants to run a Python script, it uses the one in the virtualenv.

This has several benefits compared to the old design:
1. There is no need to manually pip-install the dependencies into the local Python installation.
2. The dependencies of the Python scripts are clearly specified in the "requirements" argument to the function.
3. You don't have to worry about cmake's built-in Python searching logic finding a different interpreter than intended, which does not happen to have the required dependencies installed (e.g. https://github.com/firebase/firebase-cpp-sdk/runs/6065960193)


dconeybe:
The last of the build errors are related to gradle communicating with cmake. I think those will be fixed once we upgrade the gradle plugin, which will happen after dropping gnustl support (https://github.com/firebase/firebase-cpp-sdk/pull/889). I'm going to put this PR on hold for now.

PR #911, opened 2022-04-22
Fix storage content-type on desktop, and add test.

jonsimantov:
Storage desktop: When uploading a file with custom metadata, if the metadata specifies a content-type, use that as the content-type for the HTTP request.

This should fix https://github.com/firebase/quickstart-unity/issues/1107, which is failing because of the Content-Type of the HTTP upload.

***

Add upload of image file to integration test. (In the future, this could be extended to support image resize, etc.)
***

### Type of Change
***

#### Notes


jonsimantov:
Test failure may not be related to this, but I'll investigate.

jonsimantov:
> Test failure may not be related to this, but I'll investigate.

Was related - caused by multithreaded access. Fixed.

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -230,7 +230,8 @@ Future<void> StorageReferenceInternal::DeleteLastResult() {
 // Handy utility function, since REST calls have similar setup and teardown.
 void StorageReferenceInternal::PrepareRequest(rest::Request* request,
                                               const char* url,
-                                              const char* method) {
+                                              const char* method,
+                                              const char* content_type) {
```
AlmostMatt, line 234: Now that PrepareRequest supports content_type as a parameter, it would nice to update UpdateMetadata below to make use of this parameter so that content-type is set in a consistent way.

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -404,10 +409,12 @@ Future<Metadata> StorageReferenceInternal::PutBytes(
 
 Future<Metadata> StorageReferenceInternal::PutBytesInternal(
     const void* buffer, size_t buffer_size, Listener* listener,
-    Controller* controller_out) {
+    Controller* controller_out, const char* content_type) {
   auto* future_api = future();
   auto handle = future_api->SafeAlloc<Metadata>(kStorageReferenceFnPutBytes);
-  auto send_request_funct{[&, buffer, buffer_size, listener,
+
+  std::string content_type_str = content_type ? content_type : "";
```
AlmostMatt, line 416: What's the significance of turning this into a std::string if it is going to be turned back into a c_str before it is passed into PrepareRequest?
(Same question for content_type_str in PutFileInternal)
I suppose this might keep the string in scope for the duration of retrying, but not certain.

AlmostMatt (commented):
Looks good, just a couple minor comments/questions.

storage/src/desktop/storage_reference_desktop.cc
jonsimantov: It's copying the data into a local variable - because this is an asynchronous operation, we can't rely on the string that was passed into content_type being around forever. So we make a local copy for the duration of the retries.

storage/src/desktop/storage_reference_desktop.cc
jonsimantov: Good idea.

storage/src/desktop/storage_reference_desktop.cc
jonsimantov: (To be completely clear, adding it to the lambda capture list is what causes it to be COPIED into the other thread.)

AlmostMatt (approved):

PR #921, opened 2022-05-03
GMA - Add ResponseInfo to AdResult

DellaBitta:
The GMA ResponseInfo object can be obtained on any and result, whether it's an AdError or a successful AdLoad.  However, currently the C++ API only makes this accessible on error objects.

This will allow our customers to query which mediation adapter loaded an ad.

Additional Changes:
- [iOS] fixed a problem on Load Ad operations for all ad types if the parameters passed into the method were destroyed before the ayschronous ad load operation is scheduled.
- [iOS] Renamed the methods which complete ad load futures to clarify code paths which handle errors and successes.
- A few doxygen changes based on feedback from DevRel.

***

- Local testing on iOS and Android.
- Added new tests to ensure the ResponseInfo object is populated on successful ad loads.
- Added tests where AdRequest objects leave scope, and are destroyed, for LoadAd operations.
- Added tests which pass-in empty AdRequest objects on ad loads to ensure we correctly handle the marshalling of empty parameters.
- [Integration Test CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/2264384922).

***

### Type of Change
***

### DIFF
--- a/gma/src/common/gma_common.cc
+++ b/gma/src/common/gma_common.cc
@@ -65,10 +65,11 @@ const char* kAdUninitializedErrorMessage = "Ad has not been fully initialized.";
 
 // GmaInternal
 void GmaInternal::CompleteLoadAdFuture(
-    FutureCallbackData<AdResult>* callback_data) {
+    FutureCallbackData<AdResult>* callback_data,
+    const ResponseInfoInternal& response_info_internal) {
   callback_data->future_data->future_impl.CompleteWithResult(
       callback_data->future_handle, static_cast<int>(kAdErrorCodeNone), "",
-      AdResult());
+      AdResult(ResponseInfo(response_info_internal)));
   delete callback_data;
 }
 
@@ -100,9 +101,13 @@ AdInspectorClosedListener::~AdInspectorClosedListener() {}
 AdResult::AdResult() : is_successful_(true) {}
 AdResult::AdResult(const AdError& ad_error)
     : is_successful_(false), ad_error_(ad_error) {}
+AdResult::AdResult(const ResponseInfo& response_info)
+    : is_successful_(true), response_info_(response_info) {}
+
 AdResult::~AdResult() {}
 bool AdResult::is_successful() const { return is_successful_; }
 const AdError& AdResult::ad_error() const { return ad_error_; }
+const ResponseInfo& AdResult::response_info() const { return response_info_; }
 
 // AdSize
 // Hardcoded values are from publicly available documentation:
--- a/gma/src/include/firebase/gma/types.h
+++ b/gma/src/include/firebase/gma/types.h
@@ -397,7 +397,7 @@ class AdapterResponseInfo {
   /// occurred, and any contextual information about that error.
   ///
   /// @return the error that occurred while rendering the ad.  If no error
-  /// occurred then the AdResults's successful method will return true.
+  /// occurred then the AdResults' successful method will return true.
   AdResult ad_result() const { return ad_result_; }
 
   /// Returns a string representation of a class name that identifies the ad
--- a/gma/src/include/firebase/gma/types.h
+++ b/gma/src/include/firebase/gma/types.h
@@ -397,7 +397,7 @@ class AdapterResponseInfo {
   /// occurred, and any contextual information about that error.
   ///
   /// @return the error that occurred while rendering the ad.  If no error
-  /// occurred then the AdResults' successful method will return true.
+  /// occurred then the AdResult's successful method will return true.
   AdResult ad_result() const { return ad_result_; }
 
   /// Returns a string representation of a class name that identifies the ad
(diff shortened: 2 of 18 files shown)


gma/src/include/firebase/gma/types.h
```diff
@@ -234,6 +235,45 @@ class AdError {
   AdErrorInternal* internal_;
 };
 
+/// Information about an ad response.
+class ResponseInfo {
```
DellaBitta, line 239: Moved this class up higher so it can be defined before it's used as a member of AdResult.

gma/src/include/firebase/gma/types.h
```diff
@@ -234,6 +235,45 @@ class AdError {
...
+  /// Gets the response ID string for the loaded ad.  Returns an empty
+  /// string if the ad fails to load.
+  const std::string& response_id() const { return response_id_; }
+
+  /// Gets a log friendly string version of this object.
+  const std::string& ToString() const { return to_string_; }
+
+ private:
+  friend class AdError;
+  friend class GmaInternal;
```
DellaBitta, line 267: Adding GmaInternal as a friend is the only real change here.

gma/src/ios/gma_ios.mm
```diff
@@ -296,21 +295,20 @@ void Terminate() {
...
-// Constructs AdResult objects based on the encountered result, beit a
-// successful result, and C++ SDK Wrapper error, or an error returned from
-// the iOS GMA SDK.
-void CompleteAdResult(FutureCallbackData<AdResult>* callback_data,
-                      NSError *error, bool is_load_ad_error,
-                      AdErrorCode error_code,
-                      const char* error_message) {
+// Constructs AdResult objects based on the encountered error, beit one that
+// originated from C++ SDK Wrapper or an error returned from the iOS GMA SDK.
+void CompleteAdResultError(FutureCallbackData<AdResult>* callback_data,
```
DellaBitta, line 300: Reworked this function and it's now invoked only in error conditions.

gma/src/ios/gma_ios.mm
```diff
@@ -325,43 +323,51 @@ void CompleteAdResult(FutureCallbackData<AdResult>* callback_data,
       ad_error_internal.ad_error_type =
         AdErrorInternal::kAdErrorInternalAdError;
     }
-  } else if (ad_error_internal.code != kAdErrorCodeNone) {
+  } else {
+    FIREBASE_ASSERT_MESSAGE(ad_error_internal.code != kAdErrorCodeNone,
```
jonsimantov, line 327: When would this assert ever be triggered? if error_code is none and error = nil?

gma/src/ios/ad_view_internal_ios.mm
```diff
@@ -51,12 +52,18 @@
       kAdAlreadyInitializedErrorMessage, future_handle, &future_data_);
   } else {
     initialized_ = true;
+
+    // Guard against parameter object destruction before the async operation
+    // executes (below).
+    AdParent local_ad_parent = parent;
+    std::string local_ad_unit_id(ad_unit_id);
```
jonsimantov, line 59: nit: You could just skip the middleman here and NSString *local_ad_unit_id = @(ad_unit_id), same elsewhere. Saves a string copy.

gma/src/common/gma_common.cc
```diff
@@ -65,10 +65,11 @@ const char* kAdUninitializedErrorMessage = "Ad has not been fully initialized.";
 
 // GmaInternal
 void GmaInternal::CompleteLoadAdFuture(
```
jonsimantov, line 67: Since you renamed some of the other functions Complete\*Success and Complete\*Error, WDYT about adding Success to the name of this and Error to the name of the subsequent method, rather than rely on overloading? Just a light suggestion.

jonsimantov (approved):
Approved with light nits/suggestions.

gma/src/ios/gma_ios.mm
DellaBitta: Yep.  Basically just to ensure that those who call this function fill in the correct parameters in the future, and know immediately if they're misusing it.

gma/src/include/firebase/gma/types.h
```diff
@@ -255,13 +295,29 @@ class AdResult {
   /// information.
   const AdError& ad_error() const;
 
+  /// For debugging and logging purposes, successfully loaded ads provide a
+  /// ResponseInfo object which contains information about the adapter which
+  /// loaded the ad. If the ad failed to load then the object returned from
+  /// this method will have default values. Information about the error
```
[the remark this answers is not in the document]
DellaBitta: Hm, maybe.  We basically instruct people to look at this only on AdResult success, so that should be enough. Let me think about adding the is_default() in a different PR.

gma/src/common/gma_common.cc
DellaBitta: OK, will followup in a subsequent change.

PR #946, opened 2022-05-10
Fix inconsistent feature request issue prefix

cpsauer:
Hey wonderful Firebase folks,

I'd noticed that the Feature Request template added [FR] to the title but then asked that the title instead start with "FR:". So, I figured I'd toss up a quick PR as a concrete proposal of one way to fix the inconsistency.

Figured you all would at least want a friendly heads. [FR] and no further description seemed more consistent with the other templates, so that's what I did in the PR. But of course, it doesn't matter to me what you choose. 

Cheers!
Chris
(ex-Googler)


jonsimantov (approved):
Looks good, thanks for the change!

PR #950, opened 2022-05-10
DO NOT MERGE: firestore.cmake: point at HEAD commit in master branch: https://githu…

dconeybe:
Test PR to see if the build works at the HEAD of the iOS SDK. I will close this PR once the "testing" is done.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -214,6 +214,9 @@ endif()
 # firebase-ios-sdk since it's not needed and can sometimes fail to build.
 set(FIRESTORE_INCLUDE_OBJC OFF CACHE BOOL "Disabled for the CPP SDK")
 
+# Disable re2 build tests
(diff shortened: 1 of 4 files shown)


dconeybe:
@wu-hui Thanks for taking this over from me. For the Android build error, it looks like it may not be nicely handling the modifications of `cmake/external/firestore.cmake` that pin the dependency on firebase-ios-sdk to a specific commit. I have no idea why the `windows-latest-Release-x64-static` workflow is failing to find the firestore ctest targets. Enjoy!

dconeybe:
Now that Snappy support has been added to the iOS SDK (https://github.com/firebase/firebase-ios-sdk/pull/9596), building at head is quite a bit more complicated. This would should instead be continued in https://github.com/firebase/firebase-cpp-sdk/pull/932, which does the work to modify this repo to no longer patch in Snappy support.

PR #953, opened 2022-05-12
Handle jniresultcallback not loading due to lack of gms.Task.

jonsimantov:
app/src/util_android.h/.cc is a repository of many useful C++ functions to do Android/JNI things. Nearly every SDK we have calls util::Initialize() and util::Terminate() in its Android implementation, so that they can take advantage of the util_android functionality.

Unfortunately, although util_android mostly just utilizes basic Java and Android classes, there is one embedded class—JniResultCallback—that uses com.google.android.gms.Task, which is provided by com.google.android.gms:play-services-tasks.

This PR makes the initialization for JniResultHelper optional. If JniResultHelper failed to initialize, it will not be Terminated later.

Additionally, Firebase App was previously forcing the GoogleApiAvailability class to be initialized. Again, if play-services-base is not included, this will fail; this PR adds handling to gracefully ignore its failure, and (if the user tries to use ModuleInitializer without play-services-base included) return a detailed error.

This PR prevents apps that only use Firebase SDKs that don't use gms.Task under the hood—e.g. Analytics, Remote Config, Messaging—from crashing unless they include an explicit dependency on play-services-tasks (or play-services-base), because the class cannot be found.

This PR also adds a note to the readme about using the gradle dependency file, and adds the capability to include app *without* the play-services-base dependency, if the user wishes.

(Note, once this PR goes in, the gradle example added to the Readme should be modified from admob to gma.

***

Integration test [here](https://github.com/firebase/firebase-cpp-sdk/actions/runs/2315893460) was originally run with play-services-base not being included in any builds, to ensure it is never required. Since then, only Remote Config and AdMob have been modified to use appWithoutPlayServices.
***

### Type of Change
***

#### Notes


app/src/util.cc
```diff
@@ -77,32 +79,47 @@ static void PerformInitialize(ModuleInitializerData* data) {
...
-                LogInfo("Google Play services now available, continuing.");
-                PerformInitialize(data);
-              } else {
-                LogError("Google Play services still unavailable.");
-                int num_remaining = data->init_fns.size() - data->init_fn_idx;
-                data->future_impl.Complete(
-                    data->future_handle_init, num_remaining,
-                    "Unable to initialize due to missing Google Play services "
-                    "dependency.");
+      if (google_play_services::Initialize(data->app->GetJNIEnv(),
```
a-maurice, line 82: Should this be checking g_initialized_google_play_services?  It seems odd to be Initializing here, especially since nothing cleans up after this call, unlike the other one.

jonsimantov: Added a comment to explain that it's Terminated in the OnCompletion below.

a-maurice (approved):

PR #959, opened 2022-05-13
admob 2021 merge main with cleanup

jonsimantov:
[replace this line]: # (Describe your changes in detail.)
***

[replace this line]: # (Describe your testing in detail.)
***

### Type of Change
Place an `x` the applicable box:
- [ ] Bug fix. Add the issue # below if applicable.
- [ ] New feature. A non-breaking change which adds functionality.
- [ ] Other, such as a build process or documentation change.
***

#### Notes
- Bug fixes and feature changes require an update to the `Release Notes` section of `release_build_files/readme.md`.
- Read the contribution guidelines [CONTRIBUTING.md](https://github.com/firebase/firebase-cpp-sdk/blob/main/CONTRIBUTING.md).
- Changes to the public API require an internal API review. If you'd like to help us make Firebase APIs better, please propose your change in a feature request so that we can discuss it together.

PR #960, opened 2022-05-16
Handle Access Denied error when cleaning up build directory on Windows

jonsimantov:
Some of the build artifacts on Windows end up as read-only files, which shutil.rmtree doesn't handle by default. Pass in an onerror handler to make the files writable and try again.

This causes the disk space usage of build_testapps on Windows desktop to drop from 11GB+ to 2.5GB.

***

Tested in #959 and in this PR's integration tests.
***

### Type of Change
***

#### Notes

### DIFF
--- a/scripts/gha/build_testapps.py
+++ b/scripts/gha/build_testapps.py
@@ -79,6 +79,7 @@ import json
 import os
 import platform
 import shutil
+import stat
 import subprocess
 import sys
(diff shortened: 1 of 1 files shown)


sunmou99 (approved):
This is an interesting solution. I saw similar problems before, and I just added `os.chmod` before the code that is causing the problem.

PR #961, opened 2022-05-19
Add InitiateOnDeviceConversion method to Analytics

AlmostMatt:
### Description
Adds InitiateOnDeviceConversionMeasurementWithEmail to the Analytics API as described in go/ga4f-odm-api
On iOS this just passes through to the corresponding method in the iOS SDK
On Android this is no-op.
***
### Testing
Added unit and integration tests that call the function. No expectations other than that the relevant function is called since all functional logic is in underlying iOS SDK.

Manually ran integration test that calls thew new method to verify that the underlying iOS SDK logs "Failed to initiate on-device conversion measurement. GoogleAppMeasurementOnDeviceConversion dependency not linked in." if the relevant dependency is not linked in, and that the same error message is not logged if the dependency is linked. This verifies that the relevant code is being run and that the code compiles whether or not the dependency is present.
***

### Type of Change
***

#### Notes

### DIFF
--- a/analytics/integration_test/src/integration_test.cc
+++ b/analytics/integration_test/src/integration_test.cc
@@ -108,6 +108,9 @@ TEST_F(FirebaseAnalyticsTest, TestSetProperties) {
       firebase::analytics::kUserPropertySignUpMethod, "Google");
   // Set the user ID.
   firebase::analytics::SetUserId("my_integration_test_user");
+  // Initiate on-device conversion measurement.
+  firebase::analytics::InitiateOnDeviceConversionMeasurementWithEmailAddress(
+      "my_email@site.com");
 }
 
 TEST_F(FirebaseAnalyticsTest, TestLogEvents) {
--- a/analytics/src/include/firebase/analytics.h
+++ b/analytics/src/include/firebase/analytics.h
@@ -462,7 +462,7 @@ void LogEvent(const char* name, const Parameter* parameters,
 /// (no-op on Android). On iOS, requires dependency
 /// GoogleAppMeasurementOnDeviceConversion to be linked in, otherwise it is a
 /// no-op.
-/// @param emailAddress User email address. Include a domain name for all email
+/// @param email_address User email address. Include a domain name for all email
 /// addresses
 ///   (e.g. gmail.com or hotmail.co.jp). Remove any spaces in between the email
 ///   address.
--- a/analytics/src/include/firebase/analytics.h
+++ b/analytics/src/include/firebase/analytics.h
@@ -464,8 +464,7 @@ void LogEvent(const char* name, const Parameter* parameters,
 /// no-op.
 /// @param[in] email_address User email address. Include a domain name for all
 /// email addresses
-///   (e.g. gmail.com or hotmail.co.jp). Remove any spaces in between the email
-///   address.
+///   (e.g. gmail.com or hotmail.co.jp).
 void InitiateOnDeviceConversionMeasurementWithEmailAddress(
     const char* email_address);
 
--- a/analytics/src/include/firebase/analytics.h
+++ b/analytics/src/include/firebase/analytics.h
@@ -462,8 +462,8 @@ void LogEvent(const char* name, const Parameter* parameters,
 /// (no-op on Android). On iOS, requires dependency
 /// GoogleAppMeasurementOnDeviceConversion to be linked in, otherwise it is a
 /// no-op.
-/// @param[in] email_address User email address. Include a domain name for all email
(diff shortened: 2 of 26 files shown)


analytics/integration_test/src/integration_test.cc
```diff
@@ -108,6 +108,9 @@ TEST_F(FirebaseAnalyticsTest, TestSetProperties) {
       firebase::analytics::kUserPropertySignUpMethod, "Google");
   // Set the user ID.
   firebase::analytics::SetUserId("my_integration_test_user");
+  // Initiate on-device conversion measurement.
+  firebase::analytics::InitiateOnDeviceConversionMeasurementWithEmailAddress(
+      "my_email@site.com");
```
DellaBitta, line 113: This isn't used. I'd recommend adding the Analytics dependency to the integration test and actually invoke the method in this test to ensure our C++ version links.

analytics/src/include/firebase/analytics.h
```diff
@@ -456,7 +457,17 @@ void LogEvent(const char* name);
 /// array.
 void LogEvent(const char* name, const Parameter* parameters,
               size_t number_of_parameters);
-// clang-format on
+
+/// Initiates on-device conversion measurement given a user email address on iOS
+/// (no-op on Android). On iOS, requires dependency
+/// GoogleAppMeasurementOnDeviceConversion to be linked in, otherwise it is a
+/// no-op.
+/// @param email_address User email address. Include a domain name for all email
```
DellaBitta, line 465: @param [in]

analytics/src/include/firebase/analytics.h
```diff
@@ -456,7 +457,17 @@ void LogEvent(const char* name);
...
+/// addresses
+///   (e.g. gmail.com or hotmail.co.jp). Remove any spaces in between the email
+///   address.
```
DellaBitta, line 468: What does it mean to remove spaces in between email address?

release_build_files/readme.md
```diff
@@ -603,6 +603,8 @@ code.
     - General (Android): Fixed a bug that required Android apps to include
       `com.google.android.gms:play-services-base` as an explicit dependency when
       only using AdMob, Analytics, Remote Config, or Messaging.
+    - Added [on-device conversion measurement](https://support.google.com/google-ads/answer/12119136)
```
DellaBitta, line 606: This should be tagged with the Analytics (and subcategory of iOS) APIs.

Something like `Analytics (iOS):`.  Look for something similar in this file.

DellaBitta (changes requested):

analytics/src/include/firebase/analytics.h
DellaBitta: Ok, well I wouldn't want to hold up the merge of this waiting or a reply. For now you could mirror the public iOS documentation here, and then update it later in a new PR when you have clarity and it needs to be documented.

https://firebase.google.com/docs/reference/ios/firebaseanalytics/api/reference/Classes/FIRAnalytics#+initiateondeviceconversionmeasurementwithemailaddress:

analytics/integration_test/src/integration_test.cc
AlmostMatt: Just to clarify, are you suggesting that I
1) add FirebaseAnalyticsOnDeviceConversion to the Podfile corresponding to the integration test, right?
2) call both the analytics.h method InitiateOnDeviceConversionMeasurementWithEmailAddress and also call the native objective C method [FIRAnalytics initiateOnDeviceConversionMeasurementWithEmailAddress:@(email_address)]

DellaBitta: 1.  Does the integration test fail at all if you don't include the dependency in the podfile?  Can you ensure that our underlying C++ SDK is properly linked without the adding the dependency? 

2. No, I don't expect you to call the Objective-C method in the integration test.  Only the C++ Api.

The goal is just to ensure that the new method is accessible for our users.  If you call a method that isn't in the lib then the iTest will produce a link error on build, which is a test in of itself. 

I don't think there's anything that you can query from the iOS SDK to ensure that the iTest called the Objective-C version, though.

analytics/src/include/firebase/analytics.h
AlmostMatt: I asked about that part of the comment, the reply was that they agreed it did not make sense and removed it from iOS comment (https://critique.corp.google.com/cl/450085932)
So I removed that part of this comment as well.

DellaBitta (approved):

PR #966, opened 2022-05-25
Firestore TransactionOptions added, to specify maxAttempts

dconeybe:
This is a port of https://github.com/firebase/firebase-ios-sdk/pull/9838 and https://github.com/firebase/firebase-android-sdk/pull/3664, which adds the new `TransactionOptions` class to Firestore, which can be used to specify the maximum number of attempts to run a transaction before giving up. Previously, this maximum was hardcoded to 5.


dconeybe:
The work here is done; however, I cannot proceed until the iOS and Android dependencies are upgraded to versions that include the new "transaction options" APIs. ETA for this is late June or Early July 2022.

firestore/src/common/firestore.cc
```diff
@@ -309,13 +309,20 @@ WriteBatch Firestore::batch() const {
...
+    TransactionOptions options,
+    std::function<Error(Transaction&, std::string&)> update) {
   if (!update) {
     SimpleThrowInvalidArgument(
         "Transaction update callback cannot be an empty function.");
   }
 
   if (!internal_) return FailedFuture<void>();
-  return internal_->RunTransaction(firebase::Move(update));
+  return internal_->RunTransaction(firebase::Move(update),
```
wu-hui, line 324: `internal_->RunTransaction` should take the option instead of just `max_attempts`. It's OK to choose to do it in the future however, your choice.

wu-hui (approved):
LGTM!

firestore/src/common/firestore.cc
dconeybe: I had originally coded it with an "internal" version of TransactionOptions actually (e.g. https://github.com/firebase/firebase-cpp-sdk/pull/966/commits/20240d6850dbac8caf2714258dc48ff241542329). However, it ended up just being a lot of boilerplate. I made the call to just take `max_attempts` with the idea that if we ever add another transaction option we can revisit the idea of making an "internal" version of TransactionOptions. Moreover, the Core C++ RunTransaction API (https://github.com/firebase/firebase-ios-sdk/blob/89c38d1ed908dbc10d5f9f7aded4cf271113773f/Firestore/core/src/api/firestore.h#L98-L100) just takes a `max_attempts` instead of a TransactionOptions object. So at some point the C++ SDK has to throw away the TransactionOptions object and just pass max_attempts anyways.

dconeybe:
Everything looks good on desktop and iOS (https://github.com/firebase/firebase-cpp-sdk/actions/runs/2553982778).

Need to test Android once this repository's dependency on `com.google.firebase:firebase-firestore` is upgraded to 24.2.0 or later (BOM version 30.2.0 or later). https://github.com/firebase/firebase-cpp-sdk/pull/1007 is the currently-pending PR that will do this.

dconeybe:
Android dependencies have now been upgraded: https://github.com/firebase/firebase-cpp-sdk/pull/1007

I'm going to kick off integration tests to include Android now.

wu-hui (approved):

PR #967, opened 2022-05-25
Support git longpaths

wu-hui:


wu-hui:
Build passed, but there are some test failures and at a glance they seem to be related to the python script you are working on @dconeybe ?

.github/workflows/android.yml
```diff
@@ -62,6 +62,10 @@ jobs:
           git config --global credential.helper 'store --file /tmp/git-credentials'
           echo 'https://${{ github.token }}@github.com' > /tmp/git-credentials
 
+      - name: Support Longpaths
```
dconeybe, line 65: ultranit: Mention "Git" in the name here, just so it makes more sense when someone sees this out of context.

e.g. "Support Longpaths" -> "Enable Git Longpaths Support"

dconeybe:
Something is definitely going awry with python_setup.cmake. I'm able to reproduce on my linux box too. I'm looking into it.

dconeybe:
https://github.com/firebase/firebase-ios-sdk/pull/9847 fixes the python_setup.cmake bug. I'm going to change this branch to point to the commit with the fix and see if that fixes the problems.

dconeybe:
Windows build errors appear to be due to the use of designated initializers:

```
Firestore\core\src\model\value_util.cc(60,9): error C7555: use of designated initializers requires at least '/std:c++20'
Firestore\core\src\model\value_util.cc(58,5): error C7555: use of designated initializers requires at least '/std:c++20'
Firestore\core\src\model\value_util.cc(65,5): error C7555: use of designated initializers requires at least '/std:c++20'
Firestore\core\src\model\value_util.cc(73,5): error C7555: use of designated initializers requires at least '/std:c++20'
```

https://github.com/firebase/firebase-ios-sdk/blob/4c2c86d942c1cde9a53a3bad4cdcb0aafd31032b/Firestore/core/src/model/value_util.cc#L60
https://github.com/firebase/firebase-ios-sdk/blob/4c2c86d942c1cde9a53a3bad4cdcb0aafd31032b/Firestore/core/src/model/value_util.cc#L58
https://github.com/firebase/firebase-ios-sdk/blob/4c2c86d942c1cde9a53a3bad4cdcb0aafd31032b/Firestore/core/src/model/value_util.cc#L65
https://github.com/firebase/firebase-ios-sdk/blob/4c2c86d942c1cde9a53a3bad4cdcb0aafd31032b/Firestore/core/src/model/value_util.cc#L73

I'll fix this in the iOS SDK and try again.

dconeybe:
https://github.com/firebase/firebase-ios-sdk/pull/9851 fixes the problematic usage of designated initialization in `value_util.cc` noted in the previous comment. I'm going to update this PR to point at the commit with that fix.

.github/workflows/android.yml
wu-hui: Done.

PR #972, opened 2022-05-27
bundle_test.cc: Disable LoadBundleWithDocumentsAlreadyPulledFromBackend

dconeybe:
This test is very flaky, so I'm disabling it for now.

Googlers see b/233751585 for details.


dconeybe:
This fix was a red herring. The real fix was to fix the JDK version required by the Firestore emulator: the GitHub runners had JDK 8 and JDK 11 was required.

PR #984, opened 2022-06-07
GMA iTest NoFill tetry and loadAd tests disabled on Simulator

DellaBitta:
Numerous updates to the GMA Integration Tests to increase CI stability:
- Added detection of NoFill and BackOff errors sent by the AdMob service.  The iTest will pause and retry the add load in these situations.  The maximum number of retires has been limited to 15.
- Added a `SKIP_TEST_ON_SIMULATOR` to the Firebase App Test Framework.
- Due to the higher chance of NoFill errors occuring when running on local simulators, all LoadAd operations are skipped on simulators.

***

Local Tests on Android / iOS Simulators, and Android / iOS real devices.
[Integration Test CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/2455731843)
[Packaging CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/2455732913) - [Integration Test CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/2456473095)

***

### Type of Change
***

### DIFF
--- a/gma/integration_test/src/integration_test.cc
+++ b/gma/integration_test/src/integration_test.cc
@@ -243,6 +243,11 @@ void LogAdResultIfError(const firebase::gma::AdResult* ad_result) {
   }
 }
 
+// Attempts to load an AdView configured with the provided listeners.
+// If the service returns a NoFill error, then this function will continue to
+// reattempt the load ad until kMaxNoFillRetries is reached. If
+// kMaxNoFillRetries has been reached, either via this invocation or
+// preveiously, then this function returns the AdResult with the loadAd failure.
 firebase::gma::AdView* loadAdView(
     const firebase::gma::AdSize& ad_size,
     const firebase::gma::AdRequest& request,
@@ -280,6 +285,11 @@ firebase::gma::AdView* loadAdView(
   return ad_view;
 }
 
+// Attempts to load an AdView. If the service returns a NoFill error, then
+// this function will continue to reattempt the load ad until kMaxNoFillRetries
+// is reached. If kMaxNoFillRetries has been reached, either via this invocation
+// or preveiously, then this function returns the AdResult with the loadAd
+// failure.
 firebase::gma::AdView* loadAdView(
     const firebase::gma::AdSize& ad_size,
     const firebase::gma::AdRequest& request,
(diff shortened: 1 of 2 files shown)


sunmou99 (commented):
Not Sure if it's possible to add `SKIP_TEST_ON_SIMULATOR` into the `constructor (e.g. FirebaseGmaTest)` or `GetAdRequest`.

gma/integration_test/src/integration_test.cc
```diff
@@ -208,6 +232,141 @@ void InitializeGma(firebase::App* shared_app) {
...
+// preveiously, then this function returns the AdResult with the loadAd failure.
+firebase::gma::AdView* loadAdView(
+    const firebase::gma::AdSize& ad_size,
+    const firebase::gma::AdRequest& request,
+    firebase::Future<firebase::gma::AdResult>& load_ad_future,
+    firebase::gma::AdListener* ad_listener,
+    firebase::gma::PaidEventListener* paid_event_listener) {
+  firebase::gma::AdView* ad_view = nullptr;
+
+  do {
```
jonsimantov, line 259: This seems like something that would be ripe to use RunFlakyBlock on rather than having the retry logic replicated here.

Something like...
bool ad_load_successful = RunFlakyBlock([&]() {
  load_ad_future =  ad_view->LoadAd(request);
  FirebaseGmaTest::WaitForCompletionAnyResult(load_ad_future, "LoadAd");
  if (load_ad_future.error() == firebase::gma::kAdErrorCodeNoFill || load_ad_future.error() == firebase::gma::kAdErrorCodeInvalidRequest) {
    // do whatever setup you need to retry the LoadAd request
    return false;  // will trigger a retry
  }
  return (load_ad_future.error() == 0);
}, "LoadAd");

This will do a simple exponential backoff retry until the bool returns true or it fails enough times.

DellaBitta:
> Not Sure if it's possible to add `SKIP_TEST_ON_SIMULATOR` into the `constructor (e.g. FirebaseGmaTest)` or `GetAdRequest`.

We don't want to skip all FirebaseGmaTests. Some are valid tests that don't attempt to load ads.
And not all loadAd tests use the GetAdRequest, as they use empty / non configured AdRequest objects created on the stack using the default constructor.

gma/integration_test/src/integration_test.cc
DellaBitta: I'll give this a shot on Wednesday.

PR #985, opened 2022-06-09
Add SkipTestsOnSimulator to GMA integration test

DellaBitta:
Numerous updates to the GMA Integration Tests to increase CI stability:
- Added a `SKIP_TEST_ON_SIMULATOR` to the Firebase App Test Framework.
- Due to the higher chance of NoFill errors occuring when running on local simulators, all LoadAd operations are skipped on simulators.
- Added a FLAKY_TEST_SECTION block around all tests the invoke loadAds except for the UI driven tests.
- Some tests that do extensive testing after loadAd will not fail early instead of attempting more ad operations (like show, set position, hide, etc).

***

- Local Tests on Android / iOS Simulators, and Android / iOS real devices.
- [Integration Test CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/2474839215)
- [Integration Test CI #2](https://github.com/firebase/firebase-cpp-sdk/actions/runs/2477219715) after merging-in feature branch.
- [Packaging CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/2474840393) -> [Integration Test CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/2475496400)

***

### Type of Change
***


gma/integration_test/src/integration_test.cc
```diff
@@ -1238,166 +1288,182 @@ TEST_F(FirebaseGmaTest, TestAdView) {
...
-            bounding_box_listener.bounding_box_changes_.size());
-
-  // Try hiding and showing the AdView.
-  WaitForCompletion(ad_view->Hide(), "Hide 1");
-  PauseForVisualInspectionAndCallbacks();
-  EXPECT_EQ(expected_num_bounding_box_changes,
-            bounding_box_listener.bounding_box_changes_.size());
+  // Suppress the extensive testing below if the ad failed to load.
+  if (ad_loaded) {
+    EXPECT_EQ(ad_view->ad_size().width(), kBannerWidth);
```
DellaBitta, line 1297: This section didn't diff properly. The internals of this if block are the same as the original, but they're now indented and reformatted accordingly.

gma/integration_test/src/integration_test.cc
```diff
@@ -1238,166 +1288,182 @@ TEST_F(FirebaseGmaTest, TestAdView) {
...
+  // Clean up the ad object.
+  load_ad_future.Release();
   WaitForCompletion(ad_view->Destroy(), "Destroy AdView");
   ad_view->SetBoundingBoxListener(nullptr);
   delete ad_view;
 
   PauseForVisualInspectionAndCallbacks();
+
+  if (ad_loaded) {
+    // If the ad was show, do the final bounding box checks after the ad has
```
DellaBitta, line 1427: Same here.

gma/integration_test/src/integration_test.cc
```diff
@@ -768,6 +768,9 @@ TEST_F(FirebaseGmaTest, TestRequestConfigurationSetGet) {
 // ensure that we can load them before diving into the interactive tests.
 TEST_F(FirebaseGmaTest, TestAdViewLoadAd) {
   SKIP_TEST_ON_DESKTOP;
+  SKIP_TEST_ON_SIMULATOR;
+
+  FLAKY_TEST_SECTION_BEGIN();
 
   const firebase::gma::AdSize banner_ad_size(kBannerWidth, kBannerHeight);
   firebase::gma::AdView* ad_view = new firebase::gma::AdView();
```
jonsimantov, line 776: Does AdView let you call Load again if the first load fails? If so, I think it would be better to restrict the flaky section to just the LoadAd / WaitForCompletion. I know we can't do that with Interstitial and Rewarded, but with Banners we should be able to, no?

gma/integration_test/src/integration_test.cc
```diff
@@ -821,13 +826,15 @@ TEST_F(FirebaseGmaTest, TestInterstitialAdLoad) {
...
+  FLAKY_TEST_SECTION_END();
 }
 
 TEST_F(FirebaseGmaTest, TestRewardedAdLoad) {
   SKIP_TEST_ON_DESKTOP;
+  SKIP_TEST_ON_SIMULATOR;
 
-  // Note: while showing an ad requires user interaction (below),
-  // we test that we can simply load an ad first.
+  FLAKY_TEST_SECTION_BEGIN();
```
jonsimantov, line 837: It would be good, here and elsewhere, to explain why this may be flaky. i.e. that the backend may throttle us or not be able to fill the ad.

jonsimantov (changes requested):

gma/integration_test/src/integration_test.cc
DellaBitta: Done.

gma/integration_test/src/integration_test.cc
DellaBitta: Done.

jonsimantov (approved):

PR #991, opened 2022-06-14
Add an API flag to restore_secrets script

DellaBitta:
Whenever I do local development I run restore_secrets, which updates all of the Google Services data for all of the products. This pollutes my local branch since normally I'm only working on one product.

This change adds an optional `api` flag so that you can checkout the secrets for only one product at a time.

***

- Local testing.
- [Integration Test CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/2502019096)

***

### Type of Change
***

### DIFF
--- a/scripts/gha/restore_secrets.py
+++ b/scripts/gha/restore_secrets.py
@@ -24,8 +24,8 @@ python restore_secrets.py --passphrase_file [--repo_dir <path_to_repo>]
 --passphrase_file: Specify a file to read the passphrase from (only reads the
     first line). Use "-" (without quotes) for stdin.
 --repo_dir: Path to C++ SDK Github repository. Defaults to current directory.
---api: Specify a particular product API and retrieve only that api's
-    secret.
+--apis: Specify a list of particular product APIs and retrieve only their
+    secrets.
 
 This script will perform the following:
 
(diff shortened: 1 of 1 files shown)


scripts/gha/restore_secrets.py
```diff
@@ -51,6 +53,7 @@
 flags.DEFINE_string("passphrase_file", None,
                     "Path to file with passphrase. Use \"-\" (without quotes) for stdin.")
 flags.DEFINE_string("artifact", None, "Artifact Path, google-services.json will be placed here.")
+flags.DEFINE_string("api", None, "Retrieve secret for the particular SDK only.")
```
sunmou99, line 56: Good idea. And what about make it a list of api?
`flags.DEFINE_list("apis", None, "Retrieve secret for the particular SDKs only.")`

DellaBitta: Is there a usecase for this?  Generally we're only working on one API/SDK at a time, or all at the same time in the case of CI.

sunmou99: e.g. When someone tries to build only the failed testapps locally.

sunmou99 (approved):

scripts/gha/restore_secrets.py
DellaBitta: Done!  PTAL!

PR #997, opened 2022-06-16
Implement Desktop Heartbeat Controller LogHeartbeat

AlmostMatt:
### Description
Defined heartbeat desktop controller.
Implements LogHeartbeat.
Adds tests to verify successful logging of heartbeats.
***
### Testing
Added unit tests.
Using the heartbeat desktop storage class directly to verify successsful logging. Using a mock date provider in order to test cases related to same date or older date.
***

### Type of Change
***

#### Notes


app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -0,0 +1,74 @@
...
+    if (logged_heartbeats.last_logged_date != current_date) {
+      logged_heartbeats.last_logged_date = current_date;
+      logged_heartbeats.heartbeats[user_agent].push_back(current_date);
+      // Don't store more than 30 days for the same user agent
+      if (logged_heartbeats.heartbeats[user_agent].size() > 30) {
+        logged_heartbeats.heartbeats[user_agent].erase(
+          logged_heartbeats.heartbeats[user_agent].begin());
+      }
+    }
+    bool write_succeeded = this->storage_.Write(logged_heartbeats);
```
chkuang-g, line 63: I think we need to have a way to lock the file after read and before write to prevent race. Here are two scenarios:

- In the same application instance, the user created two `FirebaseApp` with slightly different `AppConfig`, ex. database url, but still using the same `app_id`. In this case, there will be two schedular thread attempting to read/write the same file.
- The user launched two instances of the same application.

The race condition is:
- Instance/Thread A call `LogHeartbeat()` and read from disk
- Instance/Thread B call `GetAndFlushPayload()` and clear the file
- Instance/Thread A write all the log back to the file.

app/tests/heartbeat_controller_desktop_test.cc
```diff
@@ -0,0 +1,222 @@
...
+  // Even though heartbeat logging is asynchronous, it happens in the order that it is scheduled
+  // So the heartbeats should be logged in the correct order.
+  for (auto const& entry : read_heartbeats.heartbeats) {
+    std::vector<std::string> dates = entry.second;
+    ASSERT_EQ(dates.size(), 30);
+    EXPECT_EQ(dates[0], "2000-03-01");
+    EXPECT_EQ(dates[29], "2000-03-30");
+  }
+}
+
```
chkuang-g, line 220: Is there plan to add more edge case here?
For instance:
- What if log `2000-01-25` then `2000-01-24`?
- Create destroy tests for race condition
- Having two LoggedHeartbeats instances pointing to the same app_id
etc.

chkuang-g (changes requested):

chkuang-g (changes requested):
Also
https://github.com/firebase/firebase-cpp-sdk/runs/6994613799?check_suite_focus=true

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -0,0 +1,74 @@
...
+    // If last logged timestamp is not todays date, add a heartbeat
+    if (logged_heartbeats.last_logged_date != current_date) {
+      logged_heartbeats.last_logged_date = current_date;
+      logged_heartbeats.heartbeats[user_agent].push_back(current_date);
+      // Don't store more than 30 days for the same user agent
+      if (logged_heartbeats.heartbeats[user_agent].size() > 30) {
+        logged_heartbeats.heartbeats[user_agent].erase(
+          logged_heartbeats.heartbeats[user_agent].begin());
+      }
+    }
```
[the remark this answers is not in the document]
AlmostMatt: Done.

app/src/heartbeat/heartbeat_controller_desktop.cc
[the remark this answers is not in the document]
AlmostMatt: Done. Restructured code so that the if block results in an early return.

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -0,0 +1,74 @@
...
+
+HeartbeatController::~HeartbeatController() {}
+
+void HeartbeatController::LogHeartbeat() {
+  std::string user_agent = App::GetUserAgent();
+
+  std::function<void(void)> log_heartbeat_funct = [&, user_agent]() {
+    std::string current_date = date_provider_.GetDate();
+    // Return early if a heartbeat has already been logged today.
+    if (this->last_logged_date_ == current_date) {
```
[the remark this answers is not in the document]
AlmostMatt: Changed to use <= and > instead of == and !=

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -0,0 +1,74 @@
...
+  const std::string& app_id, const Logger& logger, const DateProvider& date_provider)
+  : storage_(app_id, logger), scheduler_(), last_logged_date_(""), date_provider_(date_provider)
+{}
+
+HeartbeatController::~HeartbeatController() {}
+
+void HeartbeatController::LogHeartbeat() {
+  std::string user_agent = App::GetUserAgent();
+
+  std::function<void(void)> log_heartbeat_funct = [&, user_agent]() {
```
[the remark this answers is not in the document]
AlmostMatt: Added a comment in the header before last_logged_date_

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -0,0 +1,74 @@
...
+
+HeartbeatController::HeartbeatController(
+  const std::string& app_id, const Logger& logger, const DateProvider& date_provider)
+  : storage_(app_id, logger), scheduler_(), last_logged_date_(""), date_provider_(date_provider)
+{}
+
+HeartbeatController::~HeartbeatController() {}
+
+void HeartbeatController::LogHeartbeat() {
+  std::string user_agent = App::GetUserAgent();
```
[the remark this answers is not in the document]
AlmostMatt: Done

app/tests/heartbeat_controller_desktop_test.cc
AlmostMatt: Added a test for logging dates in reverse-chronological order and verifying that only the newest date is logged.

Added a test that calls LogHeartbeat many times and then calls the controller's Destructor
- note: current implementation of Scheduler's destructor is to stop allowing scheduling of new work and to join with the worker thread. So all already-scheduled work is actually completed.

Added a test with multiple controllers for the same app id trying to log heartbeats simultaneously. Confirmed that it currently fails due to concurrency issues of read/write happening between other read/write operations, so the test is disabled.
Filed b/237003018 to track this

app/tests/heartbeat_controller_desktop_test.cc
```diff
@@ -0,0 +1,345 @@
...
+TEST_F(HeartbeatControllerDesktopTest, DestroyControllerWhileWorkIsScheduled) {
+  HeartbeatController* destructible_controller =
+      new HeartbeatController(kAppId, logger_, mock_date_provider_);
+  {
+    // InSequence guarantees that all of the expected calls occur in order.
+    testing::InSequence seq;
+    for (int year = 2001; year <= 3000; year++) {
+      std::string date_string = std::to_string(year) + "-01-01";
+      EXPECT_CALL(mock_date_provider_, GetDate()).WillOnce(Return(date_string));
```
chkuang-g, line 276: Hum. if you delete the controller right after scheduling those tasks, there is no guarantee that those tasks will run.
https://github.com/firebase/firebase-cpp-sdk/blob/b0a90622bf6ec231dd6a9fcae8702472f73d6d39/app/src/scheduler.cc#L72

Therefore, you cannot expect data_provider being called.  Otherwise, this test is very likely to fail occasionally.

I would recommend you to just call `LogHeartbeat()` once. But do Line 269-284 1000 times :)

app/tests/heartbeat_controller_desktop_test.cc
```diff
@@ -0,0 +1,345 @@
...
+    ASSERT_EQ(dates.size(), 30);
+    EXPECT_EQ(dates[0], "2000-03-01");
+    EXPECT_EQ(dates[29], "2000-03-30");
+  }
+}
+
+TEST_F(HeartbeatControllerDesktopTest, DestroyControllerWhileWorkIsScheduled) {
+  HeartbeatController* destructible_controller =
+      new HeartbeatController(kAppId, logger_, mock_date_provider_);
+  {
```
chkuang-g, line 271: This bracket and indentation is not necessary

app/tests/heartbeat_controller_desktop_test.cc
```diff
@@ -0,0 +1,345 @@
...
+  }
+  for (int i = 1; i <= 1000; i++) {
+    destructible_controller->LogHeartbeat();
+  }
+  // Trigger the controller's destructor before all of the async calls to
+  // LogHeartbeat have completed.
+  delete destructible_controller;
+
+  // Wait a bit to verify that no async threads seg fault
+  std::this_thread::sleep_for(std::chrono::milliseconds(1000));
```
chkuang-g, line 287: Why do you need this and what is causing seg fault?

when you destroy destructible_controller, the thread created by the scheduler should be closed before the destructor ends.

chkuang-g (changes requested):

app/tests/heartbeat_controller_desktop_test.cc
chkuang-g: If you need to wait to avoid seg fault, then this API is pretty unstable.

AlmostMatt: The wait is not necessary, there is no seg fault, I have removed this.
I just thought it would be good to wait for async work to complete, but the scheduler destructor joins the worker thread so it already waits and this sleep is not relevant.

app/tests/heartbeat_controller_desktop_test.cc
AlmostMatt: Removed indendation

app/tests/heartbeat_controller_desktop_test.cc
AlmostMatt: In practice the current scheduler implementation completes all scheduled work as part of the destructor when it joins with the worker thread. So the date provider was called the expected number of times when I was testing it. 

I like your suggestion to create and destroy the controller within the loop. That should more consistently run into any timing edge cases if they exist.

app/src/heartbeat/heartbeat_controller_desktop.cc
AlmostMatt: Added this comment.

chkuang-g (dismissed):

AlmostMatt:
Merged from main

PR #999, opened 2022-06-17
Fix #994 Remove unnecessary error for FCM

### LINKED ISSUE #994: Status is pending. Return the pending future.
Operating system: iOS
Operating system version: ALL

What steps will reproduce the problem:
1. Calling RequestPermission 

What is the expected result?
At least lower level log

What happens instead of that?
SDK logs an error

Please provide any other information below, and attach a screenshot if possible.
As our analytics tools logs every error, we are getting errors from every user because of this unnecessary log.
Simply removing this log or changing its log level will fix the issue.

https://github.com/firebase/firebase-cpp-sdk/blob/6946d8a470545758b55abde8d6ec84a8e2235e0f/messaging/src/ios/messaging.mm#L183

chkuang-g:
Remove unnecessary error for FCM

There is no need to log and error when the API returns the existing future.

***

Integration test workflow
***

### Type of Change
***

#### Notes

### DIFF
--- a/messaging/src/ios/messaging.mm
+++ b/messaging/src/ios/messaging.mm
@@ -180,7 +180,6 @@ Future<void> RequestPermission() {
   FIREBASE_ASSERT_RETURN(RequestPermissionLastResult(), internal::IsInitialized());
 
   if (RequestPermissionLastResult().status() == kFutureStatusPending) {
-    LogError("Status is pending. Return the pending future.");
     return RequestPermissionLastResult();
   }
 


cynthiajoan (approved):

messaging/src/ios/messaging.mm
```diff
@@ -180,7 +180,6 @@ void NotifyListenerSet(Listener* listener) {
   FIREBASE_ASSERT_RETURN(RequestPermissionLastResult(), internal::IsInitialized());
 
   if (RequestPermissionLastResult().status() == kFutureStatusPending) {
-    LogError("Status is pending. Return the pending future.");
```
jonsimantov, line 183: Could be a LogDebug if we still need this for some reason.

chkuang-g: Ah, I merged before seeing this comment.  Do you think I should add `LogDebug` back?

I did a code search and it does not seems like we are logging anything when LastResult is pending.

PR #1000, opened 2022-06-17
Auto populate the namespace list to rename in desktop packaging

jonsimantov:
Currently, there is a large list of C++ namespaces that are renamed in the binary version of the desktop SDK. This allows the Firebase C++ SDK to embed its own C++ dependencies (e.g. absl, BoringSSL, gRPC, etc.) without affecting any other versions of those dependencies that may be present in the build (for example, if the developer's desktop app uses OpenSSL or a different version of gRPC).

Because that list of C++ namespaces is manually maintained in the desktop packaging script, it's possible that when dependencies changed (as grpc recently did), there can be new namespaces introduced in dependencies that are not added to that list. This would result in duplicate symbols when a developer ultimately links their desktop app—which, for example, already uses gRPC—with the Firebase C++ SDK, which embeds its own version of gRPC.

Fortunately, the script that performs this renaming has a mode where it will scan for all C++ namespaces and rename them all automatically, except the ones we specifically want to preserve (in this case, just "firebase::"). This will rename *all* namespaces inside the C++ libraries, except for certain system namespaces (currently "std" and "__gnu_cxx").

This PR enables this option. Ultimately the binary SDK created in this PR will need to be checked to make sure it does not have any extraneous frameworks. (This PR also adds such a check to the end of the packaging workflow.)

Finally, this PR brings back the merge_libraries_test script from the internal repo, which runs a battery of tests on merge_libraries, including new ones added to catch these very issues. The script is not in CI but can be run locally via scripts/merge_libraries_test.sh.
***

[CPP binary SDK packaging runs](https://github.com/firebase/firebase-cpp-sdk/actions/workflows/cpp-packaging.yml?query=branch%3Abugfix%2Fpackaging_auto_hide_namespaces)
***

### Type of Change
***

#### Notes


.github/workflows/cpp-packaging.yml
```diff
@@ -544,6 +549,9 @@ jobs:
           cd sdk-src
           python scripts/gha/install_prereqs_desktop.py --ssl boringssl
           cd ..
+          if [[ $(uname) == "Darwin"* ]]; then
```
jonsimantov, line 552: Fixes an issue where Mac packaging was not done in parallel due to the "parallel" tool not being installed. Speeds up packaging on Mac.

DellaBitta: :+1:

.github/workflows/cpp-packaging.yml
```diff
@@ -154,14 +159,14 @@ jobs:
...
-          ref: v${{ env.demumbleVer }}
+          ref: ${{ env.demumbleVer }}
 
       - name: build demumble
         run: |
           cd demumble-src
           cmake .
           cmake --build .
-          python demumble_test.py
+          python3 demumble_test.py
```
DellaBitta, line 169: Is there a reason this needs to be `python3` here? I thought we were going with `python` from now on?

.github/workflows/cpp-packaging.yml
```diff
@@ -605,6 +613,44 @@ jobs:
...
+          else  # linux
+            cat raw_symbols.txt | sort | uniq | bin/c++filt | grep '^[a-zA-Z_]*::' > cpp_symbols.txt
+          fi
+          # cpp_symbols.txt contains a list of all C++ symbols, sorted and deduped
+          # get a list of all top-level namespaces (except system namespaces)
+          cat cpp_symbols.txt | cut -d: -f1 | sort | uniq > cpp_namespaces.txt
+          echo "List of all namespaces found:"
+          cat cpp_namespaces.txt
+          # Filter out renamed namespaces (f_b_*), standard namespaces, and the public namespace (firebase::)
+          allow_namespaces="f_b_.*|std|stdext|__gnu_cxx|cxxabiv1|firebase"
```
DellaBitta, line 646: Thought: could this list be in a configuration file checked into the repo so it's not quite as hidden deep into a build script?

scripts/merge_libraries_test.py
```diff
@@ -0,0 +1,277 @@
+# Copyright 2018 Google LLC
```
DellaBitta, line 1: This should be 2022 if this file is new (not from g3)

DellaBitta (changes requested):

scripts/merge_libraries_test.py
jonsimantov: It's from g3

DellaBitta (approved):

PR #1005, opened 2022-06-23
cpp-packaging.yml: install patch.exe via choco on Windows

dconeybe:
This is a workaround for the `patch.exe` from Strawberry Perl crashing when attempting to patch snappy at

https://github.com/firebase/firebase-ios-sdk/blob/4cf275ff08a5998da1712f7eb8bb0f613d73b97e/cmake/external/snappy.cmake#L37

The upgrade of the firebase-ios-sdk dependency from 9.1.0 to 9.2.0 in https://github.com/firebase/firebase-cpp-sdk/pull/1003 caused some GitHub Actions Windows runners to fail to build with this error:

```
Performing patch step for 'snappy'
patching file snappy.cc
Assertation failed!

Program: C:\Strawberry\c\bin\patch.exe
File: .\src\patch\2.5.9\patch-2.5.9-src\patch.c, Line 354

Expression: hunk
```

Note that #1003 has some special logic in https://github.com/firebase/firebase-cpp-sdk/pull/1003/commits/77f3e708c1505fbf8419effd4481cc19f960b1c3 to remove the patching of the "snappy" library.

The problem appears to be that the GitHub Actions runners use the patch command that is bundled with Strawberry Perl, which cannot seem to handle the patch file format and crashes.

This PR attempts to fix the issue by installing the latest version of GNU patch using Chocolatey, in hopes that it will be chosen instead of the one from Strawberry Perl.

Googlers see b/236958574 for full details.


dconeybe:
I've triggered the cpp_packaging workflow at https://github.com/firebase/firebase-cpp-sdk/pull/1005/commits/fa8a18dd876444f2a844c45fe76941ac0714cd5b to see if it works: https://github.com/firebase/firebase-cpp-sdk/actions/runs/2551400273

dconeybe:
I've modified the workflow to explicitly prepend the directory of chocolatey's patch.exe to the front of the PATH in https://github.com/firebase/firebase-cpp-sdk/pull/1005/commits/76ac75b1c9b36c3d36dd1e6f08bad487655e8827.

I've triggered the cpp_packaging workflow at this comment to see if it works: https://github.com/firebase/firebase-cpp-sdk/actions/runs/2551478788

dconeybe:
https://github.com/firebase/firebase-cpp-sdk/actions/runs/2551868037

dconeybe:
Closing in favor of https://github.com/firebase/firebase-cpp-sdk/pull/1006

PR #1012, opened 2022-07-04
transaction_test.cc: improve error messages from TestMaxAttempts

dconeybe:
This will help reduce the amount of manual labor required to ascertain the root cause of failures of the TestMaxAttempts test, such as those noted in https://github.com/firebase/firebase-cpp-sdk/runs/7178938060.

### DIFF
--- a/firestore/integration_test_internal/src/transaction_test.cc
+++ b/firestore/integration_test_internal/src/transaction_test.cc
@@ -23,6 +23,7 @@
 #include "app/src/semaphore.h"
 #include "app/src/time.h"
 
+#include "absl/strings/str_cat.h"
 #include "absl/strings/str_join.h"
 #include "firebase/firestore.h"
 #include "firebase/firestore/firestore_errors.h"
@@ -30,6 +31,7 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 #include "util/event_accumulator.h"
+#include "util/future_test_util.h"
 #if defined(__ANDROID__)
 #include "firestore/src/android/transaction_android.h"
 #endif  // defined(__ANDROID__)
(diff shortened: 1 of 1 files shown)


firestore/integration_test_internal/src/transaction_test.cc
```diff
@@ -743,21 +745,24 @@ TEST_F(TransactionTest, TestMaxAttempts) {
...
+          transaction.Get(doc, &error, &get_error_message);
           if (error != kErrorOk) {
-            ADD_FAILURE() << "transaction.Get() failed: " << error_message
-                          << " (error code " << error << ")";
-            return Error::kErrorInternal;
+            ADD_FAILURE() << "transaction.Get() failed at Checkpoint UCA: "
+                          << get_error_message << " (error code " << error
+                          << "==" << ToFirestoreErrorCodeName(error) << ")";
+            error_message = "Test failed in update callback at Checkpoint UCA";
+            return Error::kErrorInvalidArgument;
```
dconeybe, line 765: Note: I changed kErrorInternal -> kErrorInvalidArgument because the latter is considered to be a "permanent error" and will prevent the transaction from being uselessly, and confusingly, retried.

firestore/integration_test_internal/src/transaction_test.cc
```diff
@@ -743,21 +745,24 @@ TEST_F(TransactionTest, TestMaxAttempts) {
 
   Future<void> run_transaction_future = firestore->RunTransaction(
       options,
-      [update_count, &doc](Transaction& transaction, std::string&) -> Error {
-        SCOPED_TRACE("Update callback; update_count=" +
-                     std::to_string(update_count->load()));
+      [update_count, &doc](Transaction& transaction,
+                           std::string& error_message) -> Error {
+        SCOPED_TRACE(absl::StrCat("Update callback; update_count=",
```
dconeybe, line 750: This change to absl::StrCat() is merely a code readability improvement IMO and does not actually improve any error reporting. Its change here is merely out of convenience.

dconeybe: Nevermind; I've reverted this change since using absl::StrCat opened up a whole can of worms regarding linking absl. See https://github.com/firebase/firebase-cpp-sdk/pull/1015.

wu-hui (approved):

PR #1020, opened 2022-07-08
Heartbeat Controller - Desktop Implementation - GetPayload

AlmostMatt:
Fetch heartbeats from storage, clear storage, create and return an encoded string payload of the heartbeats.
***

Added unit testing of the controller class, as well as unit tests for the payload construction and encoding methods.
Also added a test to verify that the backend can handle the generated header.
https://critique.corp.google.com/cl/460999804
***

### Type of Change
***

#### Notes

### DIFF
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -371,13 +371,19 @@ target_include_directories(firebase_app
   PUBLIC
     ${CMAKE_CURRENT_LIST_DIR}/src/include
     ${FIREBASE_GEN_FILE_DIR}
-    ${ZLIB_SOURCE_DIR}/..
-    ${ZLIB_BINARY_DIR}
   PRIVATE
     ${FIREBASE_CPP_SDK_ROOT_DIR}
     ${FLATBUFFERS_SOURCE_DIR}/include
     ${LIBSECRET_INCLUDE_DIRS}
 )
+
+if (DESKTOP)
+  target_include_directories(firebase_app
+    PUBLIC
+      ${ZLIB_SOURCE_DIR}/..
+      ${ZLIB_BINARY_DIR})
+endif()
+
 target_compile_definitions(firebase_app
   PRIVATE
     -DINTERNAL_EXPERIMENTAL=1
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -192,6 +192,7 @@ build_flatbuffers("${desktop_flatbuffers_schemas}"
                   ""
                   "")
 
+# TODO(almostmatt): find correct way to depend on zlibwrapper
 set(app_desktop_SRCS
     src/app_desktop.cc
     src/heartbeat_date_storage_desktop.cc
@@ -201,6 +202,8 @@ set(app_desktop_SRCS
     src/heartbeat/date_provider.cc
     src/heartbeat/heartbeat_storage_desktop.cc
     src/heartbeat/heartbeat_controller_desktop.cc
+    rest/gzipheader.cc
+    rest/zlibwrapper.cc
     ${FIREBASE_GEN_FILE_DIR}/app/logged_heartbeats_generated.h)
 if(ANDROID)
   set(app_platform_SRCS
@@ -369,6 +372,8 @@ target_include_directories(firebase_app
   PUBLIC
     ${CMAKE_CURRENT_LIST_DIR}/src/include
     ${FIREBASE_GEN_FILE_DIR}
+    ${ZLIB_SOURCE_DIR}/..
+    ${ZLIB_BINARY_DIR}
   PRIVATE
     ${FIREBASE_CPP_SDK_ROOT_DIR}
     ${FLATBUFFERS_SOURCE_DIR}/include
@@ -391,6 +396,7 @@ target_link_libraries(firebase_app
   PRIVATE
     ${app_flatbuffers_lib}
     ${LIBSECRET_LIBRARIES}
+    zlibstatic
 )
 # Automatically include headers that might not be declared.
 if(MSVC)
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -395,7 +395,6 @@ target_link_libraries(firebase_app
   PRIVATE
     ${app_flatbuffers_lib}
     ${LIBSECRET_LIBRARIES}
-    zlibstatic
 )
 # Automatically include headers that might not be declared.
 if(MSVC)
@@ -422,6 +421,9 @@ elseif(IOS)
       FirebaseDynamicLinks
       FirebaseInstanceID
   )
+else()
+  target_link_libraries(firebase_app
+    PRIVATE zlibstatic)
 endif()
 
 if (NOT ANDROID AND NOT IOS)
--- a/app/CMakeLists.txt
+++ b/app/CMakeLists.txt
@@ -192,7 +192,6 @@ build_flatbuffers("${desktop_flatbuffers_schemas}"
                   ""
                   "")
 
-# TODO(almostmatt): find correct way to depend on zlibwrapper
 set(app_desktop_SRCS
     src/app_desktop.cc
     src/heartbeat_date_storage_desktop.cc
--- a/app/src/heartbeat/heartbeat_controller_desktop.cc
+++ b/app/src/heartbeat/heartbeat_controller_desktop.cc
@@ -28,6 +28,7 @@
 #include "app/src/heartbeat/date_provider.h"
 #include "app/src/heartbeat/heartbeat_storage_desktop.h"
 #include "app/src/logger.h"
+#include "app/src/semaphore.h"
 #include "app/src/variant_util.h"
 
 namespace firebase {
@@ -38,6 +39,8 @@ const char* kHeartbeatsKey = "heartbeats";
 const char* kVersionKey = "version";
 const char* kUserAgentKey = "agent";
 const char* kDatesKey = "dates";
+const int kMaxPayloadSize = 1024;
+const int kMaxWaitTimeMs = 3000;
 
 HeartbeatController::HeartbeatController(const std::string& app_id,
                                          const Logger& logger,
@@ -92,92 +95,45 @@ void HeartbeatController::LogHeartbeat() {
 }
 
 std::string HeartbeatController::GetAndResetStoredHeartbeats() {
-  // TODO: Default to empty payload
-  std::shared_ptr<std::string> output_str =
-      std::make_shared<std::string>("empty_payload");
+  std::shared_ptr<std::string> output_str = std::make_shared<std::string>("");
+  std::shared_ptr<Semaphore> scheduled_work_semaphore = std::make_shared<Semaphore>();
 
   std::function<void(void)> get_and_reset_function = [&, output_str]() {
     std::string current_date = date_provider_.GetDate();
     // Return early if a heartbeat has already been logged today.
-    if (this->last_fetched_all_heartbeats_date_ == current_date) {
-      *output_str = "already fetched";
-      return;
+    if (this->last_fetched_all_heartbeats_date_ != current_date) {
+      LoggedHeartbeats logged_heartbeats;
+      bool read_succeeded = this->storage_.ReadTo(logged_heartbeats);
+      // If read fails, or if there are no stored heartbeats, return an empty
+      // payload and don't attempt to write.
+      if (read_succeeded && logged_heartbeats.heartbeats.size() > 0) {
+        // Clear all logged heartbeats, but keep the last logged date
+        LoggedHeartbeats cleared_heartbeats;
+        cleared_heartbeats.last_logged_date = logged_heartbeats.last_logged_date;
+        bool write_succeeded = this->storage_.Write(cleared_heartbeats);
+        // Only update last-logged date and return a payload if the write succeeds.
+        if (write_succeeded) {
+          this->last_logged_date_ = current_date;
+          *output_str = GetStringPayloadForHeartbeats(logged_heartbeats);
+        }
+      }
     }
-    LoggedHeartbeats logged_heartbeats;
-    bool read_succeeded = this->storage_.ReadTo(logged_heartbeats);
-    // If read fails, return an empty payload and don't attempt to write.
-    if (!read_succeeded) {
-      // TODO: return an empty payload
-      *output_str = "read failed";
-      return;
-    }
-    // TODO: if there are no stored heartbeats, return early
-    // Clear all logged heartbeats, but keep the last logged date
-    LoggedHeartbeats cleared_heartbeats;
-    cleared_heartbeats.last_logged_date = logged_heartbeats.last_logged_date;
-    bool write_succeeded = this->storage_.Write(cleared_heartbeats);
-    // Only update last-logged date if the write succeeds.
-    if (write_succeeded) {
-      this->last_logged_date_ = current_date;
-      // TODO: return a non-empty payload
-      *output_str = GetStringPayloadForHeartbeats(logged_heartbeats);
-    } else {
-      // TODO: return an empty payload
-      *output_str = "write failed";
-    }
+    // Post the semaphore to signal the main thread to read the result.
+    scheduled_work_semaphore->Post();
+    return;
   };
 
-  // TODO: have a member variable that is the "result" and have the scheduled
-  // work modify this member variable probably have a mutex for read/write to
-  // this member variable
-  scheduler::RequestHandle handle = scheduler_.Schedule(get_and_reset_function);
-  // TODO: wait until handle has completed
-  // Then read from the member variable to construct a payload
-
-  // In terms of concurrency:
-  // If multiple calls to fetch occur, they will both be scheduled
-  // information needs to transfer from the scheduler thread to the calling
-  // thread there is some potential ambiguity if main thread waits a while
-  // example, if two calls to fetch back to back, first should return a large
-  // payload second should return empty payload concurrency concern: want to be
-  // sure payload is not cleared before first reads from it if scheduled work
-  // had a return value that would be nice ios dispatch queue has a return type
-
-  /**
-   Can I update scheduler so that the returned RequestHandle has an
-   "is-completed" ? Seems like if it is doing work sequentially it should be
-   able to determine when something is finished and this is much more useful
-   than is-started (IsTriggered) or is this how IsTriggered works? IsTriggered
-   is set after Run() has returned unsure if Run() just starts a new async thing
-   or if it runs completely in the current thread
-
-    if Run() is syncronous, then IsTriggered is just what I want
-    Run() appears to be the callback function
-
-    so maybe IsTriggered is all I really need to syncronize this internally
-
-    what is the memory scope of the requesthandle?
-    Can I add a method like "WaitUntilIsTriggeredOrTimeout" to the handle?
-   *
-   */
-  // TODO: Use a semaphore or something to handle this better
-  // also make a constant for max-wait duration
-  for (int i = 0; i < 50; i++) {
-    if (handle.IsTriggered()) {
-      break;
-    }
-    std::this_thread::sleep_for(std::chrono::milliseconds(100));
-  }
+  scheduler_.Schedule(get_and_reset_function);
+  // Wait until the scheduled work completes.
+  scheduled_work_semaphore->TimedWait(kMaxWaitTimeMs);
   return *output_str;
 }
 
 std::string HeartbeatController::GetAndResetTodaysStoredHeartbeats() {
-  // TODO
+  // TODO - modified version of the above
+  // Only find one heartbeat, only clear that one heartbeat
   return "";
 }
-
-// TODO: define a helper to generate empty payload
-// Or rely on the fact that LoggedHeartbeats defaults to an empty payload
 
 std::string HeartbeatController::GetStringPayloadForHeartbeats(
     LoggedHeartbeats heartbeats) {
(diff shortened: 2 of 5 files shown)


app/CMakeLists.txt
```diff
@@ -201,6 +201,8 @@ set(app_desktop_SRCS
     src/heartbeat/date_provider.cc
     src/heartbeat/heartbeat_storage_desktop.cc
     src/heartbeat/heartbeat_controller_desktop.cc
+    rest/gzipheader.cc
```
AlmostMatt, line 204: I'm not very confident about the changes to this file. I know that controller_desktop depends on /rest/zlibwrapper which depends on /rest/gzipheader and zlib itself, and so I imitated the contents of the /rest/CmakeLists file in a way that caused my build to succeed.

AlmostMatt: I updated this file to only include/link zlib source_dir and zlibstatic for desktop (non-ios and non-android)

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -79,5 +97,171 @@ void HeartbeatController::LogHeartbeat() {
...
+  std::string decoded;
+  if (!firebase::internal::Base64Decode(input, &decoded)) {
+    // Failed to decode.
+    return "";
+  }
+
+  ZLib zlib;
+  zlib.SetGzipHeaderMode();
+  uLongf result_size = ZLib::MinCompressbufSize(decoded.length());
+  std::unique_ptr<char[]> result(new char[result_size]);
```
chkuang-g, line 255: Similarly, you can use Firebase UniquePtr as well

app/tests/heartbeat_controller_desktop_test.cc
```diff
@@ -336,5 +338,181 @@ TEST_F(HeartbeatControllerDesktopTest,
...
+        }
+      ],
+      "version":"2"
+    })json"));
+
+  std::string second_payload = controller_.DecodeAndDecompress(
+      controller_.GetAndResetStoredHeartbeats());
+  EXPECT_EQ(second_payload, "");
+}
+
```
chkuang-g, line 516: I recommend to add the following tests:

1. Dedicated tests against `GetStringPayloadForHeartbeats()` since this function parse the `LoggedHeartbeats` object manually. 
2. Directly test the expected output gzip-archived and BASE64-encoded string similar to what is done in PlatformLoggingParserTest.java, in addition to validating the value after decoding.

chkuang-g (changes requested):

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -16,17 +16,35 @@
...
 namespace firebase {
 namespace heartbeat {
 
+// Named keys in the generated JSON payload
+const char* kHeartbeatsKey = "heartbeats";
+const char* kVersionKey = "version";
+const char* kUserAgentKey = "agent";
+const char* kDatesKey = "dates";
+const int kMaxPayloadSize = 1024;
+const int kMaxWaitTimeMs = 3000;
```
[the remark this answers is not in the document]
jonsimantov: Is this blocking? Can we say if we don't get the disk data after the initial 300-500ms, continue loading it async and have it ready for the *next* REST request?

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -16,17 +16,35 @@
...
+#include <chrono>
 #include <iomanip>
+#include <map>
+#include <memory>
 #include <sstream>
 #include <string>
+#include <thread>
+#include <vector>
 
+#include "app/rest/zlibwrapper.h"  // Should I use zlib directly? or move this out of rest?
```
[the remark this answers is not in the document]
AlmostMatt: Removed comment.

app/tests/heartbeat_controller_desktop_test.cc
```diff
@@ -210,7 +212,7 @@ TEST_F(HeartbeatControllerDesktopTest, LogTwoUserAgentsOnDifferentDays) {
   // Log a heartbeat for UserAgent1 on day3
   app_common::RegisterLibrariesFromUserAgent(kCustomUserAgent1);
   controller_.LogHeartbeat();
-  std::this_thread::sleep_for(std::chrono::milliseconds(300));
+  std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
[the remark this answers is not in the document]
AlmostMatt: Accidental change from copy pasting I think. Back to 300ms now, which is fairly arbitrary but seems to be sufficient.

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -79,5 +97,171 @@ void HeartbeatController::LogHeartbeat() {
   scheduler_.Schedule(log_heartbeat_funct);
 }
 
+std::string HeartbeatController::GetAndResetStoredHeartbeats() {
+  std::shared_ptr<std::string> output_str = std::make_shared<std::string>("");
```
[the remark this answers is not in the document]
AlmostMatt: Ok, using SharedPtr

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -79,5 +97,171 @@ void HeartbeatController::LogHeartbeat() {
...
+  std::shared_ptr<Semaphore> scheduled_work_semaphore =
+      std::make_shared<Semaphore>(0);
+
+  std::function<void(void)> get_and_reset_function = [&, output_str]() {
```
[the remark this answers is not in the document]
AlmostMatt: Good point, done.

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -79,5 +97,171 @@ void HeartbeatController::LogHeartbeat() {
...
+        for (auto& entry : stored_heartbeats.heartbeats) {
+          std::string user_agent = entry.first;
+          std::vector<std::string>& dates = entry.second;
+          auto itr = std::find(dates.begin(), dates.end(), current_date);
+          if (itr != dates.end()) {
+            dates.erase(itr);
+            todays_heartbeats.heartbeats[user_agent].push_back(current_date);
+          }
+        }
+        bool write_succeeded = this->storage_.Write(stored_heartbeats);
```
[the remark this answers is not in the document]
AlmostMatt: Done.

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -79,5 +97,171 @@ void HeartbeatController::LogHeartbeat() {
...
+
+  std::string json_object = util::VariantToJson(root);
+  return CompressAndEncode(json_object);
+}
+
+std::string HeartbeatController::CompressAndEncode(const std::string& input) {
+  ZLib zlib;
+  zlib.SetGzipHeaderMode();
+  uLongf result_size = ZLib::MinCompressbufSize(input.length());
+  std::unique_ptr<char[]> result(new char[result_size]);
```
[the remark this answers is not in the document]
AlmostMatt: Done.

Also, as a full disclaimer, I basically copied this block of code from 
https://github.com/firebase/firebase-cpp-sdk/blob/main/app/rest/tests/response_binary_test.cc
So I'm not sure if there is any other simpler way to use gzip

app/src/heartbeat/heartbeat_controller_desktop.cc
AlmostMatt: Done

app/src/heartbeat/heartbeat_controller_desktop.cc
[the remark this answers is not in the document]
AlmostMatt: UniquePtr currently has an assertion indicating that is it not yet supported for array[] types, so I switched it back to unique_ptr for now.
Since I just copied the logic I'm not sure if unique_ptr is even necessary, it seems like a regular pointer might be fine.

app/tests/heartbeat_controller_desktop_test.cc
AlmostMatt: I added two tests. One calls GetStringPayloadForHeartbeats directly and compares the result to an expected encoded value. The other calls GetAndResetStoredHeartbeats and compares the result to an expected encoded value.

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -79,5 +97,171 @@ void HeartbeatController::LogHeartbeat() {
...
+
+std::string HeartbeatController::GetAndResetTodaysStoredHeartbeats() {
+  std::shared_ptr<std::string> output_str = std::make_shared<std::string>("");
+  std::shared_ptr<Semaphore> scheduled_work_semaphore =
+      std::make_shared<Semaphore>(0);
+
+  std::function<void(void)> get_and_reset_function = [&, output_str]() {
+    std::string current_date = date_provider_.GetDate();
+    // Return early if a heartbeat has already been logged today.
+    if (this->last_fetched_all_heartbeats_date_ != current_date) {
```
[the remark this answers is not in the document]
AlmostMatt: Done

app/tests/heartbeat_controller_desktop_test.cc
AlmostMatt: In order to unit test GetStringPayloadForHeartbeats, I made the method public. I saw also a macro FRIEND_TEST, but I think all current usages of FRIEND_TEST are actually no-op (guarded by an #ifdef which is false)

chkuang-g: You can try to use friend class directly

chkuang-g (changes requested):

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -79,5 +97,171 @@ void HeartbeatController::LogHeartbeat() {
...
+    }
+    heartbeat_entry_map[kUserAgentKey] = user_agent;
+    heartbeat_entry_map[kDatesKey] = dates_variant;
+    heartbeats_vector.push_back(heartbeat_entry);
+  }
+  // Construct a final object with 'heartbeats' and 'version'
+  Variant root = Variant::EmptyMap();
+  std::map<Variant, Variant>& root_map = root.map();
+  root_map[kHeartbeatsKey] = heartbeats_variant;
+  root_map[kVersionKey] = "2";
```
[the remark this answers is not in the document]
AlmostMatt: Done

app/src/heartbeat/heartbeat_controller_desktop.cc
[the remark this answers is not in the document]
AlmostMatt: Ok, filed a bug and added a TODO to reference it. I think it would be kind of nice to have such a feature but a bit out of scope for current version.

app/src/heartbeat/heartbeat_controller_desktop.h
```diff
@@ -36,14 +36,34 @@ class HeartbeatController {
...
   HeartbeatStorageDesktop storage_;
   scheduler::Scheduler scheduler_;
   const DateProvider& date_provider_;
 
+  std::string GetStringPayloadForHeartbeats(LoggedHeartbeats heartbeats);
+
   // For thread safety, the following variables should only be read or written
   // by the scheduler thread.
   std::string last_logged_date_;
+  std::string last_fetched_all_heartbeats_date_;
```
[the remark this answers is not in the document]
AlmostMatt: Done.

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -79,5 +97,171 @@ void HeartbeatController::LogHeartbeat() {
...
+  std::shared_ptr<std::string> output_str = std::make_shared<std::string>("");
+  std::shared_ptr<Semaphore> scheduled_work_semaphore =
+      std::make_shared<Semaphore>(0);
+
+  std::function<void(void)> get_and_reset_function = [&, output_str]() {
+    std::string current_date = date_provider_.GetDate();
+    // Return early if a heartbeat has already been logged today.
+    if (this->last_fetched_all_heartbeats_date_ != current_date) {
+      LoggedHeartbeats logged_heartbeats;
+      bool read_succeeded = this->storage_.ReadTo(logged_heartbeats);
```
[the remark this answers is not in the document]
AlmostMatt: Added a cooldown period so that GetPayload returns an empty string if it is called within X seconds of a previous nontrivial call (30s by default)
I made this 30s parameter a global and override it in tests

app/src/heartbeat/heartbeat_controller_desktop.h
```diff
@@ -36,14 +36,34 @@ class HeartbeatController {
...
+  // Synchronously fetches and clears today's heartbeat from storage and returns
+  // a JSON payload that has been compressed with gzip and base64 encoded.
+  // If there is no new heartbeat, an empty string is returned instead.
+  std::string GetAndResetTodaysStoredHeartbeats();
+
+  // Compress a string with gzip and base 64 encode the result.
+  std::string CompressAndEncode(const std::string& input);
+
+  // Decode a base64 encoded string and decompress the result using gzip.
+  std::string DecodeAndDecompress(const std::string& input);
```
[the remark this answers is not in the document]
AlmostMatt: Ok, I made this and other methods private and friended each of the test classes that depend on them.

app/src/heartbeat/heartbeat_controller_desktop.cc
```diff
@@ -79,5 +97,171 @@ void HeartbeatController::LogHeartbeat() {
...
+            dates.erase(itr);
+            todays_heartbeats.heartbeats[user_agent].push_back(current_date);
+          }
+        }
+        bool write_succeeded = this->storage_.Write(stored_heartbeats);
+        // Only update last-logged date and return a payload if the write
+        // succeeds.
+        if (write_succeeded) {
+          this->last_fetched_todays_heartbeat_date_ = current_date;
+          *output_str = GetStringPayloadForHeartbeats(todays_heartbeats);
```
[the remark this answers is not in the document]
AlmostMatt: I think the idea is that the v1 calling code will just check whether or not the result is entry. I could change this to return a boolean, but it was simpler to duplicate the code and return the same payload.
I'll look a bit more closely into how this would be used.

app/tests/heartbeat_controller_desktop_test.cc
AlmostMatt: Using friend class directly. Each test case is technically a different class, which I guess is what FRIEND_TEST macro tries to simplify, but I'm not sure how to include the gtest_prod header in the cmakelists, so I just call friend class directly.

app/src/heartbeat/heartbeat_controller_desktop.h
```diff
@@ -36,14 +36,52 @@ class HeartbeatController {
...
+  // If there are no new heartbeats, an empty string is returned instead.
+  std::string GetAndResetStoredHeartbeats();
+
+  // Synchronously fetches and clears today's heartbeat from storage and returns
+  // a JSON payload that has been compressed with gzip and base64 encoded.
+  // If there is no new heartbeat, an empty string is returned instead.
+  std::string GetAndResetTodaysStoredHeartbeats();
+
  private:
+  friend class HeartbeatControllerDesktopTest_EncodeAndDecode_Test;
```
chkuang-g, line 50: You need to put these friend classes between this macro

```
#ifdef FIREBASE_TESTING
#end
```

AlmostMatt: Done. Added this to the Cmake file for the test:
DEFINES
    -DFIREBASE_TEST

app/tests/heartbeat_controller_desktop_test.cc
```diff
@@ -56,6 +60,8 @@ class HeartbeatControllerDesktopTest : public ::testing::Test {
     storage_.Write(empty_heartbeats_struct);
     // Default to registering a user agent with version 1
     app_common::RegisterLibrariesFromUserAgent(kDefaultUserAgent);
+    // Override the min time between fetches to 0s
+    g_min_time_between_fetches_sec = 0.0;
```
chkuang-g, line 64: Overriding this for the entire test is a bit dangerous. We should only override it for the test which requires the override.

Here is my recommendation:
1. Add a variable to `HeartbeatControllerDesktopTest` class to store the original value of `g_min_time_between_fetches_sec`. 
2. Use `SetUp()` to always change it back to the original value. Also reset `last_read_all_heartbeats_time_` so that we can make sure the first fetch in any test works properly.
3. Only override `g_min_time_between_fetches_sec` inside the tests. Set to '0' only for tests which will call fetch more than once, ex.`GetTodaysHeartbeatThenGetAllHeartbeats`. Or set to to the test value like `GetHeartbeatsPayloadTimeBetweenFetches` to test the cooldown.

app/tests/heartbeat_controller_desktop_test.cc
```diff
@@ -336,5 +342,289 @@ TEST_F(HeartbeatControllerDesktopTest,
...
+        }
+      ],
+      "version":"2"
+    })json";
+  std::string encoded = controller_.CompressAndEncode(original_str);
+  std::string decoded = controller_.DecodeAndDecompress(encoded);
+  std::string expected_encoded_payload =
+      "H4sIAAAAAAAC_6vmUgADpYzUxKKSpNTEkmIlK4VoqKiCQjWcpaCQmJ6aV2KloFSSWlyiC-"
+      "Yo6SBJpyQCJYB6lYwMDE11DYx0DYyVYuHytVBWLEyLUllqUXFmfp6SlZKRElisFgDch9hUjQ"
+      "AAAA";
```
chkuang-g, line 360: Since you are testing against the same encoded string again and again, define this as a constant.

If the code styling allows, put the entire string the same line.

This will help
1. Code search this value in the future.  This should match the one we put in the server parser test
2. Avoid code formatter to cut this into different chunks, which make it pretty hard to tell if they are the same.

AlmostMatt: I tweaked EncodeAndDecode so that it and CreatePayloadString can generate the same encoded payload and made it a constant. This constant should match the server parser test.

For GetExpectedHeartbeatPayload, it requires an actual user-agent which is of the form agent/version and so I don't have any way to set the user agent to "test-agent". So it has a different encoded string.

app/tests/heartbeat_controller_desktop_test.cc
AlmostMatt: I now reset the value to the original value in the test's TearDown to make the change less permanent.
Resetting in setup doesn't make sense to me since at that point I dont see how the test would know what to reset to.
last_read_all_heartbeats_time_ should not need to be reset since every test creates a new instance of the controller so it should take a default value of 0.

app/src/heartbeat/heartbeat_controller_desktop.cc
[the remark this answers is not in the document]
AlmostMatt: To be more specific:
This seems to be the relevant iOS code which just checks whether or not the payload is empty:
https://source.corp.google.com/github/firebase/firebase-ios-sdk/FirebaseCore/Sources/FIRHeartbeatLogger.m;rcl=6472d7c8e0e974d5a78370353e58226d1bb07ca1;l=81

This is the relevant C++ code that we are trying to replace for v1 heartbeats:
https://source.corp.google.com/piper///depot/google3/third_party/firebase/cpp/app/src/heartbeat_info_desktop.cc;l=83;rcl=463226221
This c++ code basically just compares a timestamp and decides whether to return heartbeat code SDK, global, or None.

So For the new method to get todays heartbeat, I think it could return a payload string, or even just return a boolean. The contents of the string don't seem to matter at all.
Which would you prefer?

app/tests/heartbeat_controller_desktop_test.cc
AlmostMatt: As suggested in the chat, I have now updated it so that the override and undo is part of the test case, and it is only done for the test cases that perform multiple fetches.

app/tests/heartbeat_controller_desktop_test.cc
```diff
@@ -336,5 +344,291 @@ TEST_F(HeartbeatControllerDesktopTest,
...
+          dates: ["2000-01-23"]
+        }
+      ],
+      "version":"2"
+    })json"));
+}
+
+TEST_F(HeartbeatControllerDesktopTest, GetHeartbeatPayloadMultipleTimes) {
+  // Override the min time between fetches to allow multiple fetches
+  double time_between_fetches_original_val = g_min_time_between_fetches_sec;
```
chkuang-g, line 509: You can put this line in `HeartbeatControllerDesktopTest::HeartbeatControllerDesktopTest()` and make `time_between_fetches_original_val` as a member of `HeartbeatControllerDesktopTest`

app/tests/heartbeat_controller_desktop_test.cc
```diff
@@ -336,5 +344,291 @@ TEST_F(HeartbeatControllerDesktopTest,
...
+        }
+      ],
+      "version":"2"
+    })json"));
+
+  std::string second_payload = controller_.DecodeAndDecompress(
+      controller_.GetAndResetStoredHeartbeats());
+  EXPECT_EQ(second_payload, "");
+  // Reset the time between fetches to original value
+  g_min_time_between_fetches_sec = time_between_fetches_original_val;
```
chkuang-g, line 538: You can put the following lines in `HeartbeatControllerDesktopTest::SetUp()`

```
  void SetUp() override {
    // Reset the time between fetches to original value
    g_min_time_between_fetches_sec = time_between_fetches_original_val;
  }
```

app/tests/heartbeat_controller_desktop_test.cc
```diff
@@ -336,5 +344,291 @@ TEST_F(HeartbeatControllerDesktopTest,
...
+        }
+      ],
+      "version":"2"
+    })json"));
+}
+
+TEST_F(HeartbeatControllerDesktopTest, GetHeartbeatPayloadMultipleTimes) {
+  // Override the min time between fetches to allow multiple fetches
+  double time_between_fetches_original_val = g_min_time_between_fetches_sec;
+  g_min_time_between_fetches_sec = 0.0;
```
chkuang-g, line 510: Also, add a comment before this line explaining why you override this value. ex.

```
// Overriding g_min_time_between_fetches_sec so that we can call GetAndResetStoredHeartbeats() twice in a row.
```

AlmostMatt: Done.

app/tests/heartbeat_controller_desktop_test.cc
AlmostMatt: ok. Changed time_between_fetches_original_val to be a member variable and set it in the SetUp method so that it happens before each test case.

app/tests/heartbeat_controller_desktop_test.cc
AlmostMatt: as mentioned in chat, I put this in the TearDown instead so that it happens after the test instead of before.

chkuang-g (approved):
Thanks for all the changes!

PR #1021, opened 2022-07-09
semaphore.h: handle spurious wakeups in TimedWait() on Linux

dconeybe:
This fixes a latent bug where [`Future::Wait(int timeout_milliseconds)`](https://github.com/firebase/firebase-cpp-sdk/blob/c722ca77e4cbfd38f530f51c205a55d1c76bb212/app/src/include/firebase/future.h#L213) would occasionally return prematurely, when neither the timeout had expired nor the Future been completed. This was due to the implementation of [`Semaphore::TimedWait(int milliseconds)`](https://github.com/firebase/firebase-cpp-sdk/blob/c722ca77e4cbfd38f530f51c205a55d1c76bb212/app/src/semaphore.h#L165-L177) which calls [`sem_timedwait()`](https://linux.die.net/man/3/sem_timedwait) on Linux and Android and neglected to check if the `errno` was `EINTR`, in which case the wait should be restarted. 

This bug surfaced as the integration tests for Firestore's `TransactionTest.TestMaxAttempts` flakily failing due to a call to `Future.Await(int timeout_milliseconds)` returning as if it had timed out when, in fact, no timeout had occurred.

Note that this fix only affects Linux and Android (which runs Linux under the hood).

### DIFF
--- a/app/src/semaphore.h
+++ b/app/src/semaphore.h
@@ -192,12 +192,6 @@ class Semaphore {
         default:
           assert("sem_timedwait() failed with an unknown error" == 0);
       }
-    } else if (errno == EINTR) {
-        continue;
-      } else {
-        assert(errno == ETIMEDOUT);
-        return false;
-      }
     }
 #endif
   }
--- a/app/src/semaphore.h
+++ b/app/src/semaphore.h
@@ -172,7 +172,33 @@ class Semaphore {
     return WaitForSingleObject(semaphore_, milliseconds) == 0;
 #else  // not windows and not mac - should be Linux.
     timespec t = internal::MsToAbsoluteTimespec(milliseconds);
-    return sem_timedwait(semaphore_, &t) == 0;
+    while (true) {
+      int result = sem_timedwait(semaphore_, &t);
+      if (result == 0) {
+        // Return success, since we successfully locked the semaphore.
+        return true;
+      }
+      switch (errno) {
+        case EINTR:
+          // Restart the wait because we were woken up spuriously.
+          continue;
+        case ETIMEDOUT:
+          // Return failure, since the timeout expired.
+          return false;
+        case EINVAL:
+          assert("sem_timedwait() failed with EINVAL" == 0);
+        case EDEADLK:
+          assert("sem_timedwait() failed with EDEADLK" == 0);
+        default:
+          assert("sem_timedwait() failed with an unknown error" == 0);
+      }
+    } else if (errno == EINTR) {
+        continue;
+      } else {
+        assert(errno == ETIMEDOUT);
+        return false;
+      }
+    }
 #endif
   }
 
--- a/app/src/semaphore.h
+++ b/app/src/semaphore.h
@@ -187,10 +187,13 @@ class Semaphore {
           return false;
         case EINVAL:
           assert("sem_timedwait() failed with EINVAL" == 0);
+          return false;
         case EDEADLK:
           assert("sem_timedwait() failed with EDEADLK" == 0);
+          return false;
         default:
           assert("sem_timedwait() failed with an unknown error" == 0);
+          return false;
       }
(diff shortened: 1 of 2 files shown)


app/src/semaphore.h
```diff
@@ -172,7 +172,27 @@ class Semaphore {
...
+      }
+      switch (errno) {
+        case EINTR:
+          // Restart the wait because we were woken up spuriously.
+          continue;
+        case ETIMEDOUT:
+          // Return failure, since the timeout expired.
+          return false;
+        case EINVAL:
+          assert("sem_timedwait() failed with EINVAL" == 0);
```
DellaBitta, line 189: This seems to be a new style of asserting. This assert message doesn't get logged, correct?

DellaBitta, line 189: These should return false in case assertions are disabled (and therefore compiled out).

dconeybe: I saw this assert style somewhere else (I forget where exactly). This assert will fail unconditionally, since the string argument is _never_ equal to `0`. When this style of `assert` fails, sometimes the _only_ trace of it that you see in the log is the text between the assert's `(` and `)`. On several occasions I've had to track down which assertion failed by searching through the code base for the exact text specified to `assert`. So with this style it is easy to track it down and know exactly what the problematic value was that was returned from `sem_timedwait`.

dconeybe: Excellent point. Done.

app/src/semaphore.h
```diff
@@ -172,7 +172,30 @@ class Semaphore {
...
-    return sem_timedwait(semaphore_, &t) == 0;
+    while (true) {
+      int result = sem_timedwait(semaphore_, &t);
+      if (result == 0) {
+        // Return success, since we successfully locked the semaphore.
+        return true;
+      }
+      switch (errno) {
+        case EINTR:
+          // Restart the wait because we were woken up spuriously.
```
DellaBitta, line 183: Shouldn't we reduce the amount of time that has already transpired?

DellaBitta (changes requested):

app/src/semaphore.h
dconeybe: In this case, no, we should _not_ reduce the wait on the retry by the amount of time that has already transpired. The [`sem_timedwait()`](https://linux.die.net/man/3/sem_timedwait) function takes a _wall time_ as its argument, as opposed to the amount of time to wait. So you can just repeatedly specify the same `struct timespec` and if the given time is in the past then it will return immediately with `ETIMEDOUT`.

Looking at the `MsToAbsoluteTimespec()` function, which calculates the `struct timespec` specified to `sem_timedwait()`, it gets the current clock time and adds the desired timeout to it:

https://github.com/firebase/firebase-cpp-sdk/blob/c722ca77e4cbfd38f530f51c205a55d1c76bb212/app/src/time.h#L89-L95

DellaBitta: Got it.  Thanks!

DellaBitta (approved):

app/src/semaphore.h
```diff
@@ -172,7 +172,30 @@ class Semaphore {
     return WaitForSingleObject(semaphore_, milliseconds) == 0;
 #else  // not windows and not mac - should be Linux.
     timespec t = internal::MsToAbsoluteTimespec(milliseconds);
-    return sem_timedwait(semaphore_, &t) == 0;
+    while (true) {
```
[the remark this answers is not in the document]
dconeybe: I'll take a look.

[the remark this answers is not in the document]
dconeybe: Unit tests added in https://github.com/firebase/firebase-cpp-sdk/pull/1036 and https://github.com/firebase/firebase-cpp-sdk/pull/1037

PR #1025, opened 2022-07-14
Firestore: transaction_test.cc: Capture the DocumentReference by value in TestMaxAttempts

dconeybe:
This is an experiment to see if it fixes the flakes in TransactionTest.TestMaxAttempts.


cherylEnkidu (approved):
Thank you!

PR #1027, opened 2022-07-15
desktop_tester.py: some random improvements to help with debugging failed runs

dconeybe:
This PR makes a bunch of random improvements to `scripts/gha/desktop_tester.py`. These improvements primarily help with making sense of failed test runs.

Here is a non-exhaustive list of changes:
* Use `shlex.split()` instead of `str.split()` on the `--cmd_line` argument, to intelligently handle quoted strings with spaces in them.
* Avoid initializing `result = None` and instead leverage Python's try/except/else structure.
* Log the full command line of the test execution.
* Print the elapsed time of the test execution.
* Specify `errors="replace"` to `subprocess.run()` so that any decoding errors of stdout/stderr won't cause a crash.


dconeybe:
FYI The one failing integration tests, [test-desktop-windows-latest-x64-openssl](https://github.com/firebase/firebase-cpp-sdk/runs/7382251924) can be fixed by rolling back #1010, which re-arranged a bunch of the GitHub Actions logic.

sunmou99 (approved):
LGTM

PR #1033, opened 2022-07-18
[Bugfix] Windows openssl build

sunmou99:
1. Reverts firebase/firebase-cpp-sdk#1029
2. Also fix "Windows openssl build": 
force install openssl on windows. https://github.com/firebase/firebase-cpp-sdk/pull/1033/files#diff-25c0603f4da06db9a473d4a3e1093affddd49b779fa080a61167da830a1afaa0R65

Tests passed: https://github.com/firebase/firebase-cpp-sdk/actions/runs/2694121935


AlmostMatt:
desktop-windows-latests-x64-openssl test passed, which seems like a pretty good sign.

There is an integration test failure for android-windows-latest-android_target-gameloop. Does the failure look like a flake to you? It mentions errors, but I don't see any specific error messages in the summary.

I'll trigger re-running the failure in the meantime

scripts/gha/install_prereqs_desktop.py
```diff
@@ -54,14 +54,16 @@ def main():
           # brew install go
           utils.run_command(['brew', 'install', 'go'])
 
-    # Install openssl on linux/mac if its not installed already
-    if args.ssl == 'openssl' and not utils.is_command_installed('openssl'):
```
AlmostMatt, line 58: Is there some reason to remove this condition?
" and not utils.is_command_installed('openssl') "

sunmou99: It seems openssl already exists on windows runner. With this condition, the cmd won't run.

AlmostMatt (changes requested):
Integration tests passed.
As Denver mentioned, please update the changes of https://github.com/firebase/firebase-cpp-sdk/pull/1035 as part of this change as well since this change renames requirements to python_requirements.

AlmostMatt (approved):

PR #1037, opened 2022-07-20
semaphore_test.cc: add TimedWaitSpuriousWakeupLinux

dconeybe:
Add a unit test for the Semaphore::TimedWait() spurious wakeup fix in https://github.com/firebase/firebase-cpp-sdk/pull/1021.

### DIFF
--- a/app/tests/semaphore_test.cc
+++ b/app/tests/semaphore_test.cc
@@ -125,7 +125,7 @@ TEST(SemaphoreTest, TimedWaitSpuriousWakeupLinux) {
 }
 #endif  // #if SEMAPHORE_LINUX
 
-TEST(SemaphoreTest, MultithreadedStressTest) {
+TEST(SemaphoreTest, DISABLED_MultithreadedStressTest) {
   for (int i = 0; i < 10000; ++i) {
     firebase::Semaphore sem(0);
 
--- a/app/tests/semaphore_test.cc
+++ b/app/tests/semaphore_test.cc
@@ -21,17 +21,11 @@
 #include "gmock/gmock.h"
 #include "gtest/gtest.h"
 
-// Set SEMAPHORE_LINUX to reflect whether the platform is "Linux", using the
-// same logic that semaphore.h uses.
-#if FIREBASE_PLATFORM_OSX || FIREBASE_PLATFORM_IOS || \
-    FIREBASE_PLATFORM_TVOS || FIREBASE_PLATFORM_WINDOWS
-#define SEMAPHORE_LINUX 0
-#else
-#define SEMAPHORE_LINUX 1
+#if FIREBASE_PLATFORM_ANDROID || FIREBASE_PLATFORM_LINUX
 #include <pthread.h>
 
 #include <csignal>
-#endif
+#endif  // #if FIREBASE_PLATFORM_ANDROID || FIREBASE_PLATFORM_LINUX
 
 namespace {
 
@@ -88,12 +82,12 @@ TEST(SemaphoreTest, TimedWait) {
       0.20 * firebase::internal::kMillisecondsPerSecond);
 }
 
-#if SEMAPHORE_LINUX
+#if FIREBASE_PLATFORM_ANDROID || FIREBASE_PLATFORM_LINUX
 // Tests that Timed Wait handles spurious wakeup (Linux/Android specific).
 TEST(SemaphoreTest, TimedWaitSpuriousWakeupLinux) {
   // Register a dummy signal handler for SIGUSR1; otherwise, sending SIGUSR1
   // later on will crash the application.
-  signal(SIGUSR1, [](int signum) -> void {});
+  signal(SIGUSR1, [](int signum) {});
 
   // Start a thread that will send SIGUSR1 to this thread in a few moments.
   pthread_t main_thread = pthread_self();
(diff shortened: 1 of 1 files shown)


app/tests/semaphore_test.cc
```diff
@@ -76,6 +88,43 @@ TEST(SemaphoreTest, TimedWait) {
       0.20 * firebase::internal::kMillisecondsPerSecond);
 }
 
+#if SEMAPHORE_LINUX
```
jonsimantov, line 91: WDYT about just using #ifdef FIREBASE_ANDROID || FIREBASE_LINUX here and above? To me that's clearer, even though it's not the same as the check in semaphore.h.

app/tests/semaphore_test.cc
```diff
@@ -76,6 +88,43 @@ TEST(SemaphoreTest, TimedWait) {
...
+  // Wait for the "signal sending" thread to terminate, since it references
+  // memory on the stack and we can't have it running after this method returns.
+  signal_sending_thread.Join();
+
+  // Unregister the signal handler for SIGUSR1, since it's no longer needed.
+  signal(SIGUSR1, NULL);
+
+  // Make sure that Semaphore::TimedWait() blocked for the entire timeout, and,
+  // specifically, did NOT return early as a result of the SIGUSR1 interruption.
+  ASSERT_LT(labs((finish_ms - start_ms) -
```
jonsimantov, line 122: Can we just use EXPECT_NEAR here?

app/tests/semaphore_test.cc
```diff
@@ -76,6 +88,43 @@ TEST(SemaphoreTest, TimedWait) {
...
+      [](void* arg) {
+        firebase::internal::Sleep(500);
+        pthread_kill(*static_cast<pthread_t*>(arg), SIGUSR1);
+      },
+      &main_thread);
+
+  // Call Semaphore::TimedWait() and keep track of how long it blocks for.
+  firebase::Semaphore sem(0);
+  int64_t start_ms = firebase::internal::GetTimestamp();
+  EXPECT_FALSE(sem.TimedWait(2 * firebase::internal::kMillisecondsPerSecond));
```
jonsimantov, line 110: 2 * firebase::internal::kMillisecondsPerSecond should be in a constant in this function, for clarity

app/tests/semaphore_test.cc
```diff
@@ -76,6 +88,43 @@ TEST(SemaphoreTest, TimedWait) {
       0.20 * firebase::internal::kMillisecondsPerSecond);
 }
 
+#if SEMAPHORE_LINUX
+// Tests that Timed Wait handles spurious wakeup (Linux/Android specific).
+TEST(SemaphoreTest, TimedWaitSpuriousWakeupLinux) {
+  // Register a dummy signal handler for SIGUSR1; otherwise, sending SIGUSR1
+  // later on will crash the application.
+  signal(SIGUSR1, [](int signum) -> void {});
```
jonsimantov, line 96: Does this actually need the trailing return type syntax?

app/tests/semaphore_test.cc
```diff
@@ -76,6 +88,43 @@ TEST(SemaphoreTest, TimedWait) {
...
+  signal_sending_thread.Join();
+
+  // Unregister the signal handler for SIGUSR1, since it's no longer needed.
+  signal(SIGUSR1, NULL);
+
+  // Make sure that Semaphore::TimedWait() blocked for the entire timeout, and,
+  // specifically, did NOT return early as a result of the SIGUSR1 interruption.
+  ASSERT_LT(labs((finish_ms - start_ms) -
+                 (2 * firebase::internal::kMillisecondsPerSecond)),
+            0.20 * firebase::internal::kMillisecondsPerSecond);
```
jonsimantov, line 124: The allowed delta should probably also be in a constant

app/tests/semaphore_test.cc
dconeybe: Done.

app/tests/semaphore_test.cc
dconeybe: No it does not. I've removed the trailing return.

app/tests/semaphore_test.cc
dconeybe: Done.

app/tests/semaphore_test.cc
dconeybe: Yes, EXPECT_NEAR works perfectly. Done.

app/tests/semaphore_test.cc
dconeybe: Done.

app/tests/semaphore_test.cc
```diff
@@ -76,6 +84,50 @@ TEST(SemaphoreTest, TimedWait) {
...
+      [](void* arg) {
+        firebase::internal::Sleep(500);
+        pthread_kill(*static_cast<pthread_t*>(arg), SIGUSR1);
+      },
+      &main_thread);
+
+  // Call Semaphore::TimedWait() and keep track of how long it blocks for.
+  firebase::Semaphore sem(0);
+  auto start_ms = firebase::internal::GetTimestamp();
+  const int timed_wait_timeout = 2 * firebase::internal::kMillisecondsPerSecond;
```
jonsimantov, line 110: I think the constants should probably follow the kConstantNaming syntax

jonsimantov (approved):
Approved with one nit

app/tests/semaphore_test.cc
dconeybe: Done.

dconeybe:
@jonsimantov Please re-approve, as your previous approval was dismissed as being "stale".

PR #1056, opened 2022-08-11
Call HeartbeatController.LogHeartbeat from Auth's desktop implementation

AlmostMatt:
### Description
Update the LogHeartbeat method in auth_desktop to get an instance of HeartbeatController and call LogHeartbeat on that controller.
This PR also adds a static map from app_name to HeartbeatController instances and a static method to get such an instance. This is defined in app_desktop.h and implemented in app_desktop.cc
This PR creates app_desktop.h and moves an existing class definition from app_desktop.cc to app_desktop.h  
***

Verified via unit test that by calling the static method in AppDesktop that a functional instance of HeartbeatController
Will manually perform end to end testing after code to add payload to auth requests is completed.
***

### Type of Change
- [ x] Other, such as a build process or documentation change.
***

#### Notes

### DIFF
--- a/app/src/include/firebase/app.h
+++ b/app/src/include/firebase/app.h
@@ -702,6 +702,23 @@ class App {
 #endif  // INTERNAL_EXPERIMENTAL
 
 #ifdef INTERNAL_EXPERIMENTAL
+  // These methods are only visible to SWIG and internal users of firebase::App.
+
+  // On Desktop, log a Heartbeat
+  void LogHeartbeat();
+
+  // On Desktop, clear stored heartbeats and return a payload.
+  ///
+  /// @return encoded payload of all logged heartbeats.
+  std::string GetAndResetStoredHeartbeats();
+
+  // On Desktop, clear today's stored heartbeat and return a payload.
+  ///
+  /// @return encoded payload of today's logged heartbeat.
+  std::string GetAndResetTodaysStoredHeartbeats();
+#endif // INTERNAL_EXPERIMENTAL
+
+#ifdef INTERNAL_EXPERIMENTAL
 #if FIREBASE_PLATFORM_ANDROID
   /// Get the platform specific app implementation referenced by this object.
   ///
--- a/app/src/include/firebase/app.h
+++ b/app/src/include/firebase/app.h
@@ -702,20 +702,22 @@ class App {
 #endif  // INTERNAL_EXPERIMENTAL
 
 #ifdef INTERNAL_EXPERIMENTAL
+#if FIREBASE_PLATFORM_DESKTOP
   // These methods are only visible to SWIG and internal users of firebase::App.
 
   /// On Desktop, log a Heartbeat
-  void LogHeartbeat();
+  void LogDesktopHeartbeat();
 
   /// On Desktop, clear stored heartbeats and return a payload.
   ///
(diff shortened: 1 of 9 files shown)


app/src/include/firebase/app.h
```diff
@@ -701,6 +701,23 @@ class App {
   static void SetDefaultConfigPath(const char* path);
 #endif  // INTERNAL_EXPERIMENTAL
 
+#ifdef INTERNAL_EXPERIMENTAL
+  // These methods are only visible to SWIG and internal users of firebase::App.
+
+  // On Desktop, log a Heartbeat
```
chkuang-g, line 707: uber nit: use three slashes. And those in line 710 and 715

auth/src/desktop/auth_desktop.cc
```diff
@@ -94,9 +94,7 @@ void* CreatePlatformAuth(App* const app) {
 void InitializeFunctionRegistryListener(AuthData* auth_data);
 void DestroyFunctionRegistryListener(AuthData* auth_data);
 
-// TODO(b/211006737): This is a stub until desktop implementation supports
-// heartbeat logging.
-void LogHeartbeat(Auth* const auth) {}
+void LogHeartbeat(Auth* const auth) { auth->auth_data->app->LogHeartbeat(); }
```
chkuang-g, line 97: Though it is very unlikely to have nullptr but please do null check on `auth`, `auth->auth_data` and `auth->auth_data->app` just in case

app/src/include/firebase/app.h
AlmostMatt: Done. Also, to see if I am understanding correctly, does triple slash indicate docstring comment whereas double slash is a comment on implementation details?

auth/src/desktop/auth_desktop.cc
AlmostMatt: Done. Added a series of conditions within an if

app/src/include/firebase/app.h
```diff
@@ -701,6 +701,23 @@ class App {
...
+  ///
+  /// @return encoded payload of all logged heartbeats.
+  std::string GetAndResetStoredHeartbeats();
+
+  /// On Desktop, clear today's stored heartbeat and return a payload.
+  ///
+  /// @return encoded payload of today's logged heartbeat.
+  std::string GetAndResetTodaysStoredHeartbeats();
+#endif  // INTERNAL_EXPERIMENTAL
+
```
chkuang-g, line 720: Since these are used for desktop platform only, I feel it is a bit confusing that this API is available for iOS and Android as well.

To reduce the confusion, I would recommend:
1. Wrap this inside `#if FIREBASE_PLATFORM_WINDOWS | FIREBASE_PLATFORM_OSX | FIREBASE_PLATFORM_LINUX` and remove the stub impl for Android, iOS and stub.
2. Renaming the functions to be more specific, ex. `LogDesktopHeartbeat`, `GetAndResetStoredDesktopHeartbeats` and `GetAndResetTodaysStoredDesktopHeartbeats`

chkuang-g (changes requested):

app/src/include/firebase/app.h
AlmostMatt: Done

chkuang-g (approved):

PR #1057, opened 2022-08-12
Include logcat for failed Android emulator tests

dconeybe:


dconeybe:
Here is an example where the logcat logs were uploaded: https://github.com/firebase/firebase-cpp-sdk/actions/runs/2850292429 (see the `mobile-simulator-test-logcat-artifact` artifact)

.github/workflows/integration_tests.yml
```diff
@@ -1050,6 +1050,13 @@ jobs:
           name: mobile-simulator-test-video-artifact
           path: testapps/video-*-android-${{ matrix.build_os }}-${{ matrix.android_device }}-${{ matrix.test_type }}.mp4
           retention-days: ${{ env.artifactRetentionDays }}
+      - name: Upload Android test logcat artifact
+        if: ${{ steps.get-device-type.outputs.device_type == 'virtual' && !cancelled() }}
+        uses: actions/upload-artifact@v3
+        with:
+          name: mobile-simulator-test-logcat-artifact
```
jonsimantov, line 1057: I think the filename should also include matrix variables, no? build_os, android_device, test_type

dconeybe: I copied/adapted this from line 1051 above. I think the asterisk will capture the other matrix variables from the filename.

jonsimantov: What I'm saying is, this action will upload an artifact named "mobile-simulator-test-logcat-artifact".

Since it will be run by multiple matrix variants, the artifact name should contain the variants, otherwise they'll all upload the same artifact name.

dconeybe: Ahh. The artifacts will aggregate all of the individual files into a single zip file, in the same way that it does for the captured video. That's why the artifact name doesn't need all of the matrix variants.

dconeybe: @jonsimantov See the [mobile-simulator-test-logcat-artifact](https://github.com/firebase/firebase-cpp-sdk/suites/7882404858/artifacts/334906500) artifact: it's a zip file that contains the individual logcat files.

jonsimantov: Oh, my mistake. Thanks for explaining with the example.

sunmou99: `mobile-simulator-test-logcat-artifact` is more like the folder name. Under this folder, you will see file named with matrix config.

sunmou99 (approved):
LGTM.
We considered printing out Logcat info before. But the logcat contains too many unrelated info. Filter the logcat info by TAG/package name may be helpful (may be not).

dconeybe:
@sunmou99 At least logcat is ONLY captured if the test _fails_, in which case the logcat may be quite useful.

PR #1058, opened 2022-08-19
Fix Android and iOS emulator screen recording in GitHub Actions

dconeybe:
This PR makes a bunch of improvements and fixes to the screen recording feature of the GitHub CI testing for Android and iOS simulators.

- add `--bugreport` on Android, which overlays timestamps on the videos
- properly kill the screen recording process by sending CTRL+C
- print a detailed warning if screen recording fails
- run android emulator in "headless" mode, to potentially reduce CPU usage, by adding `-no-window` to the emulator's command line
- actually upload the tvOS screen recording artifact in `integration_tests.yml`
- fix the video number scheme to start at 1 and count up to 3, instead of starting at 3 and counting down to 1
- fix screen recording on tvOS by specifying `--display=external`
- clean up the retry logic in `test_simulator.py` to use iteration instead of recursion


scripts/gha/test_simulator.py
```diff
@@ -388,17 +390,61 @@ def _build_tvos_helper(helper_project, device_name, device_os):
...
+    creationflags=
+        subprocess.CREATE_NEW_PROCESS_GROUP
+        if platform.system() == "Windows"
+        else 0,
+  )
+  logging.info("Started screen recording with PID %s", proc.pid)
+  return (proc, output_file)
+
+
+def _stop_recording(start_recording_retval):
```
dconeybe, line 425: note to self: add the logic to upload the captured video to the workflow yml file

dconeybe:
Here is an example of how the videos from test failures look: https://github.com/firebase/firebase-cpp-sdk/actions/runs/2923963604

scripts/gha/test_simulator.py
```diff
@@ -774,12 +837,20 @@ def _get_android_test_log(test_package):
...
     try:
       subprocess.run(args, shell=shell, check=check, timeout=timeout)
-    except:
+    except subprocess.SubprocessError:
+      logging.exception("run_with_retry: %s (attempt %s of %s) FAILED", args, attempt_num, max_attempts)
+
+      # If retries have been exhausted, just raise the exception
+      if attempt_num >= max_attempts:
+        raise
+
```
sunmou99, line 852: else: attempt_num += 1

dconeybe: Fixed. Good catch!

scripts/gha/test_simulator.py
```diff
@@ -387,23 +389,77 @@ def _build_tvos_helper(helper_project, device_name, device_os):
...
+  args = [
+    "xcrun",
+    "simctl",
+    "io",
+    "booted",
+    "recordVideo",
+    "-f",
+    "--codec=h264",
+  ]
+  if display is not None:
```
sunmou99, line 402: nit: if display:
Also, we could provide the default value "external" or "internal" (Not sure what's the difference here).

scripts/gha/test_simulator.py
```diff
@@ -387,23 +389,77 @@ def _build_tvos_helper(helper_project, device_name, device_os):
...
+
+  # Specify CREATE_NEW_PROCESS_GROUP on Windows because it is required for
+  # sending CTRL_C_SIGNAL, which is done by _stop_recording().
+  proc = subprocess.Popen(
+    args,
+    stdout=output_file,
+    stderr=subprocess.STDOUT,
+    creationflags=
+        subprocess.CREATE_NEW_PROCESS_GROUP
+        if platform.system() == "Windows"
```
sunmou99, line 421: In **CI**, this script only works on **mac** runner. 
It's save to delete it, or you could try it locally and make sure it works on local windows machine.

scripts/gha/test_simulator.py
```diff
@@ -387,23 +389,77 @@ def _build_tvos_helper(helper_project, device_name, device_os):
...
+
+def _stop_recording(start_recording_retval):
+  (proc, output_file) = start_recording_retval
+  logging.info("Stopping screen recording with PID %s by sending CTRL+C",
+    proc.pid)
+
+  # Make sure that `returncode` is set on the `Popen` object.
+  proc.poll()
+
+  if proc.returncode is not None:
```
sunmou99, line 436: nit: `if proc.returncode:`
Also, on Android Emulator, if the exit code is 0, there is no log: https://github.com/firebase/firebase-cpp-sdk/runs/8029138294?check_suite_focus=true#step:11:6091

sunmou99 (dismissed):
LGTM. 
The tests on Android Emulators still hit the timeout. I will take another look (possible fix could be something like using another Emulator).

scripts/gha/test_simulator.py
dconeybe: I tried this on all 3 OS's (windows, mac, linux) and the logic works correctly; that is, it saves the screen recordings and is able to upload them as artifacts. Check out https://github.com/firebase/firebase-cpp-sdk/actions/runs/2923963604, which didn't suffer from the Android emulator timeout because I hardcoded the first test to crash.

scripts/gha/test_simulator.py
dconeybe: I'd rather not use `if display` here because then it conflates the empty string with None, and, IMO, they should be treated differently.

I only add the `--display` argument to the command line if something other than the default (which is "internal") is specified. Although explicitly specifying `--display=internal` for iOS tests should work, it's not necessary.

If you feel strongly about these comments then I'm happy to discuss and I can fix in a follow-up PR.

scripts/gha/test_simulator.py
dconeybe: I intentionally checked for `is not None` here because I want to check if the process has completed. If returncode is None then the process is still alive, which is the expected behavior. If returncode is any number, even zero, then that means that the process terminated prematurely and I want to echo whatever output there is to assist with any debugging.

In the link you provided, I can't tell which commit that integration test run was executed at, but it looks like it was from a run where there was a bug where I forgot to call proc.poll() to set returncode (which was fixed by https://github.com/firebase/firebase-cpp-sdk/pull/1058/commits/962e71e6c65e13ba48d62e6328a5679dd99c8fcf). Now, with the fix, it won't print the screenrecord output if the process is still in progress.

dconeybe:
@sunmou99 Could you re-approve? Your approval has been dropped by the bot.

> LGTM. The tests on Android Emulators still hit the timeout. I will take another look (possible fix could be something like using another Emulator).

sunmou99 (approved):

PR #1064, opened 2022-08-30
Upgrade node-version to 16 (was 14) in GitHub Actions workflows

dconeybe:
Upgrade `node-version` to 16 (was 14) in GitHub Actions workflows (`desktop.yml` and `integration_tests.yml`).

Node version 16 is the "long term support" version.

This may also fix intermittent failures of `npm install -g firebase-tools` with the error "npm ERR! cb() never called!"

### DIFF
--- a/.github/workflows/desktop.yml
+++ b/.github/workflows/desktop.yml
@@ -354,7 +354,7 @@ jobs:
           if [[ "${{ job.status }}" != "success" ]]; then
             exit 1
           fi
-      - name: Set up Node (14)
+      - name: Set up Node (16)
         uses: actions/setup-node@v2
         with:
           node-version: 16.x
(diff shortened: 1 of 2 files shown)


.github/workflows/desktop.yml
```diff
@@ -357,7 +357,7 @@ jobs:
       - name: Set up Node (14)
```
sunmou99, line 357: Nit: Set up Node (16)

dconeybe: Done.

jonsimantov:
Is this incompatible with the retry PR? (I'm merging that one - I figure it's still best to retry any network-related thing if possible).

dconeybe:
> Is this incompatible with the retry PR? (I'm merging that one - I figure it's still best to retry any network-related thing if possible).

No. They are compatible. I agree, that the retry is useful regardless of which node version is being installed.

sunmou99 (approved):

PR #1075, opened 2022-09-08
Update firestore heartbeat logging to depend on new heartbeat controller

AlmostMatt:
…ler instead of from old heartbeat date storage

### Description
Firestore heartbeat logging previously reported heartbeat code based on the date in heartbeat date storage. With this change it will instead log heartbeats based on the presence of heartbeats in the new heartbeat controller.
It will still log a v1 heartbeat (with heartbeat code COMBINED).
***
### Testing
- Updated unit tests for heartbeat code provider to verify that they are still functional with new logic
- (TODO) Run firestore integration test and verify logs in sawmill
***

### Type of Change
- [x ] Other, such as a build process or documentation change.
***

#### Notes


firestore/src/main/firebase_metadata_provider_desktop.cc
```diff
@@ -23,10 +23,12 @@ namespace firebase {
 namespace firestore {
 
 FirebaseMetadataProviderCpp::FirebaseMetadataProviderCpp(const App& app)
-    : gmp_app_id_(app.options().app_id()) {}
+    : heartbeat_controller_(app->GetHeartbeatController()),
```
chkuang-g, line 26: nit: use `Move()` to avoid triggering copy constructor for the SharedPtr. That is:

```
    : heartbeat_controller_(Move(app->GetHeartbeatController())),
```

Very minor improvement but just for good habit.

app/src/heartbeat_info_desktop.cc
```diff
@@ -16,87 +16,22 @@
...
 
-  if (!send_sdk && !send_global) {
-    return Code::None;
-  } else if (send_sdk && !send_global) {
-    return Code::Sdk;
-  } else if (!send_sdk && send_global) {
+HeartbeatInfo::Code HeartbeatInfo::GetHeartbeatCode(
+    SharedPtr<heartbeat::HeartbeatController> heartbeat_controller) {
+  std::string payload =
+      heartbeat_controller->GetAndResetTodaysStoredHeartbeats();
```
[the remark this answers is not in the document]
AlmostMatt: I expect the content of the HeartbeatController's payload to be irrelevant. I was just using the payload to determine whether or not a heartbeat was logged.

As for which user agents are attached, I assumed that the existing logic would be unchanged, which is that firebase metadata provider logs the following:
  context.AddMetadata(kXFirebaseClientHeader, App::GetUserAgent());
  
I suppose there is a possibility to improve the firestore heartbeat logging, so instead of reading user agents at the time that the request is created, it could instead get user agents from the returned heartbeat payload. Is this what you were expecting?

firestore/src/main/create_firebase_metadata_provider.h
```diff
@@ -32,7 +32,7 @@ class App;
 namespace firestore {
 
 std::unique_ptr<remote::FirebaseMetadataProvider>
-CreateFirebaseMetadataProvider(App& app);
+CreateFirebaseMetadataProvider(App* app);
```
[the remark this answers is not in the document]
AlmostMatt: Oh, oops. This is a change that I meant to undo.

auth/src/desktop/rpcs/auth_request.cc
```diff
@@ -62,7 +62,8 @@ AuthRequest::AuthRequest(::firebase::App& app, const char* schema,
     add_header("X-Client-Version", extended_auth_user_agent.c_str());
   }
   if (deliver_heartbeat) {
-    std::string payload = app.GetAndResetStoredDesktopHeartbeats();
+    std::string payload =
+        app.GetHeartbeatController()->GetAndResetStoredHeartbeats();
```
chkuang-g, line 66: null-check

chkuang-g (commented):
One more null-check and it is good to go

chkuang-g (approved):
LGTM.

Let's have Firestore team taking a look of this. I'll add them as reviewers

firestore/src/main/firebase_metadata_provider_desktop.cc
```diff
@@ -16,28 +16,27 @@
...
 #include "app/src/include/firebase/app.h"
 
 namespace firebase {
 namespace firestore {
 
+const char* kHeartbeatCodeGlobal = "2";
+
 FirebaseMetadataProviderCpp::FirebaseMetadataProviderCpp(const App& app)
-    : gmp_app_id_(app.options().app_id()) {}
+    : heartbeat_controller_(Move(app.GetHeartbeatController())),
```
dconeybe, line 27: Although it may be technically safe, this `Move` here makes me nervous. Can it be safely removed?

dconeybe (commented):
I've reviewed the code in the `firestore` directory. I'll leave it to the other reviewers to LGTM the rest of this PR.

firestore/src/main/firebase_metadata_provider_desktop.cc
AlmostMatt: This move was just meant to avoid copying a shared pointer. It could be removed, and a copy of the shared pointer would be created.

dconeybe: Ohh I see. I missed that `app.GetHeartbeatController()` was returning a shared_ptr. Please ignore this comment.

PR #1089, opened 2022-09-27
Desktop implementation of platform logging v2

AlmostMatt:
### Description
Desktop implementation of platform logging v2 (go/firebase-platform-logging-design)
A heartbeat controller maintains an internal storage of logged heartbeats, which can be added as an additional header request to supported backends. A heartbeat is just a user-agent string identifying things like operating system and firebase sdk version.

LogHeartbeat is called when App is created and when Auth getter is called, but at most once per day.

Logged heartbeats are added as a header to Auth or Firestore requests.
For Firestore requests, this header is just the user-agent string of the most recent heartbeat.
For auth requests, this header is a compressed payload which can contain multiple days worth of heartbeats. 
***
### Testing
Unit tests verify functionality of the heartbeat storage component and heartbeat controller.

Auth testapp has been run manually and confirmed that correct user-agents are logged in sawmill and concord.
***

### Type of Change
***

#### Notes


chkuang-g (commented):
LGTM.

Let's have Jon to double check just in case I miss anything, since this is a pretty large merge

app/src/heartbeat/date_provider.cc
```diff
@@ -0,0 +1,36 @@
...
+
+namespace firebase {
+namespace heartbeat {
+
+std::string DateProviderImpl::GetDate() const {
+  std::time_t t = std::time(nullptr);
+  // Use UTC time so that local time zone changes are ignored.
+  std::tm* tm = std::gmtime(&t);
+  std::ostringstream ss;
+  ss << std::put_time(tm, "%Y-%m-%d");
```
jonsimantov, line 31: Could you put a comment here saying that the %Y, %m, and %d format specifiers for put_time are locale-independent? Some of the specifiers for put_time are specific to the locale, so it would be good to just add a note that this is not an issue here (in case we end up looking for a locale-related bug in the future).

app/src/include/firebase/app.h
```diff
@@ -19,6 +19,12 @@
 
 #include "firebase/internal/platform.h"
 
+#ifdef INTERNAL_EXPERIMENTAL
+#if FIREBASE_PLATFORM_DESKTOP
+#include "app/memory/shared_ptr.h"
```
jonsimantov, line 24: firebase/include/app.h is a public header, that users will be referring to for API calls and such. We very much should not be including internal headers like shared_ptr.h in this header. Even with the ifdef INTERNAL_EXPERIMENTAL this is a bad dependency.

I really don't like an internal SDK header being included by a public header like this. 
Furthermore, even if it *should* be in a public header, probably heartbeat stuff should probably be in the internal/common.h header (or another header in the internal/ directory) rather than in the user-facing app.h header.

Anyway, I'd like to see this heartbeat stuff move into internal::AppInternal and none of our internal SharedPtr code exposed in app.h. If you must use SharedPtr to return the heartbeat controller, I would recommend using a C++ shared pointer std::shared_ptr instead.

jonsimantov (changes requested):
I realize this is just a big merge to main, but I've found an issue with how this stuff is placed in app.h that I'd like to see addressed. It could be done in the branch in a separate PR first, and then this merge updated, or it can be done directly in this merge PR, your choice.

auth/src/desktop/rpcs/verify_assertion_request.cc
```diff
@@ -43,16 +45,17 @@ VerifyAssertionRequest::VerifyAssertionRequest(const char* const api_key,
...
-  auto request = std::unique_ptr<VerifyAssertionRequest>(  // NOLINT
-      new VerifyAssertionRequest{api_key, provider_id});
+    ::firebase::App& app, const char* const api_key,
+    const char* const provider_id, const char* const id_token,
+    const char* nonce) {
+  auto request = std::unique_ptr<VerifyAssertionRequest>(
+      new VerifyAssertionRequest{app, api_key, provider_id});  // NOLINT
 
   if (id_token) {
     request->post_body_ += std::string{"&id_token="} + id_token;
```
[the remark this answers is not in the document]
jonsimantov: > ⚠️ Lint warning: `Missing space before {`

Fixed these in https://github.com/firebase/firebase-cpp-sdk/pull/1095

app/src/include/firebase/app.h
AlmostMatt: I agree that it is not great, and I would be happy to change it if we have a good alternative.

The reasoning that led to the current state was:
- HeartbeatController as a property of AppInternal means that the controller is created when app is created, and destroyed when app is destroyed, and this seems good.
- Code from auth or firestore that wants to interact with the heartbeat controller will have a pointer to App and can therefore call public methods of App, but cannot reference AppInternal directly. So a public method was added to App, and guarded by INTERNAL_EXPERIMENTAL.
- For firestore, rather than having the metadata provider keep an App* indefinitely, it seemed safer if it just interacts with App once to get the heartbeat controller. So App has a public method to return a SharedPointer of the HeartbeatController.
- To minimize the dependencies of this public header, it contains a forward declaration of HeartbeatController instead of including the actual heartbeat controller header, and it only conditionally includes shared_ptr.

Some possible alternatives or changes:
1a - We could have a static map from App to HeartbeatController and a static method to fetch a heartbeat controller given an app pointer. This would simplify the public API, but it would mean that the heartbeat controller persists even after App is destroyed, and it would require some synchronization of access to the map from App to controller.
1b - A variation of 1a would have a map from app name to HeartbeatController instead of App* to HeartbeatController. The advantage of this is that cases like firestore's metadata_provider could keep a string of the app name instead of an App* or HeartbeatController*. This would be similar to FindAppByName and FindAppLoggerByName.
2 - app.h could use std::shared_ptr instead of firebase::internal::SharedPtr
3 - App could have public methods like `std::string GetHeartbeatPayload()` instead of `SharedPtr<HeartbeatController> GetHeartbeatController()`. However this would mean that firestore's metadata provider would need to keep an App* indefinitely in order to be able to call GetHeartbeatPayload on demand.

2 would be the easiest, but 1a or 1b might be the most correct thing to do.
Let me know what you think, or if you see an alternative solution

jonsimantov: (In the simplest terms: public headers should never include private headers. Even #ifdef'ed out.)

jonsimantov: Honestly in the interest of changing the minimum, I would say just switch to std::shared_ptr, as long as that means no public header has to include a private header.

Oh, there's one other option that might be a good one, which would be to move our own SharedPtr class to a public header under `internal`. The Firestore team decided to do this with our Mutex class so they could use it from public headers as well. The description of #795 has the info on that.

app/src/heartbeat/date_provider.cc
AlmostMatt: Sure. I added the following comment:
// Format the date as yyyy-mm-dd, independent of locale.

app/src/include/firebase/app.h
AlmostMatt: Based on discussion, I will switch to std::shared_ptr, as this is the simplest change and in general std::shared_ptr should be functionally equivalent to firebase::SharedPtr and should be safe to use (we no longer support old android versions that relied on old versions of C++).

firestore/src/main/firebase_metadata_provider_desktop.cc
```diff
@@ -16,28 +16,29 @@
 
 #include "firestore/src/main/firebase_metadata_provider_desktop.h"
 
-#include "app/src/heartbeat_info_desktop.h"
 #include "app/src/include/firebase/app.h"
 
+#include <utility>
```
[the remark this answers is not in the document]
jonsimantov: Please fix.

jonsimantov (dismissed):
LGTM except please fix that lint error. :)

jonsimantov (dismissed):
LGTM, assuming tests look OK.

AlmostMatt:
Green integration test run:
https://github.com/firebase/firebase-cpp-sdk/actions/runs/3183198567/jobs/5208533268

a-maurice (approved):

PR #1097, opened 2022-10-03
Remove the Firestore patch for the Windows decltype of captured lambda build error

dconeybe:
This patch was added for the M121 release (https://github.com/firebase/firebase-cpp-sdk/pull/1076) to get the build fixed on Windows. It patched in https://github.com/firebase/firebase-ios-sdk/pull/10237. Since this fix will be included in the M122 release, there is no need to patch it in anymore. In fact, patching it in will likely cause a build error because the patch will not apply cleanly.

This PR should not be merged directly; rather, it should be patched into the PR that upgrades the dependencies on the firebase-ios-sdk.


dconeybe:
FYI: This PR has been manually cherry-picked into https://github.com/firebase/firebase-cpp-sdk/pull/1104. Close this PR once that PR is merged.

PR #1098, opened 2022-10-04
Add an initial version of App Check integration test

a-maurice:
Add an initial version of an integration test for App Check. It currently doesn't test much beyond creation of the providers.
***

Building and running the Android and Desktop versions locally
***

### Type of Change
***

#### Notes


AlmostMatt:
To clarify, once this is committed will app check integration test be run as part of automated GHA integration test workflows? Will it run for all platforms?

If yes, can you link to a passing GHA workflow run, and if no, can you mention this in the PR description

AlmostMatt:
This generally looks good, I just want to be sure that this won't break existing GHA workflows.

AlmostMatt (approved):
Looks good.
As discussed, runs of the integration test workflow specify which projects to test and the default is to not test app check. So this won't break existing integration test workflows.

Some of the tests currently failed because the stubs are not yet implemented (for example, create provider returns null), but this is ok since the existence of this test is helpful to assist in such implementation, and this test is not run automatically.

a-maurice:
Note that the lint warning is because this adds encrypted files, which obviously aren't lint compatible.

PR #1111, opened 2022-10-17
Firestore: Remove copies in FieldValueInternal to reduce global refs consumption.

dconeybe:
In `field_value_android.h`, change the `std::vector<FieldValue>` and `MapFieldValue` constructors to take const references, instead of values. This reduces by half the number of global references needed when `DocumentReference::Set()`, `DocumentReference::Update()`, and `CollectionReference::Add()` are called on Android.

Since these 3 methods are part of the public interface and take const references as their arguments, they are not able to `std::move` these values into the `FieldValueInternal` constructor, thus requiring copies. On Android, these copies necessitate doubling the number of JNI global references in use, making it much easier to hit the 51,200 hard limit.

This is the first part of a proper fix for the global refs exhaustion issue reported in https://github.com/firebase/quickstart-unity/issues/1303 and https://github.com/firebase/quickstart-unity/issues/1193#issuecomment-1242203289 (Googlers see b/251869890 for more details). This should help alleviate the immediate problems being experienced by these customers. A proper fix is much more complicated and will take time to implement. Hopefully the improvement in this PR will be sufficient to unblock those customers.

### DIFF
--- a/firestore/src/android/field_value_android.cc
+++ b/firestore/src/android/field_value_android.cc
@@ -156,7 +156,7 @@ FieldValueInternal::FieldValueInternal(GeoPoint value)
   object_ = GeoPointInternal::Create(env, value);
 }
 
-FieldValueInternal::FieldValueInternal(std::vector<FieldValue> value)
+FieldValueInternal::FieldValueInternal(const std::vector<FieldValue>& value)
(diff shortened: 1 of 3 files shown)


cherylEnkidu (approved):
Thank you!

PR #1114, opened 2022-10-18
Implement iOS AppCheck providers

AlmostMatt:
Implements Debug, DeviceCheck, and AppAttest AppCheck Providers and Factories.
Updated the app_check integration_test to call the GetToken method for each of these providers.

For DeviceCheck and AppAttest providers, this is expected to fail.
For the Debug provider, a valid debug token was registered in the console. restore_secrets will set that debug token in the xcode project environment variables, and GetToken will return a valid token.
***
### Testing
Ran restore_secrets locally, ran build_testapps, ran iOS testapp, and the integration test succeeded. The debug provider returned a valid token and the other providers returned error codes.
***

### Type of Change
***

#### Notes


build_scripts/packaging.conf
```diff
@@ -1,6 +1,6 @@
 # This config file is Bash shell syntax, and is included in all packaging scripts.
 
 # List of all Firebase products to include in the binary SDK package.
-readonly -a product_list=(admob analytics app auth database
+readonly -a product_list=(admob analytics app app_check auth database
```
a-maurice, line 4: We don't want to update this one yet, since this is used for the release build.

app_check/integration_test/src/integration_test.cc
```diff
@@ -335,6 +337,31 @@ TEST_F(FirebaseAppCheckTest, TestSignIn) {
...
+  auto token_callback{[](firebase::app_check::AppCheckToken token,
+                         int error_code, const std::string& error_message) {
+    EXPECT_EQ(firebase::app_check::kAppCheckErrorNone, error_code);
+    EXPECT_EQ("", error_message);
+    EXPECT_NE(0, token.expire_time_millis);
+    EXPECT_NE("", token.token);
+  }};
+  provider->GetToken(token_callback);
+  // Sleep for a bit to give GetToken time to succeed
+  std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
a-maurice, line 358: Using a semaphore, or even just a spin loop that checks some variable the callback can change seems like it would be better to guarantee the callback actually happens.

app_check/src/ios/common_ios.h
```diff
@@ -0,0 +1,34 @@
...
+#ifndef FIREBASE_APP_CHECK_SRC_IOS_COMMON_IOS_H_
+#define FIREBASE_APP_CHECK_SRC_IOS_COMMON_IOS_H_
+
+#import "FIRAppCheckToken.h"
+#include "firebase/app_check.h"
+
+namespace firebase {
+namespace app_check {
+
+AppCheckError AppCheckErrorFromNSError(NSError *_Nullable error);
```
a-maurice, line 26: Any reason to not include these two in the app_check_ios.h header?

AlmostMatt: I was thinking that app_check_ios.h contains internal versions of methods in the public AppCheck API whereas these are helpers to be used by providers and other app check implementation code.
But I would be ok with moving these into the app_check_ios header

app_check/integration_test/src/integration_test.cc
AlmostMatt: I tried <semaphore.h> but found that it was not supported on iOS. I also noticed that app/src/semaphore.h has different implementations per platform, and so getting a semaphore to work in a test seems like extra work.
I'll try to get a std::future or spin loop or something similar to work.

app_check/src/ios/common_ios.h
AlmostMatt: Actually, there is a relevant distinction:
These methods reference iOS concepts: NSError and FIRAppCheckToken
So the header that defines these methods cannot be included from a .cc file. And app_check.cc conditionally includes "app_check/src/ios/app_check_ios.h".

That said, since it is a separate header, I will make it a separate source file as well. And I will rename to util_ios and put it in the internal namespace to match app, database, and storage

build_scripts/packaging.conf
AlmostMatt: Reverted this file in later commit

app_check/integration_test/src/integration_test.cc
AlmostMatt: I updated the test to use std::promise / std::future in order to synchronize.

AlmostMatt:
I've updated this PR to include DeviceCheck and AppAttest providers, following the same pattern as the Debug provider.
I updated integration test cases for those two providers to just assert that an error is returned for now, since the app is not fully configured for those providers.

For the AppAttest provider, there is an implementation complication in that the iOS SDK does not define a factory and recommends writing a custom factory to conditionally pick between DeviceCheck and AppAttest based on your version of iOS.
https://firebase.google.com/docs/app-check/ios/app-attest-provider#initialize

I'm not sure if we want to do the same for the AppAttest provider, or if we would recommend they pick between factories at runtime instead of having a custom factory to pick between providers at runtime.

a-maurice (approved):

app_check/src/ios/common_ios.h
a-maurice: SGTM

app_check/src/ios/debug_provider_ios.mm
```diff
@@ -26,11 +67,19 @@
...
+  }
+}
 
 DebugAppCheckProviderFactory::~DebugAppCheckProviderFactory() {}
 
-AppCheckProvider* DebugAppCheckProviderFactory::CreateProvider(App* app) { return nullptr; }
+AppCheckProvider* DebugAppCheckProviderFactory::CreateProvider(App* app) {
+  FIRAppCheckDebugProvider* createdProvider =
+      [g_ios_debug_app_check_provider_factory createProviderWithApp:app->GetPlatformApp()];
+  return new internal::DebugAppCheckProvider(createdProvider);
```
a-maurice, line 81: I realize this is potentially leaking memory.  The cleanup for these things is sorta confusing.  We should probably use a map to save the returned providers, which will also let us return the same Provider for the same App (which based on the documentation it should be doing.)

a-maurice (changes requested):

a-maurice:
> I'm not sure if we want to do the same for the AppAttest provider, or if we would recommend they pick between factories at runtime instead of having a custom factory to pick between providers at runtime.

I think the best was is to provide the factories, and let the user determine what they want to do, instead of having to expose a single provider.

app_check/src/ios/debug_provider_ios.mm
AlmostMatt: I updated device check and app attest to also have an internal class and I updated all three ios providers to have a map from app* to provider*

app_check/src/ios/app_attest_provider_ios.mm
```diff
@@ -12,10 +12,69 @@
...
+      getTokenWithCompletion:^(FIRAppCheckToken* _Nullable token, NSError* _Nullable error) {
+        completion_callback(firebase::app_check::internal::AppCheckTokenFromFIRAppCheckToken(token),
+                            firebase::app_check::internal::AppCheckErrorFromNSError(error),
+                            util::NSStringToString(error.localizedDescription).c_str());
+      }];
+}
+
+AppAttestProviderFactoryInternal::AppAttestProviderFactoryInternal() {}
+
+AppAttestProviderFactoryInternal::~AppAttestProviderFactoryInternal() {}
```
a-maurice, line 60: It'd be good to clear out the map in the destructor.  That is, delete the Providers, and then clear it.

app_check/src/ios/app_attest_provider_ios.mm
```diff
@@ -12,10 +12,69 @@
...
+    std::function<void(AppCheckToken, int, const std::string&)> completion_callback) {
+  [ios_provider_
+      getTokenWithCompletion:^(FIRAppCheckToken* _Nullable token, NSError* _Nullable error) {
+        completion_callback(firebase::app_check::internal::AppCheckTokenFromFIRAppCheckToken(token),
+                            firebase::app_check::internal::AppCheckErrorFromNSError(error),
+                            util::NSStringToString(error.localizedDescription).c_str());
+      }];
+}
+
+AppAttestProviderFactoryInternal::AppAttestProviderFactoryInternal() {}
```
a-maurice, line 58: It'd be good to initialize the created_providers_ in the constructor.

app_check/src/ios/debug_provider_ios.mm
```diff
@@ -16,15 +16,66 @@
...
+  /// the token or with an error code and error message.
+  virtual void GetToken(
+      std::function<void(AppCheckToken, int, const std::string&)> completion_callback) override;
+
+ private:
+  FIRAppCheckDebugProvider* ios_provider_;
+};
+
+DebugAppCheckProvider::DebugAppCheckProvider(FIRAppCheckDebugProvider* provider)
+    : ios_provider_(provider) {}
```
a-maurice, line 46: Same two comments here.

app_check/src/ios/device_check_provider_ios.mm
```diff
@@ -12,25 +12,98 @@
...
+    std::function<void(AppCheckToken, int, const std::string&)> completion_callback) {
+  [ios_provider_
+      getTokenWithCompletion:^(FIRAppCheckToken* _Nullable token, NSError* _Nullable error) {
+        completion_callback(firebase::app_check::internal::AppCheckTokenFromFIRAppCheckToken(token),
+                            firebase::app_check::internal::AppCheckErrorFromNSError(error),
+                            util::NSStringToString(error.localizedDescription).c_str());
+      }];
+}
+
+DeviceCheckProviderFactoryInternal::DeviceCheckProviderFactoryInternal() {
```
a-maurice, line 60: And here

a-maurice (approved):

app_check/src/ios/app_attest_provider_ios.mm
AlmostMatt: No code is necessary to initialize created_providers_, so I assume you mean to just add " : created_providers_()" to the constructor

app_check/src/ios/app_attest_provider_ios.mm
AlmostMatt: I was thinking that in general the thing created by a factory can outlast the factory, but I think in the case of app check it is reasonable to assume that firebase SDK is the owner for the factory and all of the providers created by it, and the factory is not expected to be destroyed while it is still being used by app check.

So I will go ahead and delete the providers and clear the map.

app_check/src/ios/app_attest_provider_ios.mm
a-maurice: Yeah

a-maurice (approved):

PR #1120, opened 2022-10-25
Make the app check debug provider use an internal class

a-maurice:
Make the App Check debug provider use an internal class, so that each platform can add privates without affecting the others.
***

Locally building and
https://github.com/firebase/firebase-cpp-sdk/actions/runs/3317166856/jobs/5479887578
***

### Type of Change
***

#### Notes

### DIFF
--- a/app_check/src/common/debug_provider_common.cc
+++ b/app_check/src/common/debug_provider_common.cc
@@ -38,7 +38,7 @@ DebugAppCheckProviderFactory* DebugAppCheckProviderFactory::GetInstance() {
 }
 
 DebugAppCheckProviderFactory::DebugAppCheckProviderFactory()
-  : internal_(new internal::DebugAppCheckProviderFactoryInternal()) {}
+    : internal_(new internal::DebugAppCheckProviderFactoryInternal()) {}
 
 DebugAppCheckProviderFactory::~DebugAppCheckProviderFactory() {
   if (internal_) {
--- a/app_check/src/common/debug_provider_common.cc
+++ b/app_check/src/common/debug_provider_common.cc
@@ -1,0 +1,58 @@
+// Copyright 2022 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


app_check/src/common/debug_provider_common.cc
```diff
@@ -0,0 +1,58 @@
...
+  }
+  return g_debug_app_check_provider_factory;
+}
+
+DebugAppCheckProviderFactory::DebugAppCheckProviderFactory()
+    : internal_(new internal::DebugAppCheckProviderFactoryInternal()) {}
+
+DebugAppCheckProviderFactory::~DebugAppCheckProviderFactory() {
+  if (internal_) {
+    delete internal_;
```
AlmostMatt, line 45: What is the significance of setting internal_ to nullptr after delete has already been called?

a-maurice: It isn't really needed, but it is just a way to validate that if anything calls into this after the deletion (which ideally shouldn't happen), it will hit a nullptr on internal_ and stop.

app_check/src/include/firebase/app_check/debug_provider.h
```diff
@@ -20,6 +20,12 @@
 namespace firebase {
 namespace app_check {
 
+/// @cond FIREBASE_APP_INTERNAL
```
AlmostMatt, line 23: How does /// @cond work compared to #IFDEF ?

Would it be simpler to have a forward declaration at the top of this file?
Forward declarations is what I have seen for our public headers for classes like
StorageInternal, AppInternal, AuthData, InstallationsInternal, etc

app_check/src/include/firebase/app_check/debug_provider.h
```diff
@@ -43,6 +49,10 @@ class DebugAppCheckProviderFactory : public AppCheckProviderFactory {
 
  private:
   DebugAppCheckProviderFactory();
+
+  /// @cond FIREBASE_APP_INTERNAL
```
AlmostMatt, line 53: Similar to the above, if unconditional forward declaration is possible, then this variable can be defined unconditionally as well.

app_check/src/common/debug_provider_common.cc
AlmostMatt: Alright, seems reasonable.

app_check/src/include/firebase/app_check/debug_provider.h
a-maurice: I believe it is more for doxygen, and for whatever reason we have been using it for other headers.  But we talked a bit offline, and it doesn't seem to be necessary, so will remove.

app_check/src/include/firebase/app_check/debug_provider.h
a-maurice: Same as above, will remove.

AlmostMatt (approved):

PR #1133, opened 2022-11-01
Fix toolchain selection and packages on linux x86

jonsimantov:
There were several issues preventing Linux x86 from being built correctly since the removal of vcpkg.

First, we were no longer setting the linux_32.cmake toolchain file in build_desktop.py - this has been fixed.

Second, 32-bit builds could not locate 32-bit versions of certain libraries. This has been fixed by adding the 32-bit package directory to the PKG_CONFIG_PATH environment variable in both source and packaged builds.

Finally, libsecret 32-bit was not being installed correctly due to package conflicts, which aptitude decided to resolve by simply not installing libsecret. Instead, on github runners, remove the conflicting packages as part of the same aptitude command line that does the installation, so it uses that as its method of resolution and does install libsecret.
***

Running expanded matrix integration tests against packaged and source builds.
***

### Type of Change
***

#### Notes


scripts/gha/utils.py
```diff
@@ -231,16 +231,35 @@ def install_x86_support_libraries(gha_build=False):
...
-      process = subprocess.run(["dpkg", "-s"] + packages, stdout=devnull, stderr=subprocess.STDOUT)
+      process = subprocess.run(["dpkg", "-s"] + packages, stdout=devnull,
+                               stderr=subprocess.STDOUT)
 
     if process.returncode != 0:
-      # This implies not all of the required packages are already installed on user's machine
-      # Install them.
-      run_command(['dpkg', '--add-architecture', 'i386'], as_root=True, check=True)
+      # This implies not all of the required packages are already installed on
+      # user's machine Install them.
```
DellaBitta, line 244: nit 'machine.'

DellaBitta (approved):

PR #1135, opened 2022-11-03
Add GetSessionId implementation for Android, iOS, and stub.

jonsimantov:
Add support for Analytics GetSessionId function, as documented on mobile [here](https://firebase.google.com/docs/reference/ios/firebaseanalytics/api/reference/Classes/FIRAnalytics#+sessionidwithcompletion:) and [here](https://firebase.google.com/docs/reference/android/com/google/firebase/analytics/FirebaseAnalytics#public-tasklong-getsessionid).
***

Integration tests in this PR. Two caveats:
1. On iOS it can take a few seconds for Analytics to be initialized so we use a flaky test block to retry until it works.
2. On Android, TestSetConsent sets AnalyticsStorage consent to Denied. Even after it's set back to Granted, there will be no session ID until the app is restarted, so we have to skip TestGetSessionId if TestSetConsent runs first. (By default, this won't happen; only if test order is randomized.)
***

### Type of Change
***

#### Notes


analytics/integration_test/src/integration_test.cc
```diff
@@ -102,6 +102,27 @@ TEST_F(FirebaseAnalyticsTest, TestGetAnalyticsInstanceID) {
...
+
+  future = firebase::analytics::GetSessionId();
+  WaitForCompletion(future, "GetSessionId");
+
+#if TARGET_OS_SIMULATOR
+  FLAKY_TEST_SECTION_END();
+#endif  // TARGET_OS_SIMULATOR
+
+  EXPECT_TRUE(future.result() != nullptr);
+  EXPECT_NE(*future.result(), 0L);
```
DellaBitta, line 123: nit: Shouldn't '0L' be '(int64_t)0' as L isn't consistently defined? In this case it's zero, so probably not a problem, but in the stub implementation (below) it's not.

analytics/src/analytics_android.cc
```diff
@@ -379,5 +381,60 @@ Future<std::string> GetAnalyticsInstanceIdLastResult() {
...
+          }
+          auto* future_data = internal::FutureData::Get();
+          if (future_data) {
+            bool success =
+                result_code == util::kFutureResultSuccess && result != nullptr;
+            FutureHandleId future_id =
+                reinterpret_cast<FutureHandleId>(callback_data);
+            FutureHandle handle(future_id);
+            future_data->api()->CompleteWithResult(
+                handle, success ? 0 : -1,
```
DellaBitta, line 413: Interesting. Analytics doesn't have an error code enum?

DellaBitta (approved):

analytics/src/analytics_android.cc
jonsimantov: No, it never had an error code until the instance ID getter which was a one-off until now. Adding an enum now would be an API change I think, but we should consider doing it if we ever get a third async method. :)

analytics/integration_test/src/integration_test.cc
jonsimantov: OK yeah, I'll add the static cast.

analytics/src/analytics_android.cc
DellaBitta: Yeah, makes sense!

jonsimantov:
This needs to be merged at the same time as https://github.com/firebase/firebase-unity-sdk/pull/526 or the Unity build will break.

PR #1136, opened 2022-11-03
Pin to a version of iOS with nanopb fix

a-maurice:
Nanopb has an issue with different versions of python that is causing problems on the build machines. This was fixed on the iOS SDK side, so pin to that version until a release is out with it.
***

***

### Type of Change
***

#### Notes

### DIFF
--- a/firestore/src/main/query_main.cc
+++ b/firestore/src/main/query_main.cc
@@ -237,7 +237,7 @@ core::Bound QueryInternal::ToBound(
     const std::vector<FieldValue>& field_values) const {
   const core::Query& internal_query = query_.query();
   // Use explicit order bys  because it has to match the query the user made.
-#if TARGET_OS_IOS
+#if TARGET_OS_IOS || TARGET_OS_TV
   const core::OrderByList& explicit_order_bys =
       internal_query.explicit_order_bys();
 #else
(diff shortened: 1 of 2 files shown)


dconeybe:
Just a thought... it may be easier to leave the firebase-ios-sdk pinned at the CocoaPods version, and update https://github.com/firebase/firebase-cpp-sdk/blob/main/cmake/external/firestore.patch.txt to patch in the nanopb fix.

firestore/src/main/query_main.cc
```diff
@@ -237,8 +237,13 @@ core::Bound QueryInternal::ToBound(
     const std::vector<FieldValue>& field_values) const {
   const core::Query& internal_query = query_.query();
   // Use explicit order bys  because it has to match the query the user made.
+#if TARGET_OS_IOS || TARGET_OS_TV
```
jonsimantov, line 240: Is this just from the merge from main, or related to the fix?

jonsimantov (approved):

firestore/src/main/query_main.cc
a-maurice: Related to the fix. The next version of the iOS SDK will change this, but because on desktop it uses the source code directly, but on iOS it uses the last released cocoapods, this needs to be done in the meantime.

PR #569, opened 2021-07-22
Feature/enable tvos packaging

vimanyu:
Enable building tvos libraries in nightly packaging builds and support for tvos in integration tests against packaged sdk.

PR #1145, opened 2022-11-16
Implement ObjectArena and ArenaRef to solve JNI Global Reference Issue

cherylEnkidu:
***
***

### Type of Change
- [x] Bug fix. ([#569](https://github.com/firebase/firebase-unity-sdk/issues/569))([#1193](https://github.com/firebase/quickstart-unity/issues/1193#issuecomment-1242203289))
***

#### Notes

### DIFF
--- a/firestore/src/jni/arena_ref.h
+++ b/firestore/src/jni/arena_ref.h
@@ -39,13 +39,13 @@ class ArenaRef {
 
   explicit ArenaRef(const Object& ob) {
     Env env(GetEnv());
-    id_ = assignState(ObjectArena::GetInstance().Put(env, ob));
+    assignState(ObjectArena::GetInstance().Put(env, ob));
   }
 
   ArenaRef(const ArenaRef& other) {
     if (other.id_.first) {
       Env env(GetEnv());
-      id_ = assignState(ObjectArena::GetInstance().Dup(env, other.id_.second));
+      assignState(ObjectArena::GetInstance().Dup(env, other.id_.second));
     }
   }
 
@@ -57,8 +57,7 @@ class ArenaRef {
       }
       reset();
       if (other.id_.first) {
-        id_ =
-            assignState(ObjectArena::GetInstance().Dup(env, other.id_.second));
+        assignState(ObjectArena::GetInstance().Dup(env, other.id_.second));
       }
     }
     return *this;
@@ -72,7 +71,7 @@ class ArenaRef {
       if (id_.first) {
         ObjectArena::GetInstance().Remove(env, id_.second);
       }
-      id_ = assignState(other.release());
+      assignState(other.release());
     }
     return *this;
   }
@@ -84,7 +83,10 @@ class ArenaRef {
     }
   }
 
+  bool has_value() const { return id_.first; }
+
   Local<Object> get() const {
+    FIREBASE_DEV_ASSERT(id_.first);
     Env env(GetEnv());
     return ObjectArena::GetInstance().Get(env, id_.second);
   }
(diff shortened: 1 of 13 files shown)


firestore/src/jni/arena_ref.h
```diff
@@ -0,0 +1,125 @@
...
+    return id;
+  }
+
+  void assignState(int64_t id) { id_ = {true, id}; }
+
+  void reset() { id_ = {false, -1}; }
+
+  // A pair of variables which indicate the state of ArenaRef. If id_.first set
+  // to
+  std::pair<bool, int64_t> id_{false, -1};
```
dconeybe, line 118: Instead of using a `std::pair` just create two separate variables. This will greatly increase readability. e.g. name these variables `valid_` and `id_`.

firestore/src/jni/arena_ref.h
```diff
@@ -0,0 +1,125 @@
...
+ * automatically deletes the JNI local reference when it goes out of scope.
+ * Copies and moves are handled by creating additional references as required.
+ */
+class ArenaRef {
+ public:
+  ArenaRef() = default;
+
+  explicit ArenaRef(const Object& ob) {
+    Env env(GetEnv());
+    assignState(ObjectArena::GetInstance().Put(env, ob));
```
dconeybe, line 42: For maximal robustness, you should verify that `env.ok()` returns `true` before assuming that the value returned from Put() or Dup() is valid. If it is `false` then I'm not 100% sure what the right thing is to do. I think the best idea would be to make an ArenaRef that is not valid (i.e. its boolean "valid" flag is false).

firestore/src/jni/arena_ref.h
```diff
@@ -0,0 +1,125 @@
...
+      }
+      reset();
+      if (other.id_.first) {
+        assignState(ObjectArena::GetInstance().Dup(env, other.id_.second));
+      }
+    }
+    return *this;
+  }
+
+  ArenaRef(ArenaRef&& other) noexcept : ArenaRef(other.release()) {}
```
dconeybe, line 66: This is going to fail if other.id_.first is not true. Please improve the robustness.

firestore/src/jni/arena_ref.h
```diff
@@ -0,0 +1,125 @@
...
+
+  ArenaRef& operator=(ArenaRef&& other) noexcept {
+    if (id_.second != other.id_.second) {
```
dconeybe, line 69: You should compare id_.first here as well, since the value of id_.second should be treated as unspecified if id_.first is false.

firestore/src/jni/arena_ref.h
```diff
@@ -0,0 +1,125 @@
...
+  ~ArenaRef() {
+    if (id_.first) {
+      Env env(GetEnv());
+      ObjectArena::GetInstance().Remove(env, id_.second);
+    }
+  }
+
+  bool has_value() const { return id_.first; }
+
+  Local<Object> get() const {
```
dconeybe, line 88: get() and clear() should take the Env as an argument, rather than creating the Env out of thin air.

firestore/src/jni/object_arena.cc
```diff
@@ -0,0 +1,59 @@
...
+#include "firestore/src/jni/hash_map.h"
+#include "firestore/src/jni/loader.h"
+#include "firestore/src/jni/long.h"
+
+namespace firebase {
+namespace firestore {
+namespace jni {
+namespace {
+
+Global<ObjectArena> kInstance;
```
dconeybe, line 29: kInstance and kHashMap are not initialized anywhere. They should be initialized in ObjectArena::Initialize().

firestore/src/jni/object_arena.cc
```diff
@@ -0,0 +1,59 @@
...
+Global<HashMap> kHashMap;
```
dconeybe, line 30: The Global<HashMap> kHashMap should be a member variable of ObjectArena, not a global constant.

dconeybe (changes requested):
Please write a suite of unit tests for ArenaRef. I see some potential bugs, especially with the move constructor and move assignment operator. Please thoroughly test that class.

PR #1167, opened 2022-12-10
[macOS sandbox mode] Application group name mangling for semaphores

DellaBitta:
When macOS Sandbox mode is enabled, [macOS requires](https://developer.apple.com/library/archive/documentation/Security/Conceptual/AppSandboxDesignGuide/AppSandboxInDepth/AppSandboxInDepth.html#//apple_ref/doc/uid/TP40011183-CH3-SW24) that semaphores have a name that is prefixed by the [App's Group Name](https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_security_application-groups). If the semaphore's name doesn't match this convention then its creation fails. 

Unfortunately there's no official way for the SDK to query the app's group name at runtime, so we can't automatically mangle the semaphore names.

Instead I've updated the SDK to use an `Info.plist` property named `FBAppGroupEntitlementName` on macOS. If that property is present then the SDK will use it's value to prefix the semaphore names. 

As an additional issue, the SDK attempted to detect semaphore creation errors by comparing the semaphore handle to `nil`. But in the case of macOS, a semaphore creation error returns `SEM_FAILED` which is `0xFFFFFFFFFFFFFFFF`, not `nil`. And on Linux, `sem_init` returns `-1`. I've updated the corresponding platform implementations to detect the correct  values for these errors.

***

[Integration Test](https://github.com/firebase/firebase-cpp-sdk/actions/runs/3666797810)
[Packaging](https://github.com/firebase/firebase-cpp-sdk/actions/runs/3742396565) -> [Integration Test](https://github.com/firebase/firebase-cpp-sdk/actions/runs/3743731562)
***

### Type of Change
***

#### Notes


app/src/semaphore.h
```diff
@@ -50,29 +51,43 @@ class Semaphore {
 #if FIREBASE_PLATFORM_OSX || FIREBASE_PLATFORM_IOS || FIREBASE_PLATFORM_TVOS
     // MacOS requires named semaphores, and does not support unnamed.
     // Generate a unique string for the semaphore name:
-    static const char kPrefix[] = "/firebase-";
-    // String length of the name prefix.
-    static const int kPprefixLen = sizeof(kPrefix);
```
jonsimantov, line 55: Pprefix, with two P's?

app/src/semaphore.h
```diff
@@ -50,29 +51,43 @@ class Semaphore {
...
+    snprintf(buffer, kBufferSize, "%016llx",
              static_cast<unsigned long long>(  // NOLINT
                  reinterpret_cast<intptr_t>(this)));
 
-    semaphore_ = sem_open(buffer,
+#if FIREBASE_PLATFORM_OSX
+    // Sandbox mode is currently only available for macOS.
+    std::string semaphore_name = util::GetSandboxModeSemaphorePrefix();
+    if (semaphore_name.empty()) {
+      semaphore_name = "/firebase-";
```
jonsimantov, line 69: Is it possible to check at application start whether we are able to create a semaphore, and if not, print a loud fatal error to the log? (Similar to what happens with the iOS SDK when you don't have a REVERSED_CLIENT_ID URL handler set up.)

jonsimantov (approved):
Generally approved, but I had a few comments - it would be okay to address some in a followup.

app/src/semaphore.h
DellaBitta: Do you have a link to that check?

jonsimantov: https://github.com/firebase/firebase-ios-sdk/blob/eb30b6fdd809cbfe3a9eb7f3ccf1dd46326bef2d/FirebaseAuth/Sources/AuthProvider/OAuth/FIROAuthProvider.m#L139

app/src/semaphore.h
DellaBitta: That's what it was.. yeah. :)

app/src/semaphore.h
```diff
@@ -50,44 +51,61 @@ class Semaphore {
...
```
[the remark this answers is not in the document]
DellaBitta: Done.

app/src/semaphore.h
DellaBitta: Got it, thanks. I thought the example would have been more tightly integrated into the Firebase App initialization so I could use it as an example.

Trying my own thing. We'll see how it works out.

app/src/semaphore.h
DellaBitta: It's the Prefix prefix.

app/src/semaphore.h
```diff
@@ -61,24 +64,31 @@ class Semaphore {
...
     semaphore_ = sem_open(semaphore_name.c_str(),
                           O_CREAT | O_EXCL,   // Create if it doesn't exist.
                           S_IRUSR | S_IWUSR,  // Only the owner can read/write.
                           initial_count);
 
+    // Check for errors.
+#if FIREBASE_PLATFORM_OSX
+    FIREBASE_ASSERT_MESSAGE(
+        semaphore_ != SEM_FAILED,
+        "Failed to create semaphore. If running in sandbox mode be sure to "
```
jonsimantov, line 88: Maybe it would be good to include "Firebase" in the message?

DellaBitta: Done.

jonsimantov (approved):

PR #1172, opened 2022-12-16
Token-changed listeners for iOS AppCheck

AlmostMatt:
### Description
Implements token-changed listeners for the iOS implementation of AppCheck

### Testing
Added integration test cases to verify token-changed notification is received and token-changed listeners are called.
Integration tests passed when running locally.
***

### Type of Change
***

#### Notes


app_check/src/ios/app_check_ios.mm
```diff
@@ -91,13 +92,60 @@ - (id)initWithProviderFactory:(firebase::app_check::AppCheckProviderFactory* _No
 
 @end
 
+@implementation AppCheckNotificationCenterWrapper
+
+// A collection of NSValues containing AppCheckListener*
```
a-maurice, line 97: Should this be in the header?

app_check/src/ios/app_check_ios.mm
```diff
@@ -91,13 +92,60 @@ - (id)initWithProviderFactory:(firebase::app_check::AppCheckProviderFactory* _No
...
+
 namespace firebase {
 namespace app_check {
 namespace internal {
 
 AppCheckInternal::AppCheckInternal(App* app) : app_(app) {
   future_manager().AllocFutureApi(this, kAppCheckFnCount);
   impl_ = MakeUnique<FIRAppCheckPointer>([FIRAppCheck appCheck]);
+  notification_center_wrapper_ = MakeUnique<AppCheckNotificationCenterWrapperPointer>(
+      [[AppCheckNotificationCenterWrapper alloc] init]);
```
a-maurice, line 148: You likely need to unsubscribe in the destructor.  That might be why you are getting multiple responses back in the tests.

app_check/src/ios/app_check_ios.mm
AlmostMatt: I'm a bit fuzzy on objective-C, but my understanding is that by defining NSMutableArray* listeners_ in the implementation block it is a private variable, whereas if it was in the interface it would be public. And it seems like an implementation detail that should be private.

app_check/src/ios/app_check_ios.mm
AlmostMatt: Done. I also updated the test's TearDown to delete the AppCheck instance which will internally unsubscribe the notification center observer.
After unsubscribing the notification observer, I am seeing the correct number of notifications in integration tests.

a-maurice (approved):

PR #1174, opened 2022-12-19
Firestore COUNT API

tom-andersen:
***

### Type of Change
***


firestore/src/include/firebase/firestore/aggregate_query.h
```diff
@@ -0,0 +1,158 @@
...
+class Query;
+
+/**
+ * @brief A query that calculates aggregations over an underlying query.
+ */
+class AggregateQuery {
+ public:
+  /**
+   * @brief AggregateQuery an invalid Query that has to be reassigned before it
+   * can be used.
```
dconeybe, line 43: Something in this wording is off. Can you fix it?

```
AggregateQuery an invalid Query that has to be reassigned before it can be used.
```

firestore/src/include/firebase/firestore/aggregate_query_snapshot.h
```diff
@@ -0,0 +1,157 @@
...
+ * @brief A QuerySnapshot contains zero or more DocumentSnapshot objects.
+ *
+ * QuerySnapshot can be iterated using a range-based for loop, and its size can
+ * be inspected with empty() and size().
+ *
+ * @note Firestore classes are not meant to be subclassed except for use in test
+ * mocks. Subclassing is not supported in production code and new SDK releases
+ * may break code that does so.
+ */
+class AggregateQuerySnapshot {
```
dconeybe, line 39: This class-level comment appears to have been copied from the QuerySnapshot class. Can you fix it to be appropriate for the AggregateQuerySnapshot class?

firestore/src/include/firebase/firestore/aggregate_query_snapshot.h
```diff
@@ -0,0 +1,157 @@
...
+ public:
+  /**
+   * @brief AggregateQuerySnapshot an invalid Query that has to be reassigned
+   * before it can be used.
```
dconeybe, line 43: This wording is also weird: `AggregateQuerySnapshot an invalid Query that has to be reassigned
* before it can be used.` Please fix.

firestore/src/common/aggregate_query.cc
```diff
@@ -0,0 +1,108 @@
...
+AggregateQuery::AggregateQuery(AggregateQuery&& query) {
+  CleanupFnQuery::Unregister(&query, query.internal_);
+  std::swap(internal_, query.internal_);
+  CleanupFnQuery::Register(this, internal_);
+}
+
+AggregateQuery::AggregateQuery(AggregateQueryInternal* internal)
+    : internal_(internal) {
+  // NOTE: We don't assert internal != nullptr here since internal can be
+  // nullptr when called by the CollectionReference copy constructor.
```
dconeybe, line 52: This inline comment is not correct because this constructor is _never_ called from the "CollectionReference" copy constructor. Maybe just say that the assertion is not validated because nullptr is an acceptable value for internal_.

Also in `aggregate_query_snapshot.cc`.

firestore/src/common/aggregate_query_snapshot.cc
```diff
@@ -0,0 +1,107 @@
...
+AggregateQuery AggregateQuerySnapshot::query() const {
+  if (!internal_) return {};
+  return internal_->query();
+}
+
+int64_t AggregateQuerySnapshot::count() const {
+  if (!internal_) return 0;
+  return internal_->count();
+}
+
```
dconeybe, line 105: It looks like the definition of `operator==` is missing from for AggregateQuerySnapshot.

firestore/src/main/aggregate_query_main.h
```diff
@@ -0,0 +1,64 @@
...
+#error "This header should not be used on Android."
+#endif
+
+namespace firebase {
+namespace firestore {
+
+class AggregateQueryInternal {
+ public:
+  explicit AggregateQueryInternal(api::AggregateQuery&& aggregate_query);
+  ~AggregateQueryInternal() = default;
```
dconeybe, line 34: Is there a reason to explicitly declare the destructor? I think this can just be omitted.

firestore/src/main/aggregate_query_main.cc
```diff
@@ -0,0 +1,79 @@
...
+Query AggregateQueryInternal::query() {
+  return MakePublic(api::Query(aggregate_query_.query()));
+}
+
+Future<AggregateQuerySnapshot> AggregateQueryInternal::Get(
+    AggregateSource source) {
+  auto promise =
+      promise_factory_.CreatePromise<AggregateQuerySnapshot>(AsyncApis::kGet);
+  // TODO(tomandersen) Is there a better way to pass `aggregate_query_` than
+  // through `this`
```
dconeybe, line 55: RE: "Is there a better way to pass `aggregate_query_` than through `this`"...

Yes, and I believe you _should_ do it differently because `this` may become invalid before the callback gets invoked.

The way I'd do it is create a local variable, `auto aggregate_query = aggregate_query_` then pass it by value into the lambda using the lambda's capture clause, instead of passing `this` in the lambda's capture clause.

firestore/src/main/aggregate_query_snapshot_main.h
```diff
@@ -0,0 +1,61 @@
...
+
+  AggregateQuery query() const;
+  int64_t count() const;
+
+  friend bool operator==(const AggregateQuerySnapshotInternal& lhs,
+                         const AggregateQuerySnapshotInternal& rhs);
+
+ private:
+  api::AggregateQuery aggregate_query_;
+  int64_t count_result;
```
dconeybe, line 50: nit: `count_result` -> `count_result_` (append a trailing underscore)

firestore/src/main/aggregate_query_snapshot_main.cc
```diff
@@ -0,0 +1,45 @@
...
+
+#include "firestore/src/main/converter_main.h"
+#include "firestore/src/main/util_main.h"
+
+namespace firebase {
+namespace firestore {
+
+AggregateQuerySnapshotInternal::AggregateQuerySnapshotInternal(
+    api::AggregateQuery aggregate_query, int64_t count_result)
+    : aggregate_query_(aggregate_query), count_result(count_result) {}
```
dconeybe, line 27: nit: `aggregate_query` -> `std::move(aggregate_query)` in the initializer for `aggregate_query_`.

firestore/src/main/aggregate_query_snapshot_main.h
```diff
@@ -0,0 +1,61 @@
...
+#if defined(__ANDROID__)
+#error "This header should not be used on Android."
+#endif
+
+namespace firebase {
+namespace firestore {
+
+class AggregateQuerySnapshotInternal {
+ public:
+  explicit AggregateQuerySnapshotInternal(api::AggregateQuery aggregate_query,
```
dconeybe, line 36: Optional: Consider changing the `api::AggregateQuery` argument to an rvalue reference (i.e. `api::AggregateQuery&&`) to be consistent with AggregateQueryInternal's constructor.

firestore/src/main/aggregate_query_snapshot_main.cc
```diff
@@ -0,0 +1,45 @@
...
+const FirestoreInternal* AggregateQuerySnapshotInternal::firestore_internal()
+    const {
+  return GetFirestoreInternal(&aggregate_query_.query());
+}
+
+AggregateQuery AggregateQuerySnapshotInternal::query() const {
+  return MakePublic(api::AggregateQuery(aggregate_query_));
+}
+
+int64_t AggregateQuerySnapshotInternal::count() const { return count_result; }
```
dconeybe, line 42: The definition of `operator==` for `AggregateQuerySnapshotInternal` is missing.

dconeybe (changes requested):
Overall, looks pretty straight-forward. Just a bunch of nits from me.

firestore/src/main/aggregate_query_main.h
tom-andersen: No. Removed.

firestore/src/main/aggregate_query_snapshot_main.h
tom-andersen: Renamed.

firestore/src/main/aggregate_query_snapshot_main.h
tom-andersen: Done

release_build_files/readme.md
```diff
@@ -642,6 +642,9 @@ workflow use only during the development of your app, not for publicly shipping
 code.
 
 ## Release Notes
+### Upcoming Release
+    - Firestore: Add Count API.
```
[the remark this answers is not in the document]
tom-andersen: Done

firestore/src/main/aggregate_query_main.cc
```diff
@@ -0,0 +1,78 @@
...
+  return MakePublic(api::Query(aggregate_query_.query()));
+}
+
+Future<AggregateQuerySnapshot> AggregateQueryInternal::Get(
+    AggregateSource source) {
+  auto aggregate_query = aggregate_query_;
+  auto promise =
+      promise_factory_.CreatePromise<AggregateQuerySnapshot>(AsyncApis::kGet);
+  aggregate_query_.Get(
+      [aggregate_query, promise](util::StatusOr<int64_t> maybe_value) mutable {
```
dconeybe, line 56: Add a `//TODO(C++14)` comment above this line to capture `aggregate_query_` directly.

Namely, this

```
auto aggregate_query = aggregate_query_;
...
[aggregate_query, promise](util::StatusOr<int64_t> maybe_value) mutable {
```

can be replaced by this


```
[aggregate_query = aggregate_query_, promise](util::StatusOr<int64_t> maybe_value) mutable {
```

once we upgrade to c++14 (which will be happening in the foreseeable future).

https://en.wikipedia.org/wiki/C%2B%2B14#Lambda_capture_expressions

firestore/src/main/aggregate_query_main.h
```diff
@@ -0,0 +1,63 @@
...
+
+  Future<AggregateQuerySnapshot> Get(AggregateSource source);
+
+  friend bool operator==(const AggregateQueryInternal& lhs,
+                         const AggregateQueryInternal& rhs);
+
+ private:
+  enum class AsyncApis {
+    kGet,
+    kCount,
```
dconeybe, line 48: `kCount` can be removed from `AsyncApis`

dconeybe (changes requested):

firestore/src/main/aggregate_query_main.cc
```diff
@@ -0,0 +1,78 @@
...
+        if (maybe_value.ok()) {
+          int64_t count = maybe_value.ValueOrDie();
+          AggregateQuerySnapshotInternal internal =
+              AggregateQuerySnapshotInternal(aggregate_query, count);
```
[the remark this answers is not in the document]
tom-andersen: Nice. Still not quite used to this syntax.

firestore/src/main/aggregate_query_main.h
tom-andersen: I don't exactly understand why this is required, but keeping `kCount` is intentional.

https://github.com/firebase/firebase-cpp-sdk/blob/dfa4248bc7310152db5dc92c5c973b4700d7b6d7/firestore/src/main/promise_factory_main.h#L37-L41

release_build_files/readme.md
```diff
@@ -643,7 +643,9 @@ code.
 
 ## Release Notes
 ### Upcoming Release
-    - Firestore: Add Count API.
+-   Changes
+    - Added `Query::Count()`, which fetches the number of documents in the result
```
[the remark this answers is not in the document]
tom-andersen: Added

firestore/src/main/query_main.cc
```diff
@@ -94,6 +96,10 @@ Future<QuerySnapshot> QueryInternal::Get(Source source) {
   return promise.future();
 }
 
+AggregateQuery QueryInternal::Count() {
+  return MakePublic(std::move(query_.Count()));
```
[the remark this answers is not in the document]
tom-andersen: Thanks for explanation. Changed.

firestore/src/common/aggregate_query_snapshot.cc
```diff
@@ -0,0 +1,113 @@
...
+namespace firebase {
+namespace firestore {
+
+using CleanupFnAggregateQuerySnapshot = CleanupFn<AggregateQuerySnapshot>;
+
+AggregateQuerySnapshot::AggregateQuerySnapshot() {}
+
+AggregateQuerySnapshot::AggregateQuerySnapshot(
+    const AggregateQuerySnapshot& query) {
```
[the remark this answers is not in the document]
dconeybe: Please use `other` for the parameter names of all copy/move constructors and assignment operators, in both the .h and .cc files.

Personally, I like to omit parameter names in the header file in cases where the name does not add any additional value.

For example, instead of

```cpp
AggregateQuerySnapshot(const AggregateQuerySnapshot& other);
```

I like to omit `other` and just use

```cpp
AggregateQuerySnapshot(const AggregateQuerySnapshot&);
```

Then you don't have to worry about keeping the names in sync.

But if you _do_ choose to use a parameter name, they should match in the .h and .cc files. This is something that clang-tidy checks for, and is even enforced in google3: https://clang.llvm.org/extra/clang-tidy/checks/readability/inconsistent-declaration-parameter-name.html.

But most of all, it's best to follow local conventions for parameter names, and other places use `other`, like in [query.h](https://github.com/firebase/firebase-cpp-sdk/blob/b1a5444b10c751c4c039a29bb9fe531b39d1b9ad/firestore/src/include/firebase/firestore/query.h#L90) and [query_snapshot.h](https://github.com/firebase/firebase-cpp-sdk/blob/b1a5444b10c751c4c039a29bb9fe531b39d1b9ad/firestore/src/include/firebase/firestore/query_snapshot.h#L64).

firestore/src/common/aggregate_query.cc
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright 2022 Google LLC
```
dconeybe, line 2: nit: 2022 -> 2023, here and elsewhere in this PR.

firestore/src/common/aggregate_query.cc
```diff
@@ -0,0 +1,108 @@
...
+#if defined(__ANDROID__)
+// TODO(tomandersen) #include "firestore/src/android/aggregate_query_android.h"
+#else
+#include "firestore/src/main/aggregate_query_main.h"
+#endif  // defined(__ANDROID__)
+
+namespace firebase {
+namespace firestore {
+
+using CleanupFnQuery = CleanupFn<AggregateQuery>;
```
dconeybe, line 32: nit: `CleanupFnQuery` -> `CleanupFnAggregateQuery`

dconeybe (changes requested):

firestore/src/common/aggregate_query_snapshot.cc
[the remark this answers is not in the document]
tom-andersen: I am going to guess that omitting parameter name in header files is a bad idea for public facing APIs. I think there is a direct consequence in generation of documentation. Note the details table that makes reference to parameter name. So if the comments make reference to parameter name, then omission is not an option. 

https://firebase.google.com/docs/reference/cpp/class/firebase/firestore/query-snapshot#classfirebase_1_1firestore_1_1_query_snapshot_1a092feef2ecf7942404beaf56fe7b7b00

I'll make sure header and code files are consistent in my work, but I think we have some cleanup of existing classes if clang-tidy is ever enforced on repo.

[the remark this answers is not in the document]
dconeybe: That's a good point about needing a parameter name in order to document it.

dconeybe (approved):

PR #1176, opened 2022-12-25
Implement ObjectArena and ArenaRef to solve JNI Global Reference Issue

cherylEnkidu:
***
***

### Type of Change
- [x] Bug fix. ([#569](https://github.com/firebase/firebase-unity-sdk/issues/569))([#1193](https://github.com/firebase/quickstart-unity/issues/1193#issuecomment-1242203289))
***

#### Notes


firestore/src/android/field_value_android.cc
```diff
@@ -98,39 +99,50 @@ FieldValue FieldValueInternal::Create(Env& env,
...
+  Env env;
+  object_ = ArenaRef(env, object);
+}
 
 FieldValueInternal::FieldValueInternal(bool value)
     : cached_type_(Type::kBoolean) {
-  Env env = GetEnv();
-  object_ = Boolean::Create(env, value);
+  Env env;
+  auto boxed_ptr = Boolean::Create(env, value);
```
dconeybe, line 116: nit: Remove the `boxed_ptr` local variable and just specify `Boolean::Create(env, value)` as an argument to the ArenaRef constructor.

IMO the `boxed_ptr` local variable just adds noise and detracts from readability. Please make this change elsewhere in this file too.

firestore/src/android/field_value_android.cc
```diff
@@ -176,63 +191,65 @@ FieldValueInternal::FieldValueInternal(const MapFieldValue& value)
...
+  Env env;
+  if (!object_.get(env)) {
     return Type::kNull;
   }
 
   // We do not have any knowledge on the type yet. Check the runtime type with
   // each known type.
-  Env env = GetEnv();
-  if (env.IsInstanceOf(object_, Boolean::GetClass())) {
+  if (env.IsInstanceOf(object_.get(env), Boolean::GetClass())) {
```
dconeybe, line 209: Similarly, I think we should just override IsInstanceOf to take ArenaRef as well, like this:

```
bool Env::IsInstanceOf(const ArenaRef& object, const Class& clazz) {
  return IsInstanceOf(object.get(*this), clazz);
}
```
```

File: firestore/src/android/field_value_android.cc [hunk 3]
```diff
@@ -389,15 +406,24 @@ bool operator==(const FieldValueInternal& lhs, const FieldValueInternal& rhs) {
...
+  if (cached_type_ == Type::kNull) {
+    FIREBASE_ASSERT(env.IsInstanceOf(object_.get(env), T::GetClass()));
+    cached_type_ = type;
+  } else {
+    FIREBASE_ASSERT(cached_type_ == type);
+  }
+  return object_.get(env).CastTo<T>();
+}
+
+Local<String> FieldValueInternal::CastString(jni::Env& env, Type type) const {
```
dconeybe, line 419: You can get rid of this special `CastString` method. Here is what worked for me:

```
template <typename T>
T FieldValueInternal::Cast(jni::Env& env, Type type) const {
  if (cached_type_ == Type::kNull) {
    FIREBASE_ASSERT(env.IsInstanceOf(object_, T::GetClass()));
    cached_type_ = type;
  } else {
    FIREBASE_ASSERT(cached_type_ == type);
  }
  auto untyped_value = object_.get(env);
  auto typed_value = static_cast<jni::JniType<T>>(untyped_value.get());
  return T(typed_value);
}
```

Note that this assumes that you overrided Env::IsInstanceOf to take ArenaRef.

firestore/src/android/field_value_android.h
```diff
@@ -87,7 +90,10 @@ class FieldValueInternal {
   std::vector<FieldValue> array_value() const;
   MapFieldValue map_value() const;
 
-  const jni::Global<jni::Object>& ToJava() const { return object_; }
+  jni::Local<jni::Object> ToJava() const {
```
dconeybe, line 93: nit: Move the `ToJava()` implementation into the .cc file, now that it is not a simple "getter" method.

firestore/src/android/wrapper.h
```diff
@@ -44,9 +45,10 @@ class Wrapper {
   Wrapper& operator=(Wrapper&& wrapper) = delete;
 
   FirestoreInternal* firestore_internal() { return firestore_; }
-  const jni::Object& ToJava() const { return obj_; }
-
-  static jni::Object ToJava(const FieldValue& value);
+  jni::Local<jni::Object> ToJava() const {
```
dconeybe, line 48: Similarly, move Wrapper::ToJava() to the .cc file as well, since it is no longer a simple "getter" method.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,114 @@
...
+namespace firebase {
+namespace firestore {
+namespace jni {
+
+namespace {
+
+HashMap* gArenaRefHashMap = nullptr;
+
+int64_t GetNextArenaRefKey() {
+  static std::atomic<int64_t> next_key(1);
```
dconeybe, line 33: I think we should initialize `next_key` to `0` instead of `1`. We should use `-1` as an "invalid" key, rather than using the "default" int64_t value as invalid.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,114 @@
...
+
+ArenaRef::ArenaRef(Env& env, const Object& object)
+    : key_(GetNextArenaRefKey()) {
+  gArenaRefHashMap->Put(env, key_object(env), object);
+}
+
+ArenaRef::ArenaRef(const ArenaRef& other) : key_(GetNextArenaRefKey()) {
+  if (other.key_ != 0) {
+    Env env;
+    Local<Object> object = other.get(env);
```
dconeybe, line 47: nit: Use FirestoreInternal::GetEnv() rather than using the default-initialized Env object. Please make this change everywhere in this PR, including other places in this file.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,114 @@
...
+    gArenaRefHashMap->Put(env, key_object(env), object);
+  }
+}
+
+ArenaRef::ArenaRef(ArenaRef&& other) {
+  key_ = other.key_;
+  other.key_ = 0;
```
dconeybe, line 54: nit: Use -1 as the "invalid" key.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,114 @@
...
+
+ArenaRef& ArenaRef::operator=(const ArenaRef& other) {
+  Env env;
+
+  if (key_ != other.key_) {
+    if (key_ != 0) {
+      gArenaRefHashMap->Remove(env, key_object(env));
+    }
+
+    if (other.key_ != 0) {
```
dconeybe, line 65: Add an "else" block to set key_=-1.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,114 @@
...
+
+ArenaRef::ArenaRef(ArenaRef&& other) {
+  key_ = other.key_;
+  other.key_ = 0;
+}
+
+ArenaRef& ArenaRef::operator=(const ArenaRef& other) {
+  Env env;
+
+  if (key_ != other.key_) {
```
dconeybe, line 60: Rather than checking if `if (key_ != other.key_)`, check if `this == &other`. This is more conventional in copy assignment operator overloads. In fact, I'd put that as the first thing in this method, to reduce nesting and improve readability.

i.e.

```
ArenaRef& ArenaRef::operator=(const ArenaRef& other) {
  if (this == &other) {
    return *this;
  }
  ...
```

Please do the same thing for the move assignment operator.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,114 @@
...
+Local<Long> ArenaRef::key_object(Env& env) const {
+  return Long::Create(env, key_);
+}
+
+void ArenaRef::Initialize(Env& env) {
+  if (gArenaRefHashMap) {
+    return;
+  }
+  Global<HashMap> hash_map(HashMap::Create(env));
+  jobject hash_map_jobject = hash_map.release();
```
dconeybe, line 104: nit: Remove this line and just use `new HashMap(hash_map.release());`. IMO this improves readability because it avoids an unnecessary local variable.

firestore/src/jni/arena_ref.h
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+#ifndef FIREBASE_FIRESTORE_SRC_JNI_ARENA_REF_H_
+#define FIREBASE_FIRESTORE_SRC_JNI_ARENA_REF_H_
+
+#include <jni.h>
+
+#include <cstdlib>
+
+#include "firestore/src/jni/env.h"
```
dconeybe, line 24: Rather than including env.h, include jni_fwd.h, which forward-declares classes from the `jni` namespace.

ALSO, add ArenaRef to jni_fwd.h. I almost forgot about that.

firestore/src/jni/arena_ref.h
```diff
@@ -0,0 +1,56 @@
...
+#include "firestore/src/jni/object.h"
+#include "firestore/src/jni/ownership.h"
+
+namespace firebase {
+namespace firestore {
+namespace jni {
+
+class ArenaRef {
```
dconeybe, line 32: Please add some class-level documentation about ArenaRef. This documentation should explain why you would use ArenaRef instead of Global and that it has a performance cost compared to Global.

firestore/src/jni/arena_ref.h
```diff
@@ -0,0 +1,56 @@
...
+  ~ArenaRef();
+
+  static void Initialize(Env&);
+
+  Local<Object> get(Env&) const;
+
+ private:
+  Local<Long> key_object(Env&) const;
+
+  int64_t key_ = 0;
```
dconeybe, line 49: Initialize key_ to -1 instead of 0.

dconeybe (changes requested):
Here are my initial review comments. Overall, looks great!

firestore/src/android/field_value_android.cc
dconeybe: Hmm, on second thought this might crash. If it does crash, then instead of creating a "CastString" method, create a specialization of the Cast template method for String.

firestore/integration_test_internal/src/android/arena_ref_android_test.cc
```diff
@@ -0,0 +1,133 @@
...
+
+TEST_F(ArenaRefTestAndroid, ConstructFromEnvAndObject) {
+  Env env;
+  Local<String> string = env.NewStringUtf("hello world");
+
+  ArenaRef arena_ref(env, string);
+  EXPECT_TRUE(arena_ref.get(env).Equals(env, string));
+}
+
+TEST_F(ArenaRefTestAndroid, CopyConstructor) {
```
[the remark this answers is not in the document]
cherylEnkidu: I have a question: Both of them invokes copy ctor function, IMO they are testing one thing.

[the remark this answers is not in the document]
dconeybe: I disagree. They literally invoke the exact same constructor so there is no reason to test both of them. By testing both of them you are essentially testing the C++ compiler to make sure it calls the right constructor. The tests are supposed to be testing _your_ code, not the compiler. If you feel strongly that you'd like to leave in both ways of invoking the copy constructor then please just add a comment to explicitly state what you are doing. Otherwise, to me,  it was confusing because it looked like there was some confusion about how C++ works, which is not the case.

firestore/src/android/field_value_android.cc
```diff
@@ -98,39 +99,50 @@ FieldValue FieldValueInternal::Create(Env& env,
 FieldValueInternal::FieldValueInternal() : cached_type_(Type::kNull) {}
 
 FieldValueInternal::FieldValueInternal(const Object& object)
-    : object_(object), cached_type_(Type::kNull) {}
+    : cached_type_(Type::kNull) {
+  Env env;
```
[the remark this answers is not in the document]
cherylEnkidu: But doesn't the ctor for Env do the same thing? `Env::Env() : Env(GetEnv()) {}`

firestore/src/android/field_value_android.cc
cherylEnkidu: Do I need to write tests for this method in `arena_ref_android_test.cc`?

firestore/src/android/field_value_android.cc
cherylEnkidu: I don't think the above solution would work since typed_value contains a local reference jobject and it would be deleted when the function returns.

For template specialization, I am not sure what it would look like. `CastString` doesn't have different parameters compare to `Cast`. It can get rid of the second parameter but still need to have `String` inside the name for the readability.

firestore/src/android/field_value_android.cc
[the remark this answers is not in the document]
dconeybe: The GetEnv() defined in jni.h returns the raw `JNIEnv*` pointer. There is another GetEnv().

firestore/src/android/field_value_android.cc
dconeybe: We should add a test file for env and just add tests for the new stuff that you've added to env.h.

firestore/src/android/field_value_android.cc
dconeybe: A template specialization would look something like this:

```cpp
template <>
Local<String> FieldValueInternal::Cast<String>(jni::Env& env, Type type) const {
  ...
}
```

firestore/integration_test_internal/src/android/arena_ref_android_test.cc
```diff
@@ -42,89 +59,114 @@ TEST_F(ArenaRefTestAndroid, ConstructFromEnvAndObject) {
...
 
-  arena_ref4 = std::move(arena_ref1);
-  EXPECT_EQ(arena_ref4.get(env).get(), nullptr);
+  ArenaRef hash_map(env, HashMap::Create(env));
+  Local<Long> key = Long::Create(env, 1);
+  Local<Long> val = Long::Create(env, 2);
+  env.Call(hash_map, kPut, key, val);
+  Local<Object> result = env.Call(hash_map, kGet, key);
+  EXPECT_TRUE(result.Equals(env, val));
+  EXPECT_TRUE(env.ok());
```
dconeybe, line 169: This line, `EXPECT_TRUE(env.ok()` can be removed.

The FirestoreAndroidIntegrationTest will fail the test if the test ends with env.ok()==false.

https://github.com/firebase/firebase-cpp-sdk/blob/e61d6bb264633c720b1ce857717f4e9638f40279/firestore/integration_test_internal/src/android/firestore_integration_test_android.cc#L70-L80

firestore/integration_test_internal/src/android/arena_ref_android_test.cc
```diff
@@ -0,0 +1,175 @@
...
+
+#include "gmock/gmock.h"
+#include "gtest/gtest.h"
+
+namespace firebase {
+namespace firestore {
+namespace jni {
+namespace {
+
+jclass clazz = nullptr;
```
dconeybe, line 32: This variable `jclass clazz` doesn't appear to be used anywhere. Can it be removed?

firestore/src/android/document_reference_android.cc
```diff
@@ -198,9 +198,9 @@ ListenerRegistration DocumentReferenceInternal::AddSnapshotListener(
   Local<Object> java_listener =
       EventListenerInternal::Create(env, firestore_, listener);
 
-  Local<Object> java_registration =
-      env.Call(obj_, kAddSnapshotListener, firestore_->user_callback_executor(),
-               java_metadata, java_listener);
+  Local<Object> java_registration = env.Call(
+      obj_.get(env), kAddSnapshotListener, firestore_->user_callback_executor(),
```
dconeybe, line 202: Can `obj_.get(env)` be replaced with just `obj_` here?

firestore/src/android/field_value_android.cc
```diff
@@ -411,8 +429,15 @@ Local<Array<Object>> FieldValueInternal::MakeArray(
...
-  return value.internal_ ? value.internal_->object_ : Object();
+Local<Object> FieldValueInternal::ToJava() const {
+  Env env;
+  return object_.get(env);
+}
+
+Local<Object> FieldValueInternal::ToJava(const FieldValue& value) {
+  Env env;
+  return value.internal_ ? value.internal_->object_.get(env)
+                         : Local<Object>(Object());
```
dconeybe, line 440: I think you can omit the `Object()` argument to the `Local<Object>` constructor.

firestore/src/android/field_value_android.h
```diff
@@ -20,11 +20,14 @@
...
 #include <memory>
 #include <string>
+#include <vector>
 
 #include "firebase/firestore/geo_point.h"
 #include "firebase/firestore/timestamp.h"
 #include "firestore/src/include/firebase/firestore/document_reference.h"
 #include "firestore/src/include/firebase/firestore/field_value.h"
+#include "firestore/src/jni/arena_ref.h"
+#include "firestore/src/jni/env.h"
```
dconeybe, line 30: Remove this include of "env.h" if you can. I think it got added when ToJava() made use of Env. But now that ToJava() has moved to the .cc file the forward declaration in jni_fwd.h should be sufficient.

firestore/src/android/wrapper.cc
```diff
@@ -29,13 +29,16 @@ namespace firebase {
...
 using jni::Env;
 using jni::Object;
 
 }  // namespace
 
 Wrapper::Wrapper(FirestoreInternal* firestore, const Object& obj)
-    : firestore_(firestore), obj_(obj) {
+    : firestore_(firestore) {
+  Env env;
+  obj_ = ArenaRef(env, obj);
```
dconeybe, line 41: nit: Move the `FIREBASE_ASSERT` on line 42 to the top of this method. The assertion on `obj` should be done before using `obj`.

firestore/src/android/wrapper.cc
```diff
@@ -56,11 +59,12 @@ Wrapper::Wrapper(Wrapper* rhs) : Wrapper() {
 
 Wrapper::~Wrapper() = default;
 
-jni::Env Wrapper::GetEnv() const { return firestore_->GetEnv(); }
-
-Object Wrapper::ToJava(const FieldValue& value) {
-  return FieldValueInternal::ToJava(value);
+jni::Local<jni::Object> Wrapper::ToJava() const {
```
dconeybe, line 62: Remove the `jni::` qualifiers. This was required in the .h file, but the `using` declarations near the top of this file mean they're not needed here. Add new `using` declarations if you have to.

firestore/src/android/wrapper.h
```diff
@@ -17,6 +17,8 @@
 #ifndef FIREBASE_FIRESTORE_SRC_ANDROID_WRAPPER_H_
 #define FIREBASE_FIRESTORE_SRC_ANDROID_WRAPPER_H_
 
+#include "firestore/src/jni/arena_ref.h"
+#include "firestore/src/jni/env.h"
```
dconeybe, line 21: Similarly, remove the include of env.h from here as well.

firestore/src/common/field_value.cc
```diff
@@ -17,6 +17,7 @@
 #include "firestore/src/include/firebase/firestore/field_value.h"
 
 #include <iomanip>
+#include <iostream>
```
dconeybe, line 20: nit: Remove this line: `#include <iostream>`

firestore/src/jni/arena_ref.h
```diff
@@ -0,0 +1,65 @@
...
+ * JNI global reference (the pointer inside Global<T>) need to be under 51,200,
+ * since after that number the app will crash. The drawback is ArenaRef holds
+ * JNI Reference indirectly using HashMap so there is a efficiency lost against
+ * Global<T>.
+ */
+class ArenaRef {
+ public:
+  ArenaRef() = default;
+  ArenaRef(Env&, const Object&);
+  ArenaRef(const ArenaRef& other);
```
dconeybe, line 45: Remove the "other" argument names. IMO they are just distracting and are irrelevant in the declaration.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,119 @@
...
+}
+
+}  // namespace
+
+ArenaRef::ArenaRef(Env& env, const Object& object)
+    : key_(GetNextArenaRefKey()) {
+  gArenaRefHashMap->Put(env, key_object(env), object);
+}
+
+ArenaRef::ArenaRef(const ArenaRef& other) : key_(GetNextArenaRefKey()) {
```
dconeybe, line 45: `key_` should be initialized to `-1` if `other.key_ != -1`. This technically works because the HashMap will not have an object associated with the given key; however, it is misleading to leave a "dangling key" that does not exist in the hash map.

Consider initializing `key_` as follows:

```
key_(other.key_ == -1 ? -1 : GetNextArenaRefKey())
```

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,119 @@
...
+  if (other.key_ != -1) {
+    Env env;
+    Local<Object> object = other.get(env);
+    gArenaRefHashMap->Put(env, key_object(env), object);
+  }
+}
+
+ArenaRef::ArenaRef(ArenaRef&& other) {
```
dconeybe, line 53: In the ArenaRef move constructor, initialize `key_` using `key_(key_ = other.key_)` (like how `key_` is initialized in the other constructors) rather than letting it get initialized to the default value of `-1` then setting it to another value in the constructor.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,119 @@
...
+
+ArenaRef& ArenaRef::operator=(const ArenaRef& other) {
+  Env env;
+
+  if (this == &other) {
+    return *this;
+  }
+
+  if (key_ != -1) {
+    gArenaRefHashMap->Remove(env, key_object(env));
```
dconeybe, line 66: Move the code that sets `key_ = -1` to right here, after calling `gArenaRefHashMap->Remove()`. This ensures that there is no point in time during which `key_` is pointing to an empty entry. I realize this slightly contradicts my former comment :)

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,119 @@
...
+  }
+  key_ = other.key_;
+  other.key_ = -1;
+  return *this;
+}
+
+ArenaRef::~ArenaRef() {
+  if (key_ != -1) {
+    Env env;
+    gArenaRefHashMap->Remove(env, key_object(env));
```
dconeybe, line 97: Use `ExceptionClearGuard` to verify that the destructor will remove the entry from the HashMap even if there is a pending exception.

https://github.com/firebase/firebase-cpp-sdk/blob/e61d6bb264633c720b1ce857717f4e9638f40279/firestore/src/jni/env.h#L543-L566

firestore/src/jni/arena_ref.h
```diff
@@ -0,0 +1,65 @@
...
+  ~ArenaRef();
+
+  static void Initialize(Env&);
+
+  Local<Object> get(Env&) const;
+
+ private:
+  Local<Long> key_object(Env&) const;
+
+  int64_t key_ = -1;
```
dconeybe, line 58: The more I think about it, the more I don't like this "magic value" of -1 (formerly 0) for `key_`. IMO it detracts from readability and maintainability where you see this comparison to -1 sprinkled through the code. I think a better solution is to assign `key_` once in the constructor, and then never change it. The exception to this rule would be in the move constructor and assignment operator, which could just use `std::swap` on the key_.). Then, use a boolean flag `valid_` which is true only if `key_` maps to an entry in the HashMap. What do you think?

firestore/integration_test_internal/src/android/arena_ref_android_test.cc
```diff
@@ -0,0 +1,175 @@
...
+  ArenaRef arena_ref3(env, string2);
+  arena_ref3 = std::move(arena_ref2);
+  EXPECT_EQ(arena_ref2.get(env).get(), nullptr);
+  EXPECT_TRUE(arena_ref3.get(env).Equals(env, string1));
+
+  arena_ref3 = std::move(arena_ref1);
+  EXPECT_EQ(arena_ref3.get(env).get(), nullptr);
+}
+
+TEST_F(ArenaRefTestAndroid, EnvCallTakeArenaRefTest) {
```
dconeybe, line 160: Please move this test `EnvCallTakeArenaRefTest` into a new file, `env_android_test.cc`. You obviously don't need to write an exhaustive test suite for the Env class. Just write tests for the new methods that you are adding.

dconeybe (changes requested):

firestore/src/jni/arena_ref.h
dconeybe: I got curious and tried it out this morning and it seems to work perfectly.
https://gist.github.com/dconeybe/afdcd15cd56bebe5497c07cbb8ceb935

There was just one unit test that failed, but it was assuming that the moved-from object was "null". Unless we make some strong guarantees about the state of the moved-from object, we should probably not test anything about the moved-from object, except its destructor and re-assigning it using the operator= overloads. Formally, those are the only valid operations guaranteed by the C++ standard to perform on a moved-from object.

dconeybe: (I just realized that in that gist link I forgot to set `valid_ = other.valid_` in the copy assignment operator).

dconeybe:
I just realized something: the ArenaRef has a concurrency bug. If ArenaRefs are created/copied concurrently from multiple threads then there is a race condition that will cause bugs. This is because the multiple threads can be accessing the Java HashMap at the same time without any synchronization. This leads to undefined behavior, and could be causing that weird bug you're seeing in CI that you can't reproduce locally. One fix would be to protect `gArenaRefHashMap` with an `std::mutex`.

firestore/src/android/field_value_android.h
cherylEnkidu: But for `static jni::Env GetEnv();` this function, the return type is Env. We need to include the header file.

firestore/src/jni/arena_ref.h
cherylEnkidu: But I think `other` is conventional in functions like `copy ctor`, and it increases readability because it people will realize it is a copy ctor as soon as they see it.

firestore/src/jni/arena_ref.h
cherylEnkidu: I agree we shouldn't test the object after moving.

However, I think introduced another flag `valid` decrease the readability and maintainability. What I think is better is add MACRO like `invalid key` and set it to -1.

dconeybe: Ok fair enough. There is definitely an argument for readability either way.

firestore/integration_test_internal/src/android/arena_ref_android_test.cc
```diff
@@ -0,0 +1,148 @@
...
+#include "firestore_integration_test_android.h"
+
+#include "gtest/gtest.h"
+
+namespace firebase {
+namespace firestore {
+namespace jni {
+namespace {
+
+using ArenaRefTestAndroid = FirestoreIntegrationTest;
```
dconeybe, line 31: Please add tests for calling each method with a pending exception. You may want to use ExceptionClearGuard or just manually check for exceptions. The important thing is that we don't leak memory.

dconeybe (changes requested):

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,147 @@
...
+
+ArenaRef& ArenaRef::operator=(const ArenaRef& other) {
+  Env env = FirestoreInternal::GetEnv();
+
+  if (this == &other) {
+    return *this;
+  }
+
+  {
+    std::unique_lock<std::mutex> lock(mutex_);
```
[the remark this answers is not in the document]
cherylEnkidu: Do you think it is necessary to add lock to make sure `copy assignment` will finish in one step?

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,147 @@
...
+    if (key_ != -1) {
+      gArenaRefHashMap->Remove(env, key_object(env));
+      key_ = -1;
+    }
+
+    if (other.key_ != -1) {
+      key_ = GetNextArenaRefKey();
+      Local<Object> object = other.get(env);
+      gArenaRefHashMap->Put(env, key_object(env), object);
+    }
```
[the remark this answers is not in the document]
cherylEnkidu: But if in the case when other.key is not valid, we still want to remove the current object. Since the `arenaRef` copies from a empty `arenaRef`.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,147 @@
...
+    if (!env.ok()) {
+      env.ExceptionDescribe();
+      env.ExceptionClear();
+    }
+  }
+}
+
+Local<Long> ArenaRef::key_object(Env& env) const {
+  jobject key = env.get()->NewObject(gLongClass, gLongConstructor, key_);
```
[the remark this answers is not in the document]
cherylEnkidu: If we add a unit test to `key_object()`, then this function needs to be public, since the test class is inside anonymous namespace. To me `key_object ` is more like an internal implementation, as long as we test out the behaviour of `ArenaRef` is correct we don't need to worry about how it is implemented.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,147 @@
...
+  if (gArenaRefHashMap) {
+    return;
+  }
+  Global<HashMap> hash_map(HashMap::Create(env));
+  gArenaRefHashMap = new HashMap(hash_map.release());
+
+  gHashMapClass = env.get()->FindClass("java/util/HashMap");
+  gHashMapClass = static_cast<jclass>(env.get()->NewGlobalRef(gHashMapClass));
+
+  gLongClass = env.get()->FindClass("java/lang/Long");
```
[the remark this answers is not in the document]
cherylEnkidu: Just curious is there a reason why `gLongClass` prefer to not using twice?

firestore/src/jni/arena_ref.cc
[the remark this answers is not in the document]
dconeybe: No, we don't need to guarantee that the entire copy assignment operator completes atomically. We _only_ need to make sure that at most one thread at a time is executing a method on `gArenaRefHashMap`. That's the only thing that the lock needs to protect. So the code needs to acquire the lock before invoking a method on `gArenaRefHashMap` and release it immediately after. Everything else is thread safe.

dconeybe (changes requested):

Awais6:
Any update, please? Games crash rates are too high due to this issue, and Play Console's bad behavior threshold is crossed.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,147 @@
...
+    : key_(other.key_ == -1 ? -1 : GetNextArenaRefKey()) {
+  if (other.key_ != -1) {
+    Env env = FirestoreInternal::GetEnv();
+    Local<Object> object = other.get(env);
+    std::unique_lock<std::mutex> lock(mutex_);
+    gArenaRefHashMap->Put(env, key_object(env), object);
+  }
+}
+
+ArenaRef::ArenaRef(ArenaRef&& other) { std::swap(key_, other.key_); }
```
dconeybe, line 63: Please change this to:

```cpp
using std::swap;  // go/using-std-swap
swap(key_, other.key_);
```

I just stumbled across this advice. It probably doesn't technically matter here since we're swapping ints, but it's good to showcase best practices.

firestore/integration_test_internal/src/android/env_android_test.cc
```diff
@@ -0,0 +1,68 @@
...
+ public:
+  void SetUp() override {
+    FirestoreAndroidIntegrationTest::SetUp();
+    jclass g_clazz = util::map::GetClass();
+    loader().LoadFromExistingClass("java/util/HashMap", g_clazz, kGet, kPut);
+    ASSERT_TRUE(loader().ok());
+  }
+};
+
+TEST_F(EnvTestAndroid, EnvCallTakeArenaRefTest) {
```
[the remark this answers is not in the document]
cherylEnkidu: I try this test but it doesn't work, since the code crash before return to the cpp side:
```
TEST_F(EnvTest, CallsNullArenaRef) {
  ArenaRef null_arena_ref;

  Local<Class> clazz = env().FindClass("java/lang/String");
  jmethodID to_lower_case =
      env().GetMethodId(clazz, "toLowerCase", "()Ljava/lang/String;");

  try {
    Local<Object> result = env().Call(null_arena_ref, Method<Object>(to_lower_case));
    FAIL() << "A NullPointerException should have been thrown.";
  } catch (const std::exception& e) {
    Local<Throwable> exception = env().ClearExceptionOccurred();
    ASSERT_EQ(env().GetObjectClass(exception).GetName(env()), "java.lang.NullPointerException");
    SUCCEED() << "Caught exception";
  }

}
```
The error message is 
`'JNI DETECTED ERROR IN APPLICATION: can't call java.lang.String java.lang.String.toLowerCase() on null object in call to CallObjectMethodV'`

firestore/integration_test_internal/src/android/env_android_test.cc
```diff
@@ -0,0 +1,68 @@
...
+
+  ArenaRef hash_map(env, HashMap::Create(env));
+  Local<Long> key = Long::Create(env, 1);
+  Local<Long> val = Long::Create(env, 2);
+  env.Call(hash_map, kPut, key, val);
+  Local<Object> result = env.Call(hash_map, kGet, key);
+  EXPECT_TRUE(result.Equals(env, val));
+}
+
+TEST_F(EnvTestAndroid, EnvIsInstanceOfTakeArenaRefTest) {
```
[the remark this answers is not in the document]
cherylEnkidu: The following test would fail because when ArenaRef is invalid, `get()` returns `Object` and `Object` is equal to every Java class
```
TEST_F(EnvTest, IsInstanceOfChecksNullArenaRef) {
  Local<Class> clazz = env().FindClass("java/lang/System");
  ArenaRef arena_ref;
  EXPECT_FALSE(env().IsInstanceOf(arena_ref, clazz));
}
```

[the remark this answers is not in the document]
dconeybe: Hmm okay. That doesn't seem right. But still, please add the unit test, in such a way that it passes.

firestore/integration_test_internal/src/android/env_android_test.cc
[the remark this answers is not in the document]
dconeybe: The test is not correct: `env().Call()` does _not_ throw a c++ exception; rather, it should set a _pending exception_ on the `Env` object. Try this variation of your test code:

```cpp
TEST_F(EnvTest, CallsNullArenaRef) {
  ArenaRef null_arena_ref;
  Local<Class> clazz = env().FindClass("java/lang/String");
  jmethodID to_lower_case = env().GetMethodId(clazz, "toLowerCase", "()Ljava/lang/String;");
  ASSERT_TRUE(env.ok()) // Make sure that there is no pending exception before calling Call()

  env().Call(null_arena_ref, Method<Object>(to_lower_case));
  Local<Throwable> exception = env().ClearExceptionOccurred();
  ASSERT_TRUE(exception) << "A Java NullPointerException should have been thrown by Call()";
  ASSERT_EQ(env().GetObjectClass(exception).GetName(env()), "java.lang.NullPointerException");
}
````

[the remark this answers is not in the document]
Comment (cherylEnkidu) [reply firestore/integration_test_internal/src/android/env_android_test.cc]: The problem with the above code is the test crash at `env().Call` , when the java side crash the test app crash. And I try to catch the crash using cpp `try` and obviously it doesn't work :(

File: firestore/integration_test_internal/src/android/env_android_test.cc (hunk 40 above)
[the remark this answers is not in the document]
Comment (cherylEnkidu) [reply firestore/integration_test_internal/src/android/env_android_test.cc]: I come up with two solutions to solve this:

1. Throw an error when calling `ArenaRef::get()` with `key_ == -1`
2. Inside IsInstanceOf function, returns false when `GetClass()` returns Java base Object class

However, the second solution is better give the following test passes:
```
TEST_F(EnvTest, IsInstanceOfChecksNullArenaRef) {
  Local<Class> clazz = env().FindClass("java/lang/System");
  Local<Object> integer;
  EXPECT_TRUE(env().IsInstanceOf(integer, clazz));
}
```

So the existing `IsInstanceOf` already have such issue. I am not sure whether we want to fix this function in this PR.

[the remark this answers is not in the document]
Comment (dconeybe) [reply firestore/integration_test_internal/src/android/env_android_test.cc]: Ok I see what you're saying. Let's write the test still, but mark it as "skipped". The "skipped" message should say why it is being skipped (i.e. a bug in IsInstanceOf) and that the test should be re-enabled once that but is fixed.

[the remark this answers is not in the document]
Comment (dconeybe) [reply firestore/integration_test_internal/src/android/env_android_test.cc]: b/268420201 created to track this. Please mention "b/268420201" in the skip message.

File: firestore/integration_test_internal/src/android/arena_ref_android_test.cc (hunk 33 above)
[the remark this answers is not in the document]
Comment (cherylEnkidu) [reply firestore/integration_test_internal/src/android/arena_ref_android_test.cc]: Will be add in the later PR (tracking id: b/268421567)

File: firestore/src/jni/arena_ref.h [hunk 41]
```diff
@@ -0,0 +1,69 @@
...
+#include "firestore/src/jni/jni_fwd.h"
+#include "firestore/src/jni/object.h"
+#include "firestore/src/jni/ownership.h"
+
+namespace firebase {
+namespace firestore {
+namespace jni {
+
+namespace {
+constexpr int64_t kInvalidKey = -1;
```
dconeybe, line 33: Please move `kInvalidKey` to inside the `ArenaRef` class. It should probably be a private static variable of `ArenaRef`. Otherwise, it will be part of the `firebase::firestore::jni` namespace, which doesn't really make sense because it only applies to `ArenaRef`.

firestore/integration_test_internal/src/android/env_android_test.cc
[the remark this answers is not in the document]
dconeybe: Outcome: This test is invalid. Both Local<T> and Global<T> will cause a crash if specified to env.Call() when they are null, so it is fine if ArenaRef does too.

firestore/integration_test_internal/src/android/arena_ref_android_test.cc
```diff
@@ -0,0 +1,148 @@
...
+  Local<String> string = env.NewStringUtf("hello world");
+
+  ArenaRef arena_ref1(env, string);
+  ArenaRef arena_ref2(arena_ref1);
+
+  EXPECT_TRUE(arena_ref1.get(env).Equals(env, string));
+  EXPECT_TRUE(arena_ref2.get(env).Equals(env, string));
+}
+
+TEST_F(ArenaRefTestAndroid, CopyAssignsReferenceToNull) {
```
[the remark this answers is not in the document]
cherylEnkidu: I also add case 5. ref copy(move) assigns itself.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,146 @@
...
+namespace firebase {
+namespace firestore {
+namespace jni {
+
+namespace {
+
+HashMap* gArenaRefHashMap = nullptr;
+jclass gLongClass = nullptr;
+jmethodID gLongConstructor = nullptr;
+std::mutex mutex_;
```
dconeybe, line 37: nit: Move `std::mutex mutex_` to be defined *above* the variable that it is protecting (i.e. above the definition of gArenaRefHashMap). Although it's not super relevant in this case, you always want to make sure that the mutex outlives the variable(s) that it is protecting.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,146 @@
...
+}
+
+Local<Long> ArenaRef::key_object(Env& env) const {
+  if (!env.ok()) return {};
+  // The reason why using raw JNI calls here rather than the JNI convenience
+  // layer is because we need to get rid of JNI convenience layer dependencies
+  // in the ArenaRef destructor to avoid racing condition when firestore object
+  // gets destroyed.
+  jobject key = env.get()->NewObject(gLongClass, gLongConstructor, key_);
+  env.RecordException();
```
dconeybe, line 121: Calling `RecordException()` isn't sufficient to guard against exceptions. Instead, check `env.ok()` and return `{}` if it's not ok.

firestore/src/jni/arena_ref.cc
```diff
@@ -0,0 +1,146 @@
...
+  return {env.get(), key};
+}
+
+void ArenaRef::Initialize(Env& env) {
+  if (gArenaRefHashMap) {
+    return;
```
dconeybe, line 127: Acquire the lock at the top of `ArenaRef::Initialize()`

firestore/integration_test_internal/src/jni/arena_ref_test.cc
```diff
@@ -0,0 +1,278 @@
...
+
+namespace firebase {
+namespace firestore {
+namespace jni {
+namespace {
+
+constexpr char kException[] = "java/lang/Exception";
+Method<String> kConstructor("<init>", "(Ljava/lang/String;)V");
+
+class ArenaRefTestAndroid : public FirestoreIntegrationTest {
```
dconeybe, line 35: `ArenaRefTestAndroid` should extend `FirestoreAndroidIntegrationTest`, not `FirestoreIntegrationTest`.

This looks after everything that you are re-implementing in this class:
1. It defines a `Loader` for you to use.
2. It checks for pending exceptions at the end of the test.
3. It provides a method to create an exception.

In fact, I think you should just go back to `using ArenaRefTestAndroid = FirestoreAndroidIntegrationTest;`

firestore/integration_test_internal/src/jni/arena_ref_test.cc
```diff
@@ -0,0 +1,278 @@
...
+    EXPECT_EQ(thrown.GetMessage(env()), "Testing throw");
+  }
+
+ private:
+  // Env is declared as having a `noexcept(false)` destructor, which causes the
+  // compiler to propagate this into `EnvTest`'s destructor, but this conflicts
+  // with the declaration of the parent class. Holding `Env` with a unique
+  // pointer sidesteps this restriction.
+  std::unique_ptr<Env> env_;
+  Loader loader_;
```
dconeybe, line 74: Don't hold an Env in an instance variable. Instead, just do `Env env` when you need one. And methods of this class that need one should take one as a non-const reference.

firestore/integration_test_internal/src/jni/arena_ref_test.cc
```diff
@@ -0,0 +1,278 @@
...
+  }
+
+  Env& env() const { return *env_; }
+
+  void throwException() {
+    Local<Class> clazz = env().FindClass(kException);
+    jmethodID ctor =
+        env().GetMethodId(clazz, "<init>", "(Ljava/lang/String;)V");
+
+    Local<String> message = env().NewStringUtf("Testing throw");
```
dconeybe, line 55: I think you can just get rid of these throwException() and clearExceptionOccurred() helper methods.

throwException() can be replaced by 
```cpp
env.Throw(CreateException(env, "Test Exception"));
```

and clearExceptionOccurred() can be replaced by
```cpp
env.ExceptionClear();
```

Admittedly, this leads to a tiny bit of code duplciation between test cases but that's acceptable because, IMO, it improves readabililty of the test cases. 

See go/tott/598 about "DAMP" tests rather than the "DRY" principle that is common in non-testing code.

firestore/integration_test_internal/src/jni/arena_ref_test.cc
```diff
@@ -0,0 +1,278 @@
...
+TEST_F(ArenaRefTestAndroid, MoveConstructsFromNull) {
+  ArenaRef arena_ref1;
+  ArenaRef arena_ref2(std::move(arena_ref1));
+  EXPECT_EQ(arena_ref2.get(env()).get(), nullptr);
+}
+
+TEST_F(ArenaRefTestAndroid, MoveConstructsFromValid) {
+  Local<String> string = env().NewStringUtf("hello world");
+
+  ArenaRef arena_ref2(env(), string);
```
dconeybe, line 171: nit: Rename `arena_ref2` and `arena_ref3` back to `arena_ref1` and `arena_ref2`, respectively.

firestore/integration_test_internal/src/jni/arena_ref_test.cc
```diff
@@ -0,0 +1,278 @@
...
+  Local<String> string1 = env().NewStringUtf("hello world");
+
+  ArenaRef arena_ref1(env(), string1);
+  arena_ref1 = std::move(arena_ref1);
+  EXPECT_TRUE(env().IsSameObject(arena_ref1.get(env()), string1));
+}
+
+TEST_F(ArenaRefTestAndroid, ThrowBeforeConstruct) {
+  Local<String> string = env().NewStringUtf("hello world");
+  EXPECT_EQ(string.ToString(env()).size(), 11U);
```
dconeybe, line 226: Remove these checks for the length of the string returned from `NewStringUtf()`. You can just assume that this is true. Applies here and elsewhere in this file.

firestore/integration_test_internal/src/jni/arena_ref_test.cc
```diff
@@ -0,0 +1,278 @@
...
+
+TEST_F(ArenaRefTestAndroid, MoveAssignsFromValidObjectItself) {
+  Local<String> string1 = env().NewStringUtf("hello world");
+
+  ArenaRef arena_ref1(env(), string1);
+  arena_ref1 = std::move(arena_ref1);
+  EXPECT_TRUE(env().IsSameObject(arena_ref1.get(env()), string1));
+}
+
+TEST_F(ArenaRefTestAndroid, ThrowBeforeConstruct) {
```
dconeybe, line 224: nit: Can you rename the "ThrowBeforeXXX" test names to "XXXWithPendingException" or something like that? I think using the term "with pending exception" more clearly conveys what scenario the test is testing.

firestore/integration_test_internal/src/jni/arena_ref_test.cc
```diff
@@ -0,0 +1,278 @@
...
+  Local<String> string = env().NewStringUtf("hello world");
+  ArenaRef arena_ref1(env(), string);
+  ArenaRef arena_ref2;
+  EXPECT_EQ(arena_ref1.get(env()).ToString(env()).size(), 11U);
+  throwException();
+  arena_ref2 = std::move(arena_ref1);
+  clearExceptionOccurred();
+  EXPECT_TRUE(env().IsSameObject(arena_ref2.get(env()), string));
+}
+
```
dconeybe, line 274: Add a test for calling get() with a pending exception.

dconeybe (changes requested):

firestore/integration_test_internal/src/android/arena_ref_android_test.cc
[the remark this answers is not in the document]
dconeybe: Good idea for case 5!

PR #1180, opened 2023-01-03
Reduce number of RewardedAd loads on iOS in CI

DellaBitta:
The GMA backend doesn't whitelist iOS devices running in CI which means number of ads that can be served to iOS in CI is restricted. This is true even when using the prescribed [Demo Ad Unit Id](https://developers.google.com/admob/ios/test-ads). 

This PR reduces the number of ads we load on iOS in an attempt to minimize the chance of encountering NoFillErrors and push our CI to green.

Notes:
- Test a subset of RewardedAds behavior on iOS in GHA CI.
  - We no longer test successful RewardedAd loads.
  - Rewarded Ads stress tests are disabled on iOS.
- Added a NoFill error as an acceptable result on stress tests for other platforms. The tests were meant to stress the SDK though ad object creation and destruction. The result from the service is secondary, and NoFill would be a valid result.
   - These tests are not enabled for iOs, though, as the service often goes into backoff mode on iOS. Backoff mode is  represented by an obscure Internal Error code which could mean many different things.
- All of the tests can still be executed locally on a device simply by building the test app. We should manually run the iOS testapp whenever we update the GMA dependencies. 
- Remove Flaky block retries around tests. The "flaky" behavior stemmed from NoFill errors.
  - Retrying on a NoFill error will continue to return NoFill errors since the the pool of ads to serve has been exhausted.
  - Retrying will make the situation worse:
    - The test will still eventually fail, it will just take longer to do so.
    - Repeated NoFill errors can push the service into back-off mode which may take longer to recover from.
  - It's better to wait a few minutes and re-run the test.

***

[Integration Test CI](https://github.com/firebase/firebase-cpp-sdk/actions/runs/3833227384)

***

### Type of Change
***

### DIFF
--- a/gma/integration_test/src/integration_test.cc
+++ b/gma/integration_test/src/integration_test.cc
@@ -845,8 +845,7 @@ TEST_F(FirebaseGmaTest, TestRewardedAdLoad) {
   SKIP_TEST_ON_DESKTOP;
   SKIP_TEST_ON_SIMULATOR;
 
-  // Loading Ads has been deemed flaky as the AdMob Service has a chance to
-  // return NoFill for valid ad requests if there aren't any ads to serve.
+  // TODO(@drsanta): remove when GMA whitelists CI devices.
   FLAKY_TEST_SECTION_BEGIN();
 
   firebase::gma::RewardedAd* rewarded = new firebase::gma::RewardedAd();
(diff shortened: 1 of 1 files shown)


gma/integration_test/src/integration_test.cc
```diff
@@ -778,19 +778,13 @@ TEST_F(FirebaseGmaTest, TestAdViewLoadAd) {
   firebase::Future<firebase::gma::AdResult> load_ad_future;
   const firebase::gma::AdResult* result_ptr = nullptr;
 
-  // Loading Ads has been deemed flaky as the AdMob Service has a chance to
-  // return NoFill for valid ad requests if there aren't any ads to serve.
-  FLAKY_TEST_SECTION_BEGIN();
-
```
DellaBitta, line 784: The flakiness stemmed from NoFill error when the service runs out of Ads to serve. NoFill often takes a lot longer to recover from than our Flaky Test retry timeout, and so the flaky test block causes more problems than it solves.

chkuang-g (approved):

PR #1182, opened 2023-01-06
Upgrade to grpc 150 and C++ 14 for firestore

wu-hui:
Tests against corresponding iOS SDK change: https://github.com/firebase/firebase-ios-sdk/pull/10651

https://github.com/firebase/firebase-cpp-sdk/actions/runs/3970865722

https://github.com/firebase/firebase-cpp-sdk/actions/runs/4598864163


CMakeLists.txt
```diff
@@ -16,7 +16,7 @@
 
 cmake_minimum_required (VERSION 3.1)
 
-set (CMAKE_CXX_STANDARD 11)
+set (CMAKE_CXX_STANDARD 14)
```
dconeybe, line 19: drive-by comment: You should probably set `CMAKE_CXX_STANDARD_REQUIRED` to `true` also.

https://cmake.org/cmake/help/v3.7/variable/CMAKE_CXX_STANDARD_REQUIRED.html

cmake/external/firestore.cmake
```diff
@@ -49,6 +49,7 @@ function(GetTag t)
     DOWNLOAD_DIR ${FIREBASE_DOWNLOAD_DIR}
     GIT_REPOSITORY "https://github.com/firebase/firebase-ios-sdk.git"
     GIT_TAG ${t}
+    GIT_CONFIG "core.autocrlf=input"
```
dconeybe, line 52: Just a heads-up that `GIT_CONFIG` was added in cmake 3.8, and we need to support cmake 3.6, since Android builds are currently pinned to cmake 3.6.

https://cmake.org/cmake/help/latest/module/ExternalProject.html

wu-hui: Thanks..I think this is fine because Android does not get the dependency from git clone..

dconeybe: Oh, but also we document that cmake 3.7 is the minimum supported version. So we shouldn't use cmake syntax that is not available in cmake 3.7.

https://github.com/firebase/firebase-cpp-sdk/blob/873dba22aabb9922d5574e91f626639eb364aaf7/README.md?plain=1#L49

In any case, it's fine for this experiment; but we'd need to do something different before merging anything into main.

PR #1184, opened 2023-01-12
App Check Desktop - GetToken and Listeners support

a-maurice:
Adds support for the GetAppCheckToken functionality, and registering / calling listeners for token changes.  With this, the enabled tests in the integration testapp all pass locally.
***

Running the testapp locally, since passing the debug token hasn't been added to the GHA logic yet.
***

### Type of Change
***

#### Notes

### DIFF
--- a/app_check/integration_test/src/integration_test.cc
+++ b/app_check/integration_test/src/integration_test.cc
@@ -370,12 +370,17 @@ TEST_F(FirebaseAppCheckTest, TestGetTokenForcingRefresh) {
   EXPECT_NE(token.token, "");
   EXPECT_NE(token.expire_time_millis, 0);
 
+  // Wait a bit to make sure the expire time would be different
+  std::this_thread::sleep_for(std::chrono::milliseconds(500));
+
   // GetToken with force_refresh=false will return the same token.
   firebase::Future<::firebase::app_check::AppCheckToken> future2 =
       app_check->GetAppCheckToken(false);
   EXPECT_TRUE(WaitForCompletion(future2, "GetToken #2"));
   EXPECT_EQ(future.result()->expire_time_millis,
             future2.result()->expire_time_millis);
+
+  std::this_thread::sleep_for(std::chrono::milliseconds(500));
 
   // GetToken with force_refresh=true will return a new token.
   firebase::Future<::firebase::app_check::AppCheckToken> future3 =
@@ -452,7 +457,7 @@ TEST_F(FirebaseAppCheckTest, TestSignIn) {
 TEST_F(FirebaseAppCheckTest, TestDebugProviderValidToken) {
   firebase::app_check::DebugAppCheckProviderFactory* factory =
       firebase::app_check::DebugAppCheckProviderFactory::GetInstance();
-#if FIREBASE_PLATFORM_IOS
+#if FIREBASE_PLATFORM_IOS || FIREBASE_PLATFORM_DESKTOP
   ASSERT_NE(factory, nullptr);
   InitializeApp();
   firebase::app_check::AppCheckProvider* provider =
--- a/app_check/src/desktop/app_check_desktop.cc
+++ b/app_check/src/desktop/app_check_desktop.cc
@@ -75,18 +75,20 @@ Future<AppCheckToken> AppCheckInternal::GetAppCheckToken(bool force_refresh) {
     }
     if (provider != nullptr) {
       auto token_callback{
-        [this, handle](firebase::app_check::AppCheckToken token,
-                  int error_code, const std::string& error_message) {
-        if (error_code == firebase::app_check::kAppCheckErrorNone) {
-          UpdateCachedToken(token);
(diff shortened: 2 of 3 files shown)


app_check/integration_test/src/integration_test.cc
```diff
@@ -370,13 +370,18 @@ TEST_F(FirebaseAppCheckTest, TestGetTokenForcingRefresh) {
   EXPECT_NE(token.token, "");
   EXPECT_NE(token.expire_time_millis, 0);
 
+  // Wait a bit to make sure the expire time would be different
```
AlmostMatt, line 373: What's the reason that this sleep is needed?

app_check/integration_test/src/integration_test.cc
```diff
@@ -370,13 +370,18 @@ TEST_F(FirebaseAppCheckTest, TestGetTokenForcingRefresh) {
...
+  std::this_thread::sleep_for(std::chrono::milliseconds(500));
+
   // GetToken with force_refresh=false will return the same token.
   firebase::Future<::firebase::app_check::AppCheckToken> future2 =
       app_check->GetAppCheckToken(false);
   EXPECT_TRUE(WaitForCompletion(future2, "GetToken #2"));
   EXPECT_EQ(future.result()->expire_time_millis,
             future2.result()->expire_time_millis);
 
+  std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
AlmostMatt, line 383: Similar to above, why is this sleep needed?

app_check/src/desktop/app_check_desktop.cc
```diff
@@ -47,8 +67,33 @@ void AppCheckInternal::SetTokenAutoRefreshEnabled(
...
-  AppCheckToken token;
-  future()->CompleteWithResult(handle, 0, token);
+  if (!force_refresh && HasValidCacheToken()) {
+    // If the cached token is valid, and not told to refresh, return the cache
+    future()->CompleteWithResult(handle, 0, cached_token_);
+  } else {
+    // Get a new token, and pass the result into the future.
+    AppCheckProvider* provider = nullptr;
+    if (g_provider_factory && app_) {
+      provider = g_provider_factory->CreateProvider(app_);
```
AlmostMatt, line 77: It seems a little bit wrong to me that we call CreateProvider for every GetToken request.

For custom providers, it is possible that the factory does not cache the provider internally, and so we might be making a new provider every time.

Since we already have a separate instance of AppCheckInternal for each app, and we assume that SetFactory is called before we create an instance of AppCheck, I think it might make sense to have provider as a private variable of AppCheckInternal and call CreateProvider in the constructor of AppCheckInternal.

The only potential downside of that that I see is if we want to allow changing factory after initializing app-check, that would become more complicated.

app_check/src/desktop/app_check_desktop.cc
```diff
@@ -57,9 +102,21 @@ Future<AppCheckToken> AppCheckInternal::GetAppCheckTokenLastResult() {
       future()->LastResult(kAppCheckFnGetAppCheckToken));
 }
 
-void AppCheckInternal::AddAppCheckListener(AppCheckListener* listener) {}
+void AppCheckInternal::AddAppCheckListener(AppCheckListener* listener) {
+  if (listener) {
+    token_listeners_.push_back(listener);
+
+    if (HasValidCacheToken()) {
```
AlmostMatt, line 109: This is probably correct, though iOS does behave differently.

It seems that android does notify immediately if a valid token already exists, so this is consistent with android.
For iOS, token-changed listener is based on notifications being sent and adding a listener will not retroactively receive old notifications.

So this can probably stay as is.

a-maurice: Yeah, I copied this logic from the Android one. The joys of trying to resolve two different implementations. :)

app_check/src/desktop/app_check_desktop.cc
a-maurice: That makes sense, I was definitely waffling a bit on whether to just keep the provider or not.  Will change.

app_check/src/desktop/app_check_desktop.cc
a-maurice: I put a comment calling out the difference.

app_check/integration_test/src/integration_test.cc
AlmostMatt: Oh I see, if the two tokens are generated within a fraction of a second they may have identical expiration times. Seems reasonable.

app_check/src/desktop/app_check_desktop.cc
AlmostMatt: Sounds good.
Please update this line to use the newly created GetProvider method.

AlmostMatt (approved):
Looks good, but please make sure it calls the newly created GetProvider method.

a-maurice:
> Looks good, but please make sure it calls the newly created GetProvider method.

Whoops, good catch

PR #1198, opened 2023-02-01
Add App Check hooks for desktop rest calls

a-maurice:
Add the necessary logic to retrieve a token that can be used by the internal products to add to their rest calls.  Also adds it to storage, along with adding storage to the app check testapp.
***

***

### Type of Change
***

#### Notes


app_check/integration_test/src/integration_test.cc
```diff
@@ -298,6 +311,43 @@ void FirebaseAppCheckTest::InitializeAppAuthDatabase() {
...
+
+  WaitForCompletion(initializer.InitializeLastResult(), "InitializeStorage");
+
+  ASSERT_EQ(initializer.InitializeLastResult().error(), 0)
+      << initializer.InitializeLastResult().error_message();
+
+  LogDebug("Successfully initialized Firebase Storage.");
+}
+
+void FirebaseAppCheckTest::TerminateStorage() {
```
AlmostMatt, line 335: Call this method in TearDown

app_check/integration_test/src/integration_test.cc
```diff
@@ -157,6 +167,8 @@ void FirebaseAppCheckTest::TerminateAppCheck() {
       delete app_check;
     }
   }
+
+  firebase::app_check::AppCheck::SetAppCheckProviderFactory(nullptr);
```
AlmostMatt, line 171: It looks like this might be problematic for android as installAppCheckProviderFactory expects a non-null param
https://github.com/firebase/firebase-android-sdk/blob/master/appcheck/firebase-appcheck/src/main/java/com/google/firebase/appcheck/FirebaseAppCheck.java#L54

But I do think we want some way to reset app check between tests, so I'll worry about that while I am working on android implementation, and this line of code is fine for this PR.

app_check/src/common/common.h
```diff
@@ -19,11 +19,18 @@ namespace firebase {
 namespace app_check {
 namespace internal {
 
+// Used by the function registry, so that other products can call App Check
```
AlmostMatt, line 22: Is this comment accurate? I think this is the enum used by the Future API to identify the future, not the enum used by function registry.

app_check/src/common/common.h
```diff
@@ -19,11 +19,18 @@ namespace firebase {
...
+// functions without introducing a dependency.
 enum AppCheckFn {
   kAppCheckFnGetAppCheckToken = 0,
   kAppCheckFnCount,
 };
 
+// Helper function to get an existing AppCheck instance, since GetInstance will
+// create one if needed.
```
AlmostMatt, line 30: Maybe rephrase to
// Helper function to get an existing AppCheck instance or nullptr
// if an instance does not already exist.

I think that should be sufficient to clarify what the method does differently from getInstance. But this is just a suggestion, feel free to ignore it if the existing comment is more clear to you.

app_check/src/desktop/app_check_desktop.cc
```diff
@@ -126,6 +130,45 @@ void AppCheckInternal::RemoveAppCheckListener(AppCheckListener* listener) {
   }
 }
 
+int g_app_check_registry_count = 0;
```
AlmostMatt, line 133: make this a static int since it is only relevant to this file

app_check/src/desktop/app_check_desktop.cc
```diff
@@ -126,6 +130,45 @@ void AppCheckInternal::RemoveAppCheckListener(AppCheckListener* listener) {
...
+                                                        void* /*unused*/,
+                                                        void* out) {
+  Future<AppCheckToken>* out_future = static_cast<Future<AppCheckToken>*>(out);
+  if (!app || !out_future) {
+    return false;
+  }
+
+  AppCheck* app_check = AppCheck::GetInstance(app);
+  if (app_check) {
+    // TODO(amaurice): This should call some internal function instead of the
```
AlmostMatt, line 163: I assume this will happen in a later change, so not blocking this PR.

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -228,6 +230,7 @@ Future<void> StorageReferenceInternal::DeleteLastResult() {
 }
 
 // Handy utility function, since REST calls have similar setup and teardown.
+// Can potentially block getting Future results for the Request.
```
AlmostMatt, line 233: This change does introduce this potential blocking.
For async storage operations, is PrepareRequest already called on a separate thread? I feel like we just need to be sure that this will not cause async operations to block the main thread.

storage/src/desktop/storage_reference_desktop.cc
```diff
@@ -252,6 +255,19 @@ void StorageReferenceInternal::PrepareRequest(rest::Request* request,
...
                       storage_->user_agent().c_str());
+
+  // Use the function registry to get the App Check token.
+  Future<::firebase::app_check::AppCheckToken> app_check_future;
+  bool succeeded = storage_->app()->function_registry()->CallFunction(
+      ::firebase::internal::FnAppCheckGetTokenAsync, storage_->app(), nullptr,
+      &app_check_future);
+  if (succeeded && app_check_future.status() != kFutureStatusInvalid) {
+    const ::firebase::app_check::AppCheckToken* token =
+        app_check_future.Await(30000);
```
AlmostMatt, line 266: Probably a good idea to make this 30000 a constant with a name that clarifies that this is a number of milliseconds
30s seems like potentially a pretty long time to wait. Do we wait for a similar duration for other operations like getting auth credentials?

AlmostMatt (changes requested):

app/src/util.h
```diff
@@ -186,14 +189,14 @@ class AppCallback {
 //     }
 //   });
 #define FIREBASE_APP_REGISTER_CALLBACKS(module_name, created_code,             \
-                                        destroyed_code)                        \
+                                        destroyed_code, enabled)               \
```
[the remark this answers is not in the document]
a-maurice: Done

app_check/integration_test/src/integration_test.cc
a-maurice: Hmm, yeah.  I think having some way to clear it, especially for tests, would be good.  If it can't be done on Android, we can figure out another plan though.

app_check/src/common/common.h
a-maurice: Whoops, you are right. Fixed

app_check/src/common/common.h
a-maurice: Yeah, makes sense. Fixed

app_check/src/desktop/app_check_desktop.cc
a-maurice: Done

storage/src/desktop/storage_reference_desktop.cc
a-maurice: I renamed it to ...Blocking, to make it a bit more obvious. All the calls are currently done within a BlockingResponse, which already had a blocking call for the rest call, so I believe it should be fine.

storage/src/desktop/storage_reference_desktop.cc
a-maurice: So, Auth tends to not wait for the token, and just sorta assumes it will be there, which I didn't want to do. I did set it to 10 seconds instead, and moved it out along with a comment.

storage/src/desktop/storage_reference_desktop.cc
AlmostMatt: Sounds good. Yea that should be enough to ensure that this code doesn't accidentally block the main thread.

AlmostMatt (approved):

PR #1207, opened 2023-02-13
Firestore COUNT

tom-andersen:


firestore/integration_test_internal/src/query_test.cc
```diff
@@ -675,24 +772,38 @@ TEST_F(QueryTest, TestQueriesCanUseInFilters) {
...
+      ReadAggregate(aggregate_query1);
+  EXPECT_EQ(3, aggregate_snapshot1.count());
+  EXPECT_EQ(aggregate_query1, aggregate_snapshot1.query());
+
   // With objects.
-  snapshot = ReadDocuments(collection.WhereIn(
-      "zip", {FieldValue::Map({{"code", FieldValue::Integer(500)}})}));
+  const Query query2 = collection.WhereIn(
+      "zip", {FieldValue::Map({{"code", FieldValue::Integer(500)}})});
+  const snapshot = ReadDocuments(query2);
```
cherylEnkidu, line 796: `snapshot` needs to have a type

dconeybe: Once you resolve this comment, please add the "tests requested - quick" label to this PR to trigger the full slate of tests.

tom-andersen: Done and triggered.

release_build_files/readme.md
```diff
@@ -644,6 +644,8 @@ code.
 ## Release Notes
 ### Upcoming Release
 -   Changes
+    - Firestore: Added `Query::Count()`, which fetches the number of documents in the result
+      set without actually downloading the documents ([#1174](https://github.com/firebase/firebase-cpp-sdk/pull/1174)).
```
dconeybe, line 648: Update the PR number to this PR, 1207 (in two places in this line)

firestore/src/include/firebase/firestore/query.h
```diff
@@ -143,6 +144,23 @@ class Query {
...
+   *
+   * The returned query, when executed, counts the documents in the result set
+   * of this query without actually downloading the documents.
+   *
+   * Using the returned query to count the documents is efficient because only
+   * the final count, not the documents' data, is downloaded. The returned query
+   * can even count the documents if the result set would be prohibitively large
+   * to download entirely (e.g. thousands of documents).
+   *
+   * @return An aggregate  query that counts the documents in the result set of
```
dconeybe, line 159: nit: reduce 2 spaces to 1 in "aggregate  query"

firestore/src/include/firebase/firestore/aggregate_source.h
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+#ifndef FIREBASE_FIRESTORE_SRC_INCLUDE_FIREBASE_FIRESTORE_AGGREGATE_SOURCE_H_
+#define FIREBASE_FIRESTORE_SRC_INCLUDE_FIREBASE_FIRESTORE_AGGREGATE_SOURCE_H_
+
+namespace firebase {
+namespace firestore {
+
+/**
+ * @brief The sources from which an AggregateQuery::Get can retrieve its
```
dconeybe, line 24: nit "an AggregateQuery::Get" -> "AggregateQuery::Get"

firestore/src/include/firebase/firestore/aggregate_query_snapshot.h
```diff
@@ -0,0 +1,155 @@
...
+   *
+   * @return The number of documents in the result set of the underlying query.
+   */
+  virtual int64_t count() const;
+
+  /**
+   * @brief Returns true if this `AggregateQuerySnapshot` is valid, false if it
+   * is not valid. An invalid `AggregateQuerySnapshot` could be the result of:
+   *   - Creating a `AggregateQuerySnapshot` using the default constructor.
+   *   - Moving from the `AggregateQuery`.
```
dconeybe, line 115: nit: AggregateQuery -> AggregateQuerySnapshot (here and elsewhere in this file)

firestore/src/main/aggregate_query_snapshot_main.cc
```diff
@@ -0,0 +1,55 @@
...
+
+AggregateQuery AggregateQuerySnapshotInternal::query() const {
+  return MakePublic(api::AggregateQuery(aggregate_query_));
+}
+
+int64_t AggregateQuerySnapshotInternal::count() const { return count_result_; }
+
+bool operator==(const AggregateQuerySnapshotInternal& lhs,
+                const AggregateQuerySnapshotInternal& rhs) {
+  // TODO(tomandersen) - there needs to be equals operator defined on
```
dconeybe, line 48: Please create a Buganizer ticket (e.g. a task in taskflow) to track this work, and replace your name here with "b/NNNNNNN".

firestore/src/main/aggregate_query_main.cc
```diff
@@ -0,0 +1,80 @@
...
+        } else {
+          promise.SetError(maybe_value.status());
+        }
+      });
+  return promise.future();
+}
+
+bool operator==(const AggregateQueryInternal& lhs,
+                const AggregateQueryInternal& rhs) {
+  // TODO(tomandersen) - there needs to be equals operator defined on
```
dconeybe, line 74: Please create a Buganizer ticket (e.g. a task in taskflow) to track this work, and replace your name here with "b/NNNNNNN".

firestore/src/android/promise_android.h
```diff
@@ -22,6 +22,7 @@
 
 #include "app/src/reference_counted_future_impl.h"
 #include "app/src/util_android.h"
+#include "firestore/src/android/aggregate_query_snapshot_android.h"
```
dconeybe, line 25: Can this `#include` be removed?

firestore/src/android/load_bundle_task_progress_android.cc
```diff
@@ -25,7 +25,6 @@ namespace firestore {
 namespace {
 
 using jni::Class;
-using jni::Constructor;
```
dconeybe, line 28: Is this change related to count queries? If not, please revert it.

firestore/src/android/aggregate_query_snapshot_android.cc
```diff
@@ -0,0 +1,79 @@
...
+
+}  // namespace
+
+void AggregateQuerySnapshotInternal::Initialize(jni::Loader& loader) {
+  loader.LoadClass(kClassName, kConstructor, kGetCount, kGetQuery, kHashCode);
+}
+
+AggregateQuerySnapshot AggregateQuerySnapshotInternal::Create(
+    Env& env, AggregateQueryInternal& aggregate_query_internal, int64_t count) {
+  const Object& arg = aggregate_query_internal.ToJava();
```
dconeybe, line 51: nit: rename `arg` to a more meaningful name.

firestore/src/android/aggregate_query_android.h
```diff
@@ -0,0 +1,78 @@
...
+    kCount,  // Must be the last enum value.
+  };
+
+  static void Initialize(jni::Loader& loader);
+
+  AggregateQueryInternal(FirestoreInternal* firestore,
+                         const jni::Object& object)
+      : Wrapper(firestore, object), promises_(firestore) {}
+
+  /**
```
dconeybe, line 45: Optional: Remove these javadoc-style comments. IMO they just add noise and these methods are documented elsewhere.

firestore/integration_test_internal/src/query_test.cc
```diff
@@ -236,11 +236,18 @@ TEST_F(QueryTest, TestKeyOrderIsDescendingForDescendingInequality) {
                   {"e", {{"foo", FieldValue::Double(21.0)}}},
                   {"f", {{"foo", FieldValue::Integer(66)}}},
                   {"g", {{"foo", FieldValue::Double(66.0)}}}});
-  QuerySnapshot snapshot = ReadDocuments(
+  const Query query =
```
dconeybe, line 239: @ehsannas Here is the testing code we'd like your opinion on. This PR updates each query test to _also_ run a count query.

My thought was that this isn't the right approach because (a) this reduces the focus of these tests on query features, (b) adds coverage for various query features that should be tested on the underlying SDKs, and (c) does not scale to additional aggregations like SUM and AVERAGE.

@tom-andersen Argues that (a) this is great test coverage to have, (b) does not need to be replicated for SUM and AVERAGE, and (c) the integration tests are already not very focussed so the possible reduction in the tests' "focus" is worth it. (please correct me if I've mis-represented your position, Tom).

tom-andersen: For purpose of discussion, I think there are 3 different issues at play here:

1. Should the wrapper SDK do a thorough integration test, or simply verify it correctly wraps underlying SDK and defer to the underlying SDK to do a thorough test? A lot of tests (not just for count) could be simplified or removed if we defer to underlying SDK. 

2. Currently the C++ SDK has **_integration_** tests, in the sense that it requires making real calls to emulator/backend to function, thereby exercising many layers of code simultaneously. I think what we really want is **_unit_** tests that can test units of code in isolation by mocking out dependencies. I think this is fundamental to our disagreement, in the sense that I think Denver is applying advice normally given to unit tests that can have a narrow focus, when in reality these are integration tests that exercise the full stack.

3. Contrast between Unit Testing and Integration Testing:
[Unit Testing](https://engdoc.corp.google.com/eng/doc/devguide/testing/unit/overview.md?polyglot=cpp&cl=head):
- Focused
- Understandable
- Fast
- Deterministic
- Resilient
- Hermetic
[Integration Testing](https://engdoc.corp.google.com/eng/doc/devguide/testing/integration/what-is-integration.md?cl=head):
- Not Focused and/or Not Hermetic

Here are some of my thoughts:
1. Integration tests tell a story of multiple interactions (create data, query, then delete, makes sure is deleted, etc). When considering query operators, a part of that story is what documents are returned and does the count match.
2. If an integration test fails, there are many layers that could be at fault (lack of focus). A good integration test will give hints to isolate problem. If the count is wrong, the next natural question will be if the returned documents are wrong as well. If both are wrong, maybe there is an error in how we construct request to backend (since they share logic for serializing filter operators).
3. Our test suite is slow (mainly because it is written as integration tests). Breaking tests up into smaller tests will mean more time is spent on test setup/teardown of populating documents and other tasks. This adds up, and we should course correct. Some of this might be mitigated by tests sharing setup/teardown environment, but that could also lead to interfering tests.

firestore/src/android/aggregate_query_android.h
tom-andersen: This is the pattern from other internal classes. I agree, that for internal classes, this is just noise.

I am tempted to remove it from all of them, but maybe that should be a backlog item. If the other files are cleaned up, then the pattern will stop replicating.

firestore/src/android/load_bundle_task_progress_android.cc
dconeybe: Nope, there is no backlog item for this. IMO it doesn't really matter and including it in this PR adds noise. I'm sure there are unused includes and using statements scattered everywhere in our code base!

firestore/src/android/aggregate_query_android.h
dconeybe: Yeah it's weird that we've duplicated the docs internally. If removing these docs breaks consistency with surrounding code then I'd say to go with the local convention.

firestore/src/android/load_bundle_task_progress_android.cc
tom-andersen: For small things like this, we should probably take the [Boy Scout Rule](https://github.com/97-things/97-things-every-programmer-should-know/blob/master/en/thing_08/README.md) approach. Especially, if we can't justify tracking in backlog.

dconeybe: I tend to disagree with the "Boy Scout Rule" in this context. IMO PRs should be laser focused on what their purpose is. As unrelated changes bleed into a PR it makes it more difficult to review and understand after the fact. Also, if the PR needed to be reverted then those "unrelated" changes would also get reverted. You can always open a separate PR for unrelated cleanups like this and I would happily review them expeditiously. After all, I'm a huge fan of cleaning up the code base!

tom-andersen: I can accept excluding the change since it is in a separate file, that is otherwise untouched.

I am however concerned that we are creating unnecessary barriers that will promote tech debt. I am not about to make a separate PR to just remove a couple of imports since there is higher priority work. Which is why this barrier to be laser focused is counter productive when applied to trivial changes IMO. Changes are likely never done, and will likely create a separate cognitive burden to readers of the code.

I don't find there is a cognitive burden to pruning imports in PRs since the CI would fail if it was required. But I suppose with a mindset of being laser focused, then any deviation creates questions and cognitive load.

I do find unused imports create a burden to the reader of code. I was looking for an example of how to use `jni::Constructor` in our code base, and this import was a red herring (among others) that made my work with the code slower.

firestore/src/android/promise_android.h
tom-andersen: No.

dconeybe:
@tom-andersen Note that the "Checks / check_integration_test_labels (pull_request)" label will change from "fail" to "pass" automatically once all of the integration tests pass. It is there merely to prevent merging until the integration tests pass.

PR #1217, opened 2023-02-22
Fix user-agent register timing in C++ and Unity SDK

chkuang-g:
Patch user-agent register timing in C++ and add utility functions for Unity SDK to register BEFORE C++ App is created.

Several notes:
- The original use of `util::GetJNIEnvFromApp()` is not reliable because it only works AFTER the first app is added.
- The new `CallAfterEnsureMethodsCached()` functions is primarily for Unity SDK to register BEFORE C++ app is created, which is when the Java classes and methods are cached originally.

***

Integration test. Manually E2E test for Android and iOS Auth testapp.

***

### Type of Change
***

#### Notes


app/src/app_common.cc
```diff
@@ -433,6 +418,27 @@ void RegisterLibrariesFromUserAgent(const char* user_agent) {
   if (changed) registry->UpdateUserAgent();
 }
 
+void RegisterSdkUsage(void* platform_resource) {
+  MutexLock lock(*g_app_mutex);
+
+  // Only register libraries when no C++ apps was created before.
+  if (g_apps == nullptr) {
```
AlmostMatt, line 425: This might be a nitpick, but even if g_apps currently correlates to whether or not an app has been created previously, I think it might be a bit safer if you added an IsInitialized() method to LibraryRegistry and check that.

app/src/util_android.h
```diff
@@ -1127,6 +1127,12 @@ jint AttachCurrentThread(JavaVM* java_vm, JNIEnv** env);
 // App. If there is no instantiated App, returns nullptr.
 JNIEnv* GetJNIEnvFromApp();
 
+// Make sure the Java classes and methods are cached before triggering the
+// the callback. Can be slow if this is called BEFORE any Firebase App is
+// created.
+void CallAfterEnsureMethodsCached(JNIEnv* env, jobject activity,
```
AlmostMatt, line 1133: Maybe add a comment mentioning that the Unity SDK depends on this method so that it is not accidentally deleted.

app/src/include/firebase/app.h
```diff
@@ -697,7 +697,9 @@ class App {
   /// @param library Name of the library to register as a user of the Firebase
   /// C++ SDK.
   /// @param version Version of the library being registered.
-  static void RegisterLibrary(const char* library, const char* version);
+  /// @param context Platform specific resource. Ex. for Android, this is JNIEnv
```
AlmostMatt, line 700: @param platform_resource to match the variable name

app/src/util_android.h
AlmostMatt, line 1133: It's a little bit strange to me that this is implemented in app_android.cc and the header is util_android.h.

I think the implementation makes sense to be in app_android.cc as it calls CacheMethods and I guess there is not currently an app_android.h file, so I suppose this is fine as is.

AlmostMatt (approved):
This generally looks good to me. I just added a few minor comments.

app/src/app_common.cc
a-maurice: Agreed, this should probably check something else, like how on the Unity side there is a bool tracking registration.  If you do want to check if g_apps is null, either an assert or logging a warning/error seems like the better approach.

app/src/util_android.h
chkuang-g: Done

chkuang-g: It is.

I'm not sure if it is worth adding a `app_android.h` just for this single function. And unfortunately the implementation has to live in `app_android.cpp` so that it can access to all the JNI caching and releasing functions.

app/src/include/firebase/app.h
chkuang-g: Done

app/src/app_common.cc
chkuang-g: Done.
I'm still using `g_app_mutex` to guard from race condition. Using the same mutex also has less chance for deadlock.

a-maurice (approved):

a-maurice (approved):

app/src/app_common.cc
```diff
@@ -190,7 +190,7 @@ struct AppData {
 // Tracks library registrations.
 class LibraryRegistry {
  private:
-  LibraryRegistry() {}
+  LibraryRegistry() : is_common_libraries_registered(false) {}
```
jonsimantov, line 193: nit: what about naming this, and the accessors, "Library" instead of "Libraries", as IsCommonLibraryRegistered is more correct?

app/src/app_ios.mm
```diff
@@ -277,7 +281,7 @@ static void PlatformOptionsToAppOptions(FIROptions* platform_options, AppOptions
 
 App* App::GetInstance(const char* name) { return app_common::FindAppByName(name); }
 
-void App::RegisterLibrary(const char* library, const char* version) {
+void App::RegisterLibrary(const char* library, const char* version, void* platform_resource) {
```
jonsimantov, line 284: If it's not used on this platform, do the /* platform_resource */ thing.

app/src/util_android.h
jonsimantov: Why isn't the implementation in util_android, if it's not a method of App or related types like AppOptions?

jonsimantov (approved):
This seems mostly fine except for a few nits.

app/src/util_android.h
chkuang-g: `CallAfterEnsureMethodsCached()` is an utility by nature but requires `CacheMethods()` and `ReleaseClasses()` which is only available in `app_android.cpp`. 

https://github.com/firebase/firebase-cpp-sdk/blob/main/app/src/app_android.cc

I can think of some alternative solutions
1. Add `app_android.h` and declare `CallAfterEnsureMethodsCached()` in `app_android.h`
1. Add `app_android.h` and declare `CacheMethods()` and `ReleaseClasses()` so that they can be linked to `util_android.cpp`. Then move `CallAfterEnsureMethodsCached()` to `util_android.cpp`.
1. Move declaration and impl of `CacheMethods()` and `ReleaseClasses()` to `util_android`.

Do you see a risk to implement a function declare in `util_android.h` but implemented in `app_android.cpp`?

chkuang-g: Added `app_android.h`

app/src/app_common.cc
chkuang-g: Done

app/src/app_ios.mm
chkuang-g: Done. Also changed for desktop and stub

app/src/app_android.h
```diff
@@ -0,0 +1,32 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef FIREBASE_APP_SRC_APP_ANDROID_H_
+#define FIREBASE_APP_SRC_APP_ANDROID_H_
+
+#include <functional>
+
+#include <jni.h>
```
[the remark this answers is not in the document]
chkuang-g: Done

chkuang-g (commented):
Change applied.

jonsimantov (approved):

PR #1220, opened 2023-02-23
Android AppCheck main implementation

AlmostMatt:
Integrates c++ appcheck with android appcheck methods and classes. Supports installation of AppCheck providers.

Also Includes:
- appcheck_resources folder that includes Java classes wrapping C++ providers and tokens in order to be used with the Android SDK
- gradle and build files related to appcheck_resources
- a helper method in android_app to lookup C++ App given an instance of Java App.

***

Ran integration tests locally.
***

### Type of Change
***

#### Notes


app_check/CMakeLists.txt
```diff
@@ -107,6 +116,7 @@ set(desktop_SRCS
   src/stub/play_integrity_provider_stub.cc
 )
 
+
```
a-maurice, line 119: nit: unnecessary extra new line?

app_check/integration_test/src/integration_test.cc
```diff
@@ -611,6 +605,13 @@ TEST_F(FirebaseAppCheckTest, TestPlayIntegrityProvider) {
 #if FIREBASE_PLATFORM_ANDROID
   ASSERT_NE(factory, nullptr);
   InitializeApp();
+  // TODO(almostmatt): Once app initialization automatically initializes
+  // AppCheck, this test will no longer need to explicitly get an instance of
+  // AppCheck.
+  ::firebase::app_check::AppCheck* app_check =
```
a-maurice, line 611: Is this still necessary?

app_check/src_java/com/google/firebase/appcheck/internal/cpp/JniAppCheckProvider.java
```diff
@@ -0,0 +1,56 @@
...
+*/
+public class JniAppCheckProvider implements AppCheckProvider {
+  private long cProvider;
+
+  public JniAppCheckProvider(long cProvider) {
+    this.cProvider = cProvider;
+  }
+  @Override
+  public Task<AppCheckToken> getToken() {
+    TaskCompletionSource<AppCheckToken> taskCompletionSource = new TaskCompletionSource<>();
```
a-maurice, line 35: Missing the AppCheckToken from the constructor

app/src/app_android.cc
```diff
@@ -569,4 +572,24 @@ JavaVM* App::java_vm() const { return internal_->java_vm(); }
 
 jobject App::GetPlatformApp() const { return internal_->GetLocalRef(); }
 
+namespace internal {
+
+// Default android app name.
+const char* const kDefaultAndroidAppName = "[DEFAULT]";
+
+// Get the corresponding App from an Android SDK FirebaseApp instance.
+App* GetAppFromPlatformApp(JNIEnv* jni_env, jobject platform_app) {
```
a-maurice, line 581: Looking into it a bit, the only other way I can think of doing this would be to check against all the C++ Apps that are saved, and checking against the Java object that those should be saving.  I think going by name as you have done is fine, both will have some problems.

I do think you should add checks for null (asserts or something), just to help potentially flag any issues that might arise, though I can't think of any.

a-maurice (commented):
In general looks good, though the comments in the Java files could use a pass.

app_check/integration_test/src/integration_test.cc
AlmostMatt: No, this should be ok to remove now.

app_check/src_java/com/google/firebase/appcheck/internal/cpp/JniAppCheckProvider.java
AlmostMatt: I think in Java it is able to infer the <AppCheckToken>, but I added it to be more clear.

app/src/app_android.cc
AlmostMatt: That might be better if there are multiple apps with the same name. I'll see if I can try such an approach.

AlmostMatt:
> In general looks good, though the comments in the Java files could use a pass.

I've updated the comments in the java files.

app/src/app_android.cc
AlmostMatt: I briefly looked into what the code would look like to iterate all app instances and compare the corresponding platform_app.

The simplest way to iterate all apps would probably be for app_common to have a method that returns a list of apps. 
The java app is stored as AppInternal internal_, which is a private variable of App, so app_common doesn't have permission to compare internal_ to a jobject. So likely App would need to add a method to get or compare the internal_ to another object, and a per-platform implementation of this method.

So it can be done, but since it depends on AppInternal, it would be a bit complicated. I think I prefer lookup by name for now unless multiple apps with the same name is an important use-case.

a-maurice: I'm fine with the going by name for now, but we should take a note to confirm that names are unique.

app_check/src_java/com/google/firebase/appcheck/internal/cpp/JniAppCheckProvider.java
```diff
@@ -0,0 +1,61 @@
...
+      taskCompletionSource.setException(new Exception(error_message));
+    }
+  }
+
+  /**
+   * This function is implemented in the AppCheck C++ library (app_check_android.cc).
+   */
+  private native void nativeGetToken(
+      long cProvider, TaskCompletionSource<AppCheckToken> task_completion_source);
+}
```
a-maurice, line 61: Can you add newlines to the end of these java files?

a-maurice (approved):

app_check/src_java/com/google/firebase/appcheck/internal/cpp/JniAppCheckProvider.java
AlmostMatt: Done.

app/src/app_android.cc
AlmostMatt: I've just updated this PR to not rely on app lookup by name.
The new approach is to specify an App* at the time that JniAppCheckProviderFactory is created. This works because the jni factory's constructer is called in the AppCheckInternal initialization logic which has an associated App and because the JDK installs AppCheck factories for a single instance of AppCheck instead of installing it for all instances.

a-maurice (approved):

PR #1228, opened 2023-03-02
Add App Check support for Firestore

a-maurice:
Implements an App Check Credential Provider for Firestore to use for the desktop implementation, and hook into the iOS provider for the iOS implementation.
Adds a few tests to the App Check integration test to use Firestore.
***

Run the desktop tests locally.
***

### Type of Change
***

#### Notes


app_check/integration_test/src/integration_test.cc
```diff
@@ -821,4 +919,191 @@ TEST_F(FirebaseAppCheckTest, TestFunctionsFailure) {
...
+
+  WaitForCompletion(
+      document.Set(firebase::firestore::MapFieldValue{
+          {"str", firebase::firestore::FieldValue::String("foo")},
+          {"int", firebase::firestore::FieldValue::Integer(123)}}),
+      "document.Set");
+  firebase::Future<firebase::firestore::DocumentSnapshot> future =
+      document.Get(firebase::firestore::Source::kServer);
+  WaitForCompletion(future, "document.Get");
+  EXPECT_NE(future.result(), nullptr);
```
dconeybe, line 937: nit: `EXPECT_NE` -> `ASSERT_NE` to avoid a segfault on the next line in the case that `future.result()` erroneously returned `nullptr`.

app_check/integration_test/src/integration_test.cc
```diff
@@ -821,4 +919,191 @@ TEST_F(FirebaseAppCheckTest, TestFunctionsFailure) {
...
+          {"val", firebase::firestore::FieldValue::String("start")}}),
+      "document.Set 0");
+
+  std::vector<firebase::firestore::MapFieldValue> document_snapshots;
+  firebase::firestore::ListenerRegistration registration =
+      document.AddSnapshotListener(
+          [&](const firebase::firestore::DocumentSnapshot& result,
+              firebase::firestore::Error error_code,
+              const std::string& error_message) {
+            EXPECT_EQ(error_code, firebase::firestore::kErrorOk);
```
dconeybe, line 982: Consider adding a `SCOPED_TRACE()` here with a message that includes the size of the vector. That way, if one of the expectations fails it is clear which invocation(s) suffered from the failure.

app_check/integration_test/src/integration_test.cc
```diff
@@ -821,4 +919,191 @@ TEST_F(FirebaseAppCheckTest, TestFunctionsFailure) {
...
+  InitializeFirestore();
+
+  firebase::firestore::DocumentReference document = CreateFirestoreDoc();
+
+  WaitForCompletion(
+      document.Set(firebase::firestore::MapFieldValue{
+          {"val", firebase::firestore::FieldValue::String("start")}}),
+      "document.Set 0");
+
+  std::vector<firebase::firestore::MapFieldValue> document_snapshots;
```
dconeybe, line 976: Optional: Technically, there is a race condition accessing `document_snapshots` since it is written to by one thread and read from another. If this test ever flakes, consider protecting the vector with a mutex. Also, if this test case were to uncover bugs it could also crash, such as if a callback to the snapshot listener occurred after the vector was destructed.

Although it adds some noise to the test case, I've fixed this in other tests by creating a struct that contains both the mutex and the vector, putting that struct into a shared_ptr, and capturing that shared_ptr by value in the lambda. Doing this guarantees that the vector and the mutex stay alive as long as they are referenced.

e.g.

```cpp
struct ListenerSnapshots {
  std::mutex mutex;
  std::vector<firebase::firestore::MapFieldValue> snapshots;
};
auto listener_snapshots = std::make_shared<ListenerSnapshots>();
```

app_check/integration_test/src/integration_test.cc
```diff
@@ -821,4 +919,191 @@ TEST_F(FirebaseAppCheckTest, TestFunctionsFailure) {
...
+          {"int", firebase::firestore::FieldValue::Integer(456)}}),
+      "document.Set", firebase::firestore::kErrorPermissionDenied);
+  WaitForCompletion(document.Get(firebase::firestore::Source::kServer),
+                    "document.Get",
+                    firebase::firestore::kErrorPermissionDenied);
+
+  CleanupFirestore(firebase::firestore::kErrorPermissionDenied);
+}
+
+TEST_F(FirebaseAppCheckTest, TestFirestoreTransaction) {
```
dconeybe, line 964: Rename `TestFirestoreTransaction`. The word "transaction" has a very specific meaning in the context of Firestore (i.e. a `Transaction` object) but this test has nothing to do with that kind of "transaction".

Applies here and elsewhere in this file.

https://firebase.google.com/docs/reference/cpp/class/firebase/firestore/transaction

app_check/integration_test/src/integration_test.cc
```diff
@@ -821,4 +919,191 @@ TEST_F(FirebaseAppCheckTest, TestFunctionsFailure) {
...
+  bool received_pending_cache = false;
+  bool received_permission_denied = false;
+  firebase::firestore::ListenerRegistration registration =
+      document.AddSnapshotListener(
+          [&received_pending_cache, &received_permission_denied](
+              const firebase::firestore::DocumentSnapshot& result,
+              firebase::firestore::Error error_code,
+              const std::string& error_message) {
+            // We first expect one call for a successful change to the cache.
+            if (error_code == firebase::firestore::kErrorNone) {
```
dconeybe, line 1022: Asserting a very specific sequence of snapshots is brittle. Instead, I'd recommend just asserting that _eventually_ you get a "permission denied" error.

app_check/integration_test/src/integration_test.cc
```diff
@@ -821,4 +919,191 @@ TEST_F(FirebaseAppCheckTest, TestFunctionsFailure) {
...
+      "document.Set transaction", firebase::firestore::kErrorPermissionDenied);
+
+  registration.Remove();
+
+  WaitForCompletion(
+      document.Set(firebase::firestore::MapFieldValue{
+          {"val", firebase::firestore::FieldValue::String("final")}}),
+      "document.Set final", firebase::firestore::kErrorPermissionDenied);
+
+  EXPECT_TRUE(received_pending_cache);
```
dconeybe, line 1049: I think this test will be flaky. I can't see that there is any guarantee that the permission error would have occurred by this point, especially in the case of a slow network. Instead, I'd recommend to use a condition variable and notify it once the permission denied error is received by the snapshot listener, timing out after some reasonable amount of time (like 30 seconds).

app_check/integration_test/src/integration_test.cc
```diff
@@ -821,4 +919,191 @@ TEST_F(FirebaseAppCheckTest, TestFunctionsFailure) {
...
+                {"int", firebase::firestore::FieldValue::Integer(123)}});
+        return firebase::firestore::kErrorOk;
+      });
+
+  WaitForCompletion(transaction_future, "firestore.RunTransaction");
+
+  // Confirm the updated doc is correct.
+  auto future = document.Get(firebase::firestore::Source::kServer);
+  WaitForCompletion(future, "document.Get");
+  EXPECT_THAT(future.result()->GetData(),
```
dconeybe, line 1082: Insert `ASSERT_NE(future.result(), nullptr)` above this line to avoid a crash if `future.result()` were ever null.

dconeybe (changes requested):
I've reviewed the firestore appcheck tests and the changes in the "firestore" directory. For the latter, I didn't look in too much detail, especially the code that looks to have been copied and adapted from the existing credentials provider. Is there anything I should look at in more detail?

app_check/integration_test/src/integration_test.cc
```diff
@@ -390,6 +408,86 @@ void FirebaseAppCheckTest::TerminateFunctions() {
...
+
+  ProcessEvents(100);
+}
+
+firebase::firestore::CollectionReference
+FirebaseAppCheckTest::GetTestCollection() {
+  if (collection_name_.empty()) {
+    // Generate a collection for the test data based on the time in
+    // milliseconds.
+    int64_t time_in_microseconds =
```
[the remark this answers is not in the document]
a-maurice: I believe this logic does that for the documents, this is just to get a CollectionReference.  I copied this bit from the other Firestore testapps, so I feel like we should keep them the same just for consistency, and if this is a problem we can update all of them together.

app_check/integration_test/src/integration_test.cc
a-maurice: Ah, good catch

app_check/integration_test/src/integration_test.cc
a-maurice: Done

app_check/integration_test/src/integration_test.cc
a-maurice: Added it, just to be safe.  This test is something I copied and simplified from the other Firestore testapps.  I don't think the flakes we are seeing in those are these functions, but if they are we might want to consider updating them to be a bit safer with this logic.

app_check/integration_test/src/integration_test.cc
a-maurice: Whoops, changed to Listener which makes more sense

app_check/integration_test/src/integration_test.cc
a-maurice: Makes sense.  Changed it to just check that if it receives an Ok, the metadata is what is expected, but it no longer requires that an Ok happens.

app_check/integration_test/src/integration_test.cc
a-maurice: Done

app_check/integration_test/src/integration_test.cc
a-maurice: Done

a-maurice:
> I've reviewed the firestore appcheck tests and the changes in the "firestore" directory. For the latter, I didn't look in too much detail, especially the code that looks to have been copied and adapted from the existing credentials provider. Is there anything I should look at in more detail?

Yeah, it is mostly just copied from the AuthCredentialProvider logic, and just hooks into App Check instead.  As long as that seems fine with you, that's good enough.

app_check/integration_test/src/integration_test.cc
```diff
@@ -967,15 +973,23 @@ TEST_F(FirebaseAppCheckTest, TestFirestoreTransaction) {
...
-  std::vector<firebase::firestore::MapFieldValue> document_snapshots;
+  struct ListenerSnapshots {
+    std::mutex mutex;
+    std::vector<firebase::firestore::MapFieldValue> snapshots;
+  };
+  auto listener_snapshots = std::make_shared<ListenerSnapshots>();
   firebase::firestore::ListenerRegistration registration =
       document.AddSnapshotListener(
-          [&](const firebase::firestore::DocumentSnapshot& result,
+          [&listener_snapshots](
```
dconeybe, line 983: nit: Capture `listener_snapshots` by value, instead of by reference. Otherwise, you forego the benefits of shared_ptr.

dconeybe (changes requested):

app_check/integration_test/src/integration_test.cc
a-maurice: Doh, good point

dconeybe (commented):
LGTM, for the parts that I reviewed. I'll withhold my approval unless you feel like my partial review is sufficient. But I think someone more familiar with the non-firestore parts of this PR should take a look too.

dconeybe (dismissed):

firestore/src/main/app_check_credentials_provider_desktop.h
```diff
@@ -0,0 +1,98 @@
...
+  // credentials provider holds a shared pointer to `Contents`, while the
+  // `GetToken` callback stores a weak pointer. This makes safe the case where
+  // the `GetToken` callback might be invoked after this credentials provider
+  // has already been destroyed (Auth may outlive Firestore).
+  struct Contents {
+    explicit Contents(App& app) : app(app) {}
+
+    // FirebaseCppCredentialsProvider may be used by more than one thread. The
+    // mutex is locked in all public member functions and none of the private
+    // member functions (with the exception of `RequestToken` that locks the
```
AlmostMatt, line 79: This bit of the comment can be removed. RequestToken is an auth method and the similar logic is just done inline in the public GetToken method now.

That said, in Auth RequestToken obtains the lock during the callback where it calls the listener with the token. We may want to obtain the lock similarly in the OnCompletion callback in GetToken. If so this comment can be kept but rephrased slightly to mention the GetToken lambda instead of RequestToken lambda.

AlmostMatt (commented):
Looks ok, just some minor comments

app_check/src/desktop/app_check_desktop.cc
```diff
@@ -14,6 +14,7 @@
 
 #include "app_check/src/desktop/app_check_desktop.h"
 
+#include <algorithm>
```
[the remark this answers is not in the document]
a-maurice: std::find

app_check/integration_test/src/integration_test.cc
```diff
@@ -390,6 +410,86 @@ void FirebaseAppCheckTest::TerminateFunctions() {
...
+
+    delete firestore_;
+    firestore_ = nullptr;
+  }
+
+  ProcessEvents(100);
+}
+
+firebase::firestore::CollectionReference
+FirebaseAppCheckTest::GetTestCollection() {
```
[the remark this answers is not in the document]
a-maurice: Done

firestore/src/main/firestore_main.h
```diff
@@ -150,11 +150,13 @@ class FirestoreInternal {
 
   FirestoreInternal(
       App* app,
-      std::unique_ptr<credentials::AuthCredentialsProvider> credentials);
```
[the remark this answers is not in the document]
a-maurice: Done

firestore/src/main/create_app_check_credentials_provider_ios.mm
```diff
@@ -0,0 +1,45 @@
...
+namespace firebase {
+namespace firestore {
+
+using credentials::AppCheckCredentialsProvider;
+using credentials::FirebaseAppCheckCredentialsProvider;
+
+std::unique_ptr<AppCheckCredentialsProvider> CreateAppCheckCredentialsProvider(
+    App& app) {
+  FIRApp* ios_app = app.GetPlatformApp();
+  auto ios_auth = FIR_COMPONENT(FIRAppCheckInterop, ios_app.container);
```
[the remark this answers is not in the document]
a-maurice: Done

firestore/src/main/app_check_credentials_provider_desktop.h
a-maurice: After going over it again, I don't believe it is needed. Nothing in the listener of GetToken requires anything from the object, I simplified to just have the app directly, instead of the additional wrapping around it.

AlmostMatt: Removing the 'contents' object makes sense. I see that you also removed the mutex entirely. I think that is reasonable, as the only variable that I see that might be modified is change_listener_, and it's likely just set up-front and then left alone, compared to auth which has the token_generation field which is frequently modified. So no mutex is probably fine here.

AlmostMatt (approved):

PR #1253, opened 2023-04-02
Feature/io phone auth provider

DellaBitta:
Add PhoneAuthCredential.

***

***

### Type of Change
***

### DIFF
--- a/auth/integration_test/src/integration_test.cc
+++ b/auth/integration_test/src/integration_test.cc
@@ -1031,6 +1031,13 @@ class PhoneListener : public firebase::auth::PhoneAuthProvider::Listener {
         on_code_sent_count_(0),
         on_code_auto_retrieval_time_out_count_(0) {}
 
+  void OnVerificationCompleted(
+      firebase::auth::PhoneAuthCredential phone_auth_credential) override {
+    LogDebug("PhoneListener: successful automatic verification.");
+    on_verification_complete_count_++;
+    credential_ = phone_auth_credential;
+  }
+
   void OnVerificationCompleted(firebase::auth::Credential credential) override {
     LogDebug("PhoneListener: successful automatic verification.");
     on_verification_complete_count_++;
@@ -1154,7 +1161,7 @@ TEST_F(FirebaseAuthTest, TestPhoneAuth) {
     EXPECT_GT(listener.on_code_auto_retrieval_time_out_count(), 0);
     EXPECT_NE(listener.verification_id(), "");
     LogDebug("Signing in with verification code.");
-    const firebase::auth::Credential phone_credential =
+    const firebase::auth::PhoneAuthCredential phone_credential =
         phone_provider.GetCredential(listener.verification_id().c_str(),
                                      kPhoneAuthTestVerificationCode);
 
--- a/auth/integration_test/src/integration_test.cc
+++ b/auth/integration_test/src/integration_test.cc
@@ -1026,21 +1026,24 @@ TEST_F(FirebaseAuthTest, TestAuthPersistenceWithEmailSignin) {
 class PhoneListener : public firebase::auth::PhoneAuthProvider::Listener {
  public:
   PhoneListener()
-      : on_verification_complete_count_(0),
+      : on_verification_complete_credential_count_(0),
+        on_verification_complete_phone_auth_credential_count_(0),
         on_verification_failed_count_(0),
         on_code_sent_count_(0),
         on_code_auto_retrieval_time_out_count_(0) {}
 
   void OnVerificationCompleted(
       firebase::auth::PhoneAuthCredential phone_auth_credential) override {
-    LogDebug("PhoneListener: successful automatic verification.");
-    on_verification_complete_count_++;
-    credential_ = phone_auth_credential;
+    LogDebug(
+        "PhoneListener: PhoneAuthCredential successful automatic "
+        "verification.");
+    on_verification_complete_phone_auth_credential_count_++;
+    phone_auth_credential_ = phone_auth_credential;
   }
 
   void OnVerificationCompleted(firebase::auth::Credential credential) override {
-    LogDebug("PhoneListener: successful automatic verification.");
-    on_verification_complete_count_++;
+    LogDebug("PhoneListener: Credential successful automatic verification.");
+    on_verification_complete_credential_count_++;
     credential_ = credential;
   }
 
@@ -1072,7 +1075,14 @@ class PhoneListener : public firebase::auth::PhoneAuthProvider::Listener {
     return force_resending_token_;
   }
   int on_verification_complete_count() const {
-    return on_verification_complete_count_;
+    return on_verification_complete_phone_auth_credential_count() +
+           on_verification_complete_credential_count();
+  }
+  int on_verification_complete_phone_auth_credential_count() const {
+    return on_verification_complete_phone_auth_credential_count_;
+  }
+  int on_verification_complete_credential_count() const {
+    return on_verification_complete_credential_count_;
   }
   int on_verification_failed_count() const {
     return on_verification_failed_count_;
@@ -1084,7 +1094,8 @@ class PhoneListener : public firebase::auth::PhoneAuthProvider::Listener {
 
   // Helper functions for workflow.
   bool waiting_to_send_code() {
-    return on_verification_complete_count() == 0 &&
+    return on_verification_complete_credential_count() == 0 &&
+           on_verification_complete_phone_auth_credential_count() == 0 &&
            on_verification_failed_count() == 0 && on_code_sent_count() == 0;
   }
 
@@ -1095,12 +1106,17 @@ class PhoneListener : public firebase::auth::PhoneAuthProvider::Listener {
   }
 
   firebase::auth::Credential credential() { return credential_; }
+  firebase::auth::PhoneAuthCredential phone_auth_credential() {
+    return phone_auth_credential_;
+  }
 
  private:
   std::string verification_id_;
   firebase::auth::PhoneAuthProvider::ForceResendingToken force_resending_token_;
   firebase::auth::Credential credential_;
-  int on_verification_complete_count_;
+  firebase::auth::PhoneAuthCredential phone_auth_credential_;
+  int on_verification_complete_phone_auth_credential_count_;
+  int on_verification_complete_credential_count_;
   int on_verification_failed_count_;
   int on_code_sent_count_;
   int on_code_auto_retrieval_time_out_count_;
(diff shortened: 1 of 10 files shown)


auth/src/android/user_android.cc
```diff
@@ -26,16 +26,6 @@ using util::CharsToJniUri;
...
-// clang-format off
-#define PHONE_CREDENTIAL_METHODS(X)                                            \
-    X(GetSmsCode, "getSmsCode", "()Ljava/lang/String;")
-// clang-format on
-METHOD_LOOKUP_DECLARATION(phonecredential, PHONE_CREDENTIAL_METHODS)
-METHOD_LOOKUP_DEFINITION(phonecredential,
-                         PROGUARD_KEEP_CLASS
-                         "com/google/firebase/auth/PhoneAuthCredential",
-                         PHONE_CREDENTIAL_METHODS)
-
```
DellaBitta, line 38: Moved this to the credential source file.

auth/src/common.cc
```diff
@@ -21,9 +21,6 @@
 namespace firebase {
 namespace auth {
 
-// static member variables
-const uint32_t PhoneAuthProvider::kMaxTimeoutMs = 3000;
-
```
DellaBitta, line 26: This was already deprecated and is being removed for I/O.

auth/src/android/credential_android.cc
```diff
@@ -358,13 +366,21 @@ Credential::~Credential() {
...
-  *this = rhs;
+  if (rhs.impl_ != nullptr) {
+    JNIEnv* env = GetJniEnv();
+    jobject j_cred_ref = env->NewGlobalRef(static_cast<jobject>(rhs.impl_));
+    impl_ = static_cast<void*>(j_cred_ref);
+  } else {
+    impl_ = nullptr;
+  }
+  error_code_ = rhs.error_code_;
+  error_message_ = rhs.error_message_;
```
chkuang-g, line 377: This is pretty much the same implementation to `operator=()`, isn't it?
Why not just use the original `*this = rhs;`

auth/src/android/credential_android.cc
```diff
@@ -441,6 +457,58 @@ Credential EmailAuthProvider::GetCredential(const char* email,
...
+}
```
chkuang-g, line 480: Similarly, this is pretty much the same implementation to `operator=()`, isn't it?
Why not just use `*this = rhs;`?

auth/src/include/firebase/auth/credential.h
```diff
@@ -122,6 +122,31 @@ class Credential {
   /// @endcond
 };
 
+/// Wraps phone number and verification information for authentication purposes.
```
chkuang-g, line 125: Since this is a public API, use Doxygen style comment on class and methods?

auth/src/ios/credential_ios.mm
```diff
@@ -94,6 +94,45 @@ @implementation PhoneListenerDataObjC
...
+    impl_ = new FIRPhoneAuthCredentialPointer(PhoneAuthCredentialFromImpl(rhs.impl_));
+  }
+  error_code_ = rhs.error_code_;
+  error_message_ = rhs.error_message_;
+  sms_code_ = rhs.sms_code_;
+}
+
+PhoneAuthCredential& PhoneAuthCredential::operator=(const PhoneAuthCredential& rhs) {
+  if (rhs.impl_ != NULL) {
+    delete static_cast<FIRPhoneAuthCredentialPointer*>(impl_);
```
chkuang-g, line 118: null check for `impl_`

auth/integration_test/src/integration_test.cc
```diff
@@ -1142,10 +1165,16 @@ TEST_F(FirebaseAuthTest, TestPhoneAuth) {
     ProcessEvents(kWaitIntervalMs);
     wait_ms += kWaitIntervalMs;
   }
+
+  LogDebug("phone_auth_credential sms code: %s",
+           listener.phone_auth_credential().sms_code().c_str());
+
   if (listener.on_verification_complete_count() > 0) {
+    EXPECT_EQ(listener.on_verification_complete_count(), 2);
```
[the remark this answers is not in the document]
DellaBitta: Updated.

auth/src/android/credential_android.cc
DellaBitta: Fixed!

auth/src/android/common_android.h
```diff
@@ -43,6 +43,12 @@ METHOD_LOOKUP_DECLARATION(authresult, AUTH_RESULT_METHODS)
 // clang-format on
 METHOD_LOOKUP_DECLARATION(additional_user_info, ADDITIONAL_USER_INFO_METHODS)
 
+// clang-format off
+#define PHONE_CREDENTIAL_METHODS(X)                                            \
+    X(GetSmsCode, "getSmsCode", "()Ljava/lang/String;")
+// clang-format on
+METHOD_LOOKUP_DECLARATION(phonecredential, PHONE_CREDENTIAL_METHODS)
+
```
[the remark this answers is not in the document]
DellaBitta: Unfortunately no, as user_android.cc needs to use the class definition as well.

auth/src/android/credential_android.cc
```diff
@@ -441,6 +449,50 @@ Credential EmailAuthProvider::GetCredential(const char* email,
   return cred;
 }
 
+//
+// PhoneAuthCredential methods
+//
+PhoneAuthCredential::PhoneAuthCredential() {}
```
chkuang-g, line 455: Set `impl_` to `nullptr` and `error_code_` to `kAuthErrorNone;`.

The initial value for a pointer is not necessary to always be `nullptr` :(

Similarly in `credential_ios.mm`

auth/src/android/credential_android.cc
```diff
@@ -441,6 +449,50 @@ Credential EmailAuthProvider::GetCredential(const char* email,
...
+  error_code_ = kAuthErrorNone;
+}
+
+PhoneAuthCredential::PhoneAuthCredential(const PhoneAuthCredential& rhs) {
+  *this = rhs;
+}
+
+PhoneAuthCredential& PhoneAuthCredential::operator=(
+    const PhoneAuthCredential& rhs) {
+  if (rhs.impl_ != nullptr) {
```
chkuang-g, line 468: I think you should remove the global reference to its own `impl_` first, if not nullptr, before assigning it to any other things, even assigning it to null.

To reduce unnecessary steps to add and remove references, you probably can check if two `PhoneAuthCredential` are the same before doing the rest of the work,.

auth/src/ios/credential_ios.mm
```diff
@@ -94,6 +94,48 @@ @implementation PhoneListenerDataObjC
...
+}
+
+PhoneAuthCredential& PhoneAuthCredential::operator=(const PhoneAuthCredential& rhs) {
+  if (rhs.impl_ != NULL) {
+    if (impl_ != nullptr) {
+      delete static_cast<FIRPhoneAuthCredentialPointer*>(impl_);
+      impl_ = nullptr;
+    }
+    impl_ = new FIRPhoneAuthCredentialPointer(PhoneAuthCredentialFromImpl(rhs.impl_));
+  }
```
chkuang-g, line 123: What if `rhs.impl_` is `NULL`? I think you should delete `impl_`, if not NULL, and change it to `NULL` as well.

auth/src/android/credential_android.cc
chkuang-g: Ah, this has been done in `Credential` already. nvm

auth/src/android/credential_android.cc
chkuang-g: Huh....`Credential` is doing this as well. This looks......pretty wrong to me
https://github.com/firebase/firebase-cpp-sdk/blob/main/auth/src/android/credential_android.cc#L365

auth/src/android/credential_android.cc
chkuang-g: I think you need to call parent constructor like 
```
PhoneAuthCredential::PhoneAuthCredential() : Credential() {}
```

auth/src/android/credential_android.cc
```diff
@@ -441,6 +449,50 @@ Credential EmailAuthProvider::GetCredential(const char* email,
...
 
+//
+// PhoneAuthCredential methods
+//
+PhoneAuthCredential::PhoneAuthCredential() {}
+
+PhoneAuthCredential::PhoneAuthCredential(void* impl) {
+  impl_ = impl;
+  error_code_ = kAuthErrorNone;
+}
```
chkuang-g, line 460: This whole blocked can be reduced to 
```
PhoneAuthCredential::PhoneAuthCredential(void* impl) : Credential(impl) {}
```

auth/src/android/credential_android.cc
chkuang-g: I think this whole block can be reduced to 
```
PhoneAuthCredential& PhoneAuthCredential::operator=(
    const PhoneAuthCredential& rhs) {
  Credential::operator=(rhs);
  sms_code_ = rhs.sms_code_;
}
```

Also, please change `Credential::operator=()` to dereference before assigning `impl_`

auth/src/include/firebase/auth/credential.h
```diff
@@ -122,6 +122,31 @@ class Credential {
...
+
+  /// Copy constructor.
+  PhoneAuthCredential(const PhoneAuthCredential& rhs);
+
+  /// Copy a Credential.
+  PhoneAuthCredential& operator=(const PhoneAuthCredential& rhs);
+
+  /// Gets the automatically retrieved SMS verification code if applicable.
+  /// This method is only supported on Android.
+  std::string sms_code() const;
```
chkuang-g, line 139: After looking into this a bit deeper, I think you only need to implement `std::string sms_code() const;` per platform. All the other implementation are pretty much the same if you leverage the parent constructor and copy operator.

Perhaps introduce a `credential_common.cc` and put all the constructors and copy operator there?

auth/src/ios/credential_ios.mm
chkuang-g: I see this is the same implementation in `Credential`. 
https://github.com/firebase/firebase-cpp-sdk/blob/main/auth/src/ios/credential_ios.mm#L72

I think we should fix this in `Credential` and all the child constructor and operator just leverage the implementation in `Credential`.

auth/src/android/credential_android.cc
DellaBitta: The compiler call's the default constructor of a parent class automatically.

auth/src/android/credential_android.cc
DellaBitta: Great catch, thanks!

auth/src/ios/credential_ios.mm
DellaBitta: Fixed the `rhs_.impl == nullptr` problem, but I can't reuse the `Credential::operator=` implementation since `Credential` stores a `FIRAuthCredentialPointer` in the `void* impl`instead of the `FIRPhoneAuthCredentialPointer` here.

DellaBitta: I can reuse the implementation of `Credential::operator=` on the Android-side, though.

auth/src/include/firebase/auth/credential.h
DellaBitta: Sure! I added a `credential.cc` file with the other common files `auth.cc` and `user.cc`. I put the constructors in there but the assignment operator is still platform dependent.

auth/src/android/credential_android.cc
chkuang-g: Ah, you are right. 
Seems like `Parent()` will always be called by default from ANY variation of child constructor. That is, `Child(const Child&)` would call `Parent()` by default if not specified.

auth/src/ios/credential_ios.mm
chkuang-g: I see. Thanks for the update.

Could you also fix `Credential::operator=()` and make sure it deletes `impl_` as well very similar to what you are doing here. 
I think the current `Credential::operator=()` is not correct.

chkuang-g (commented):
One last bit: please fix `Credential::operator=()` for iOS

auth/src/ios/credential_ios.mm
DellaBitta: Done.

chkuang-g (approved):

PR #1258, opened 2023-04-03
Adding in public headers and methods for Realtime-RC

qdpham13:
Adding in public header and methods for Realtime-RC: bug/274929760.
Things added:
- ConfigUpdateListenerRegistration header and class
- ConfigUpdate struct
- RemoteConfigError enum
- AddOnConfigUpdateListener method for top level RemoteConfig
- AddOnConfigUpdateListener stub for RemoteConfigInternal in the header
- -Empty tests for AddOnConfigUpdateListener


remote_config/src/common.h
```diff
@@ -15,6 +15,9 @@
 #ifndef FIREBASE_REMOTE_CONFIG_SRC_COMMON_H_
 #define FIREBASE_REMOTE_CONFIG_SRC_COMMON_H_
 
+#include <string>
```
AlmostMatt, line 18: I think these lines can be reverted.

remote_config/src/config_update_listener_registration.h
```diff
@@ -0,0 +1,48 @@
...
+
+namespace firebase {
+namespace remote_config {
+
+//  Calling Remove stops the listener from receiving
+//  config updates and unregisters itself. If remove is called and no other
+//  listener registrations remain, the connection to the Remote Config backend
+//  is closed. Subsequently calling AddOnConfigUpdate will re-open the
+//  connection.
+class ConfigUpdateListenerRegistration {
```
AlmostMatt, line 28: Move this file to remote_config/src/include/firebase/remote_config/
Since this is a new public header, also add it to remote_config_HDRS in /app/CMakeLists.txt 
(I'm not sure exactly how it gets used, but we enumerate all public headers there)

This class is part of the public RemoteConfig API, so this header file should be under src/include/firebase. The pattern that we follow is that there is one main public header and then a subfolder that contains all other public headers and the main header includes the other public headers.

The corresponding source file can stay where you have it.

remote_config/src/config_update_listener_registration.cc
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#include "remote_config/src/config_update_listener_registration.h"
+
+namespace firebase {
+namespace remote_config {
+
+ConfigUpdateListenerRegistration::ConfigUpdateListenerRegistration(
+    std::function<void()> listener_removal_function)
```
AlmostMatt, line 21: Having registrations store a function might work, though when we get to implementation we might find a different pattern more convenient.

For example, we might have per-platform implementations of this class or of an internal version of this class in case this class needs to do any platform-specific logic such as memory management in the constructor or destructor. Or we might have this class store a reference to RemoteConfigInternal and interact with that.

Since it is a stub for now, I think just leave it as a no-parameter constructor for now.

remote_config/src/config_update_listener_registration.cc
AlmostMatt, line 1: Since this is a new source file, add it to common_SRCS in /remote_config/CMakeLists.txt
If we introduce any platform-specific source files they are listed separately in the CMake file so that the compiler knows which files to use when building.

remote_config/tests/remote_config_test.cc
```diff
@@ -703,5 +703,7 @@ TEST_F(RemoteConfigTest, GetInfo) {
   EXPECT_EQ(info.last_fetch_status, kLastFetchStatusFailure);
   EXPECT_EQ(info.last_fetch_failure_reason, kFetchFailureReasonThrottled);
 }
+
```
AlmostMatt, line 706: Probably just revert the change to this file.
I you look at the CMake file in this folder, this test appears to be disabled. When we moved to GitHub we don't have the same infra so a number of unit tests like this are no longer run and we rely more on the integration tests.

AlmostMatt (changes requested):

remote_config/src/common.h
qdpham13: Reverted

remote_config/src/config_update_listener_registration.h
qdpham13: Updated CMake file and moved to subfolder in include/firebase

remote_config/src/config_update_listener_registration.cc
qdpham13: changed to no-param constructor

remote_config/tests/remote_config_test.cc
qdpham13: reverted

AlmostMatt (approved):

PR #1264, opened 2023-04-09
Add new Auth and User methods to replace the deprecated ones

DellaBitta:
Adds the methods to replace those that were deprecated in the Auth and User classes.
Duplicated every Integration Test so that there's one test for the deprecated API and one test for the new API.
**Note:** `User::LinkAndRetreiveveDataWithCredential` has been removed as it has the same functionality as the new `User::LinkWithCredential`.

The major changes of this PR are in four different areas:

1. The addition of tests to `integration_test.cc`.
2. The addition of new methods in the User and Auth implementation.
3. The plaform specific functions that close Futures on callbacks. Functions that faciliate `Future<SignInResult>` and `Future<User*>` completions, marshalling native data into those two return types have been two new compatriot versions for `Future<AuthResult>` and `Future<User>` return types.
4. The internal experimental implementation of FederatedAuth for desktop calls into the user's application to facilitate a auth. The class `FederatedAuthProvider` has been expanded with callbacks that support `AuthResult` objects in addition to the deprecated `SignInResult` objects.

Fixes b/138788092.

***

[Integration Tests](https://github.com/firebase/firebase-cpp-sdk/actions/runs/4678920642)
Locally tested on an iPhone, Android phone (including Federated Auth tests) and macOS desktop.

***

### Type of Change
***


auth/src/ios/credential_ios.mm
```diff
@@ -625,14 +718,13 @@ void ReauthenticateWithProviderGetCredentialCallback(FIRAuthCredential* _Nullabl
...
     ios_provider.scopes = util::StringVectorToNSMutableArray(provider_data_.scopes);
-    // TODO(b/138788092) invoke FIRUser:reuthenticateWithProvider instead, once that method is added
-    // to the iOS SDK.
-    [ios_provider getCredentialWithUIDelegate:nullptr
-                                   completion:^(FIRAuthCredential* _Nullable credential,
-                                                NSError* _Nullable error) {
-                                     ReauthenticateWithProviderGetCredentialCallback(
-                                         credential, error, handle, auth_data, ios_provider);
-                                   }];
+    [UserImpl(auth_data) reauthenticateWithProvider:ios_provider
```
jonsimantov, line 721: Woo.

auth/integration_test/src/integration_test.cc
```diff
@@ -101,6 +101,10 @@ class FirebaseAuthTest : public FirebaseTest {
   // Delete the current user if it's currently signed in.
   void DeleteUser();
```
jonsimantov, line 102: I guess technically this is supposed to be a utility function to facilitate tests, not an actual test of the API surface. We should have a proper TestDeleteUser for that, and a TestDeleteUser_DEPRECATED.

However, thinking about it, let's leave this, since deprecated users should *only* be using deprecated APIs, and thus DeleteUser_DEPRECATED should be used by the deprecated tests. Might be worth a TODO to add a TestDeleteUser in the future.

jonsimantov (approved):

auth/src/common.cc
```diff
@@ -75,5 +77,10 @@ void PhoneAuthProvider::Listener::OnCodeSent(
 void PhoneAuthProvider::Listener::OnCodeAutoRetrievalTimeOut(
     const std::string& /*verification_id*/) {}
```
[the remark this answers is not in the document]
DellaBitta: I have, but it keeps reporting it as a problem for some reason.

PR #1266, opened 2023-04-11
Add in real-time RC iOS implementation

qdpham13:
Create implementation for real-time RC iOS, includes:
- Create lambda wrapper to be used by implementation stubs
- ConfigUpdateListenerRegistration implementation wrapper for iOS
- Conversions for RemoteConfigError and ConfigUpdate
- Linking C++ listener to Obj-C listener
- Linking Obj-C registration `remove` to C++ registration


remote_config/src/common.h
```diff
@@ -63,6 +66,25 @@ class FutureData {
   static FutureData* s_future_data_;
 };
 
+template <typename TConfigUpdate, typename TRemoteConfigError>
```
AlmostMatt, line 69: What is this LambdaConfigUpdateListener class trying to accomplish compared to calling the callback directly? It should be possible to pass around a std::function variable if needed.

AlmostMatt, line 69: Can the types ConfigUpdate and RemoteConfigError be used directly instead of creating a template? It looks to me like the types do not vary.

remote_config/src/ios/remote_config_ios.mm
```diff
@@ -173,6 +173,31 @@ static void GetInfoFromFIRRemoteConfig(FIRRemoteConfig *rc_instance, ConfigInfo
...
+  }
+}
+
+static ConfigUpdate ConvertConfigUpdateKeys(NSSet<NSString *> *keys) {
+  ConfigUpdate configUpdate;
+
+  for (NSString *key in keys) {
+    configUpdate.updated_keys.push_back(util::NSStringToString(key).c_str());
+  }
+
```
AlmostMatt, line 197: nitpick: the blank line before and after the for loop could be removed.

remote_config/src/ios/remote_config_ios.mm
```diff
@@ -491,9 +516,24 @@ static Variant FIRValueToVariant(FIRRemoteConfigValue *in_value, ValueInfo *valu
 }
 
 ConfigUpdateListenerRegistration* RemoteConfigInternal::AddOnConfigUpdateListener(
-      std::function<void(ConfigUpdate&&, RemoteConfigError)>
-          config_update_listener) {
-            return nullptr;
+      LambdaConfigUpdateListener<ConfigUpdate, RemoteConfigError> *config_update_listener) {
```
AlmostMatt, line 519: It might be necessary to use __block so that the callback function has a reference to config_update_listener. I'm still a bit fuzzy on how __block works in objective C though, so I'm not entirely sure if it is relevant here.

Once you update the integration test to actually call this method and the callback function it will be a bit more clear if there is any issue with the current logic.

remote_config/src/ios/remote_config_ios.mm
```diff
@@ -491,9 +516,24 @@ static Variant FIRValueToVariant(FIRRemoteConfigValue *in_value, ValueInfo *valu
...
+          if (error) {
+            config_update_listener->onError(ConvertFIRRemoteConfigUpdateError(error));
+          }
+          if (update) {
+            config_update_listener->onUpdate(ConvertConfigUpdateKeys(update.updatedKeys));
+          }
+    }];
+
+    ConfigUpdateListenerRegistration *registrationWrapper =
+      new ConfigUpdateListenerRegistration([&registration]() {
```
AlmostMatt, line 532: If registration is a pointer, it may be fine to capture by value instead of reference (removing the &).
Since registration is a local variable I worry that capture by reference will result in the variable falling out of scope before this callback is evaluated.

This is probably another case of test it and see if it works as-is.

remote_config/CMakeLists.txt
```diff
@@ -45,15 +45,15 @@ build_flatbuffers("${flatbuffer_schemas}"
...
-    src/config_update_listener_registration.cc)
+    src/remote_config.cc)
 
 # Source files used by the Android implementation.
 set(android_SRCS
     src/android/remote_config_android.cc)
 
 # Source files used by the iOS implementation.
 set(ios_SRCS
+    src/ios/config_update_listener_registration.cc
```
cynthiajoan, line 56: Is this file also going to be used by android? If so it should stay under the common source

remote_config/src/common.h
qdpham13: Just providing a wrapper for the std::function. When I tried to call the std::function in the iOS implementation and I passed in ConfigUpdate and RemoteConfigError values, I got an error when I tried to build the project. The error I got were along the lines of `candidate function not viable: no known conversion from 'long' to 'firebase::remote_config::ConfigUpdate' for 1st argument`

qdpham13: I got an error when I tried to build the project. The error I got were along the lines of candidate function not viable: no known conversion from 'long' to 'firebase::remote_config::ConfigUpdate' for 1st argument

remote_config/integration_test/src/integration_test.cc
```diff
@@ -408,12 +408,14 @@ TEST_F(FirebaseRemoteConfigTest, TestFetchSecondsParameter) {
 
 TEST_F(FirebaseRemoteConfigTest, TestAddOnConfigUpdateListener) {
```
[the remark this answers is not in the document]
qdpham13: Added a test for real-time that checks the updated values and activates after fetch.

remote_config/integration_test/src/integration_test.cc
```diff
@@ -408,12 +408,14 @@ TEST_F(FirebaseRemoteConfigTest, TestFetchSecondsParameter) {
...
   EXPECT_EQ(nullptr, rc_->AddOnConfigUpdateListener(
-                         [](ConfigUpdate&&, RemoteConfigError) {}));
+                         [](firebase::remote_config::ConfigUpdate&&,
+                            firebase::remote_config::RemoteConfigError) {}));
 }
 
 TEST_F(FirebaseRemoteConfigTest, TestRemoveConfigUpdateListener) {
```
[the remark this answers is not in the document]
qdpham13: Updated to actually make the call

remote_config/src/ios/remote_config_ios.mm
qdpham13: Updated

remote_config/src/ios/remote_config_ios.mm
qdpham13: From the integration test it looks like it works fine

remote_config/src/ios/remote_config_ios.mm
qdpham13: Added listener removal test and it looks to work fine

remote_config/CMakeLists.txt
qdpham13: This implementation was iOS specific. I'm not sure how Android will implement the registration.

remote_config/src/common.h
AlmostMatt: I assume that you are referring to LambdaEventListener, but I think that class is used to listen to multiple types of events (QuerySnapshot, DocumentSnapshot, or LoadBundleTaskProgress), so a templated type makes sense.

Here we expect TConfigUpdate to always be ConfigUpdate, so it should be possible to just use the types ConfigUpdate and RemoteConfigError directly.

The error about converting from long to ConfigUpdate is strange. I suspect that there may have been some other issue causing that, like a pointer being provided instead of the object itself, but it is hard to say without seeing the code. Do you recall the lines of code causing that error?

Anyways, I think this class should not be a templated type. It is probably fine to wrap std::function in a class like this, at least for now.

AlmostMatt: Replied in the other comment about the same error.

remote_config/src/common.h
```diff
@@ -63,6 +66,25 @@ class FutureData {
...
+class LambdaConfigUpdateListener {
+ public:
+  LambdaConfigUpdateListener(
+      std::function<void(TConfigUpdate, TRemoteConfigError)> callback)
```
AlmostMatt, line 73: when referencing the function, I think this is ConfigUpdate&& to match the function's type in the public API. Same change for line 85. I'm unsure what this means for onUpdate, as I haven't used the && address reference feature much.

remote_config/src/ios/remote_config_ios.mm
AlmostMatt: Sounds good, this can stay as-is then

remote_config/CMakeLists.txt
AlmostMatt: I'll either make an android version of this file or move it back to common if it generalizes well enough.

remote_config/integration_test/src/integration_test.cc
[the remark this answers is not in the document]
AlmostMatt: The test looks good, nice :)

remote_config/src/ios/remote_config_ios.mm
AlmostMatt: Alright, sounds good

remote_config/src/common.h
qdpham13: I got this error for both parameters RemoteConfigError and ConfigUpdate when I tried to invoke it in the iOS implementation. The code that caused this was: 
`
registration = [impl() addOnConfigUpdateListener: ^(FIRRemoteConfigUpdate *_Nullable update,
                                  NSError *_Nullable error) {
          if (error) {
            config_update_listener(nullptr, ConvertFIRRemoteConfigUpdateError(error));
          }
          if (update) {
            config_update_listener(ConvertConfigUpdateKeys(update.updatedKeys), nullptr);
          }
    }];


`

remote_config/src/common.h
qdpham13: I think the compiler is able to infer the type of lvalue from the std::function since the integration test runs and is successful. Also adding the double ampersand ends up breaking the build (see the attached photo) 
image

remote_config/src/common.h
qdpham13: After looking back at it I think you're right and I didn't need the wrapper. Removed it and use the function directly

remote_config/src/common.h
qdpham13: Removed the listener wrapper

remote_config/src/common.h
AlmostMatt: Ok, the error there would be that ConfigUpdate is a struct which is not a nullable type. It looks like you figured that out though with the later code. Calling the function directly looks good.

remote_config/integration_test/src/integration_test.cc
```diff
@@ -188,6 +193,58 @@ static Future<void> SetZeroIntervalConfigSettings(RemoteConfig* rc) {
 
 // Test cases below.
 
+TEST_F(FirebaseRemoteConfigTest, TestAddOnConfigUpdateListener) {
```
AlmostMatt, line 196: As mentioned in chat, you observed that this test currently needs to run before other test cases, as for realtime RC the last-fetched version is cached on disk. Add a comment above this test describing this requirement. And we will see whether or not this is an issue for the automated GitHub runners.

qdpham13: Right, just added that

PR #1270, opened 2023-04-13
Pass database_id from GetInstance() down to CreateFirestore()

milaGGL:
Passing database_id as an additional parameter all the way down to iOS `CreateFirestore` function.
Passing the database_id parameter only, no implementation change yet.

### DIFF
--- a/firestore/integration_test_internal/src/util/integration_test_util.cc
+++ b/firestore/integration_test_internal/src/util/integration_test_util.cc
@@ -37,7 +37,8 @@ struct TestFriend {
   static FirestoreInternal* CreateTestFirestoreInternal(App* app) {
 #if !defined(__ANDROID__)
     return new FirestoreInternal(
-        app, /*database_id=*/"(default)",absl::make_unique<credentials::EmptyAuthCredentialsProvider>(),
+        app, /*database_id=*/"(default)",
+        absl::make_unique<credentials::EmptyAuthCredentialsProvider>(),
         absl::make_unique<credentials::EmptyAppCheckCredentialsProvider>());
 #else
     return new FirestoreInternal(app);
--- a/firestore/integration_test_internal/src/util/integration_test_util.cc
+++ b/firestore/integration_test_internal/src/util/integration_test_util.cc
@@ -37,7 +37,7 @@ struct TestFriend {
   static FirestoreInternal* CreateTestFirestoreInternal(App* app) {
 #if !defined(__ANDROID__)
     return new FirestoreInternal(
-        app, absl::make_unique<credentials::EmptyAuthCredentialsProvider>(),
+        app, /*database_id=*/"(default)",absl::make_unique<credentials::EmptyAuthCredentialsProvider>(),
         absl::make_unique<credentials::EmptyAppCheckCredentialsProvider>());
 #else
     return new FirestoreInternal(app);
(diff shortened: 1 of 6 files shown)


firestore/integration_test_internal/src/util/integration_test_util.cc
```diff
@@ -37,7 +37,7 @@ struct TestFriend {
   static FirestoreInternal* CreateTestFirestoreInternal(App* app) {
 #if !defined(__ANDROID__)
     return new FirestoreInternal(
-        app, absl::make_unique<credentials::EmptyAuthCredentialsProvider>(),
+        app, /*database_id=*/"(default)",absl::make_unique<credentials::EmptyAuthCredentialsProvider>(),
```
milaGGL, line 40: fill in `(default)` for test firestore for now, will adjust to accept databaseId later

firestore/integration_test_internal/src/util/integration_test_util.cc
```diff
@@ -37,7 +37,8 @@ struct TestFriend {
   static FirestoreInternal* CreateTestFirestoreInternal(App* app) {
 #if !defined(__ANDROID__)
     return new FirestoreInternal(
-        app, absl::make_unique<credentials::EmptyAuthCredentialsProvider>(),
+        app, /*database_id=*/"(default)",
```
milaGGL, line 40: Fill in `(default)` in test firestore, will adjust to take dynamic database_id later while implementing integration tests

firestore/src/include/firebase/firestore.h
```diff
@@ -492,7 +495,8 @@ class Firestore {
   friend class csharp::ApiHeaders;
   friend class csharp::TransactionManager;
 
-  explicit Firestore(::firebase::App* app);
+  explicit Firestore(::firebase::App* app,
```
milaGGL, line 498: should I use overload functions or give default value for the second parameter?

firestore/src/main/firestore_main.h
```diff
@@ -30,6 +30,7 @@
 #include "app/src/future_manager.h"
 #include "app/src/include/firebase/app.h"
 #include "firestore/src/common/event_listener.h"
+#include "firestore/src/include/firebase/firestore.h"
```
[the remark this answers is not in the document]
tom-andersen: I don't think you need to set a default in this class. The layer above should always set a value.

firestore/src/main/firestore_main.h
```diff
@@ -55,7 +56,8 @@ class Executor;
 class FirestoreInternal {
  public:
   // Note: call `set_firestore_public` immediately after construction.
-  explicit FirestoreInternal(App* app);
+  explicit FirestoreInternal(App* app,
+                             const char* database_id = kDefaultDatabase);
```
tom-andersen, line 60: We should strive to keep logic in one place. What I see is that the default value for `database_id` is being spread into many different files.

You could set default here, but then I think you should remove it from the Firestore class. But, since is both a `main` and `android` internal classes, then the logic is still duplicated. The least amount of duplicated code can be achieved if you set default in Firestore class, and then make it a required parameter on all layers below.

tom-andersen (changes requested):

tom-andersen (approved):
LGTM

PR #1280, opened 2023-04-20
Android Implementation of Realtime RC

AlmostMatt:
### Description
Implementation for real-time RC for Android, includes:

- Implements AddOnConfigUpdateListener
- Reuses the ios ConfigUpdateListenerRegistration implementation. Provides a removal function that uses JNI to call the native registration's remove method. 
- Conversion for RemoteConfigError and ConfigUpdate
- Defines a java ConfigUpdateListener which forwards updates to a C++ listener
***
Ran integration test locally, verified that listener callback is called
(Testing in progress)
***

### Type of Change
***

#### Notes


remote_config/remote_config_resources/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+  implementation platform('com.google.firebase:firebase-bom:31.5.0')
+  implementation 'com.google.firebase:firebase-config'
+}
+
+afterEvaluate {
+  generateReleaseBuildConfig.enabled = false
+}
+
+apply from: "$rootDir/android_build_files/extract_and_dex.gradle"
+extractAndDexAarFile('remote_config_resources')
```
cynthiajoan, line 61: extra line behind

remote_config/remote_config_resources/build.gradle
```diff
@@ -0,0 +1,61 @@
...
+  }
+}
+
+apply plugin: 'com.android.library'
+
+android {
+  compileOptions {
+    sourceCompatibility 1.8
+    targetCompatibility 1.8
+  }
```
cynthiajoan, line 38: Is this necessary?

AlmostMatt: This file is necessary in order to build the new java source files for remote config.

I'm not sure what some fields like compileOptions and compileSdkVersion do exactly, but I copied most of this file from app_check_resources and it seems similar to many other _resources gradle files in the cpp sdk.

remote_config/src_java/com/google/firebase/remoteconfig/internal/cpp/JniConfigUpdateListener.java
```diff
@@ -0,0 +1,49 @@
...
+    nativeOnError(cListener, error.getCode().value());
+  }
+
+  /**
+   * These functions are implemented in the RemoteConfig C++ library.
+   */
+  private native void nativeOnUpdate(long cListener, ConfigUpdate configUpdate);
+
+  private native void nativeOnError(long cListener, int error);
+}
```
cynthiajoan, line 49: extra line in the end

remote_config/src/android/remote_config_android.cc
```diff
@@ -1073,7 +1201,55 @@ ConfigUpdateListenerRegistration*
...
+                     reinterpret_cast<jlong>(listener_wrapper));
+  FIREBASE_ASSERT(!util::CheckAndClearJniExceptions(env));
+
+  jobject j_local_registration = env->CallObjectMethod(
+      internal_obj_, config::GetMethodId(config::kAddOnConfigUpdateListener),
+      j_listener);
+  FIREBASE_ASSERT(!util::CheckAndClearJniExceptions(env));
+
+  // Create a global reference to the registration
+  jobject j_registration = env->NewGlobalRef(j_local_registration);
```
cynthiajoan, line 1222: If we need a global reference, we would need to define that in the class level, and delete it in app destroy phase.

cynthiajoan: Or you can delete it at the end of this function. but that kinda dismiss the purpose of the global ref

cynthiajoan: i see the pattern here, yeah you would need the global ref inside the callback, but you would need to delete it inside as well.

cynthiajoan (changes requested):
need to fix up the jni global ref deletion.

remote_config/src_java/com/google/firebase/remoteconfig/internal/cpp/JniConfigUpdateListener.java
AlmostMatt: added

remote_config/src/android/remote_config_android.cc
AlmostMatt: Ok. I delete the global ref in the removal function. I also added a bool to make sure the removal function is not called twice so that the ref is not deleted twice.

cynthiajoan (approved):

PR #1292, opened 2023-04-28
Add App Check to the packaging logic

a-maurice:
Adds App Check to the packaging and test logic.  Skipping the release notes, so that they can be done with other documentation.
***

***

### Type of Change
***

#### Notes


a-maurice:
I manually ran the tvOS tests separately, which passed: https://github.com/firebase/firebase-cpp-sdk/actions/runs/4863458976

app_check/integration_test/src/integration_test.cc
```diff
@@ -956,226 +795,4 @@ TEST_F(FirebaseAppCheckTest, TestFunctionsSuccess) {
   }
 }
 
-TEST_F(FirebaseAppCheckTest, TestFunctionsFailure) {
```
jonsimantov, line 959: Why is this test removed?

jonsimantov (approved):

app_check/integration_test/src/integration_test.cc
a-maurice: All the tests that involved running without App Check defined were seeing problems on mobile.  Once the Factory and Provider have been gotten for an App, it remembers it forever, so these would always fail.  On desktop, it was acting flakey, I'm not sure why since I never had problems with it locally, but would need more time to investigate it.

a-maurice:
Here is another run of the full App Check integration tests: https://github.com/firebase/firebase-cpp-sdk/actions/runs/4864098119

PR #1293, opened 2023-05-01
Cleanup for ConfigUpdateListenerRegistration

AlmostMatt:
### Description
public API change: returned ListenerRegistration instead of ListenerRegistration*

The idea is that the registration returned to the user is not dynamic memory and it can outlive RemoteConfig.
When the registration is created, it registers a cleanup function with remoteconfig's cleanupnotifier.
When RemoteConfig is destroyed, the cleanup notifier will cleanup all registrations. This clears out references to internal data.
Additionally, RemoteConfig will track all created internal registration objects, and it will delete these when RemoteConfig is destroyed.

implementation details:
moved existing registration class to an internal type
added a copyable registration class to wrap it
added a cleanupnotifier to remoteconfig to guarantee that Remove and Cleanup is called for all registrations

all of this follows the pattern of firestore's ListenerRegistration and CleanupFn.
one slight variation from the firestore pattern is that firestore has some separate code to trigger deletion of the internal registrations, but here the same CleanupNotifier is used to call Cleanup() on registrations and delete on internal registrations.

***
### Testing
Ran automatic integration tests which already create registration and terminate remote config.
***

### Type of Change
***

#### Notes


remote_config/src/config_update_listener_registration.cc
```diff
@@ -1,34 +1,131 @@
...
+
+using CleanupFnConfigUpdateListenerRegistration =
+    CleanupFn<ConfigUpdateListenerRegistration, internal::RemoteConfigInternal>;
 
 ConfigUpdateListenerRegistration::ConfigUpdateListenerRegistration(
-    std::function<void()> listener_removal_function)
-    : listener_removal_function_(listener_removal_function) {}
+    const ConfigUpdateListenerRegistration& registration)
+    : remote_config_(registration.remote_config_) {
+  internal_ = registration.internal_;
```
cynthiajoan, line 51: this can also be internal_(registration.internal_) after :

remote_config/src/config_update_listener_registration.cc
```diff
@@ -1,34 +1,131 @@
...
+#elif FIREBASE_PLATFORM_IOS || FIREBASE_PLATFORM_TVOS
+#include "remote_config/src/ios/remote_config_ios.h"
+#else
+#include "remote_config/src/desktop/remote_config_desktop.h"
+#endif  // FIREBASE_PLATFORM_ANDROID, FIREBASE_PLATFORM_IOS,
+        // FIREBASE_PLATFORM_TVOS
+
 namespace firebase {
 namespace remote_config {
+// ConfigUpdateListenerRegistration specific fact:
```
cynthiajoan, line 38: nit: comment indent, maybe run format

remote_config/src/config_update_listener_registration.cc
```diff
@@ -1,34 +1,131 @@
...
+  internal_ = nullptr;
+}
+
+ConfigUpdateListenerRegistration& ConfigUpdateListenerRegistration::operator=(
+    const ConfigUpdateListenerRegistration& registration) {
+  if (this == &registration) {
+    return *this;
+  }
+
+  remote_config_ = registration.remote_config_;
```
cynthiajoan, line 83: Add some comments to explain why need to unregister and then register

remote_config/src/desktop/remote_config_desktop.cc
```diff
@@ -582,11 +582,11 @@ const ConfigInfo RemoteConfigInternal::GetInfo() const {
   return configs_.metadata.info();
 }
 
-ConfigUpdateListenerRegistration*
+ConfigUpdateListenerRegistration
 RemoteConfigInternal::AddOnConfigUpdateListener(
     std::function<void(ConfigUpdate&&, RemoteConfigError)>
         config_update_listener) {
```
cynthiajoan, line 588: is parameter config_update_listener used inside this function?

cynthiajoan (commented):
I'm trying to follow the logic of CleanNotifier here. For android/ios/desktop, each platform instance hold an attribute of cleanup_, I found the places for these notifier to register and unregister obj to clean, but I cannot find the place that the Cleanup() function actually get called in remote_config's destruction phase. Did I miss anything?

AlmostMatt:
> I'm trying to follow the logic of CleanNotifier here. For android/ios/desktop, each platform instance hold an attribute of cleanup_, I found the places for these notifier to register and unregister obj to clean, but I cannot find the place that the Cleanup() function actually get called in remote_config's destruction phase. Did I miss anything?

cleanup_ is a non-pointer member variable, so it should be automatically destroyed when RemoteConfigInternal is destroyed.
The destructor of CleanupNotifier calls CleanupAll.

But I can also add an explicit call to CleanupAll to make this more obvious, which happens for some other classes that use CleanupNotifier.

remote_config/src/desktop/remote_config_desktop.cc
AlmostMatt: Technically no, but that is just because the desktop implementation is a stub for now. The Android and iOS implementations make use of the listener, and desktop will once it is implemented.

remote_config/src/config_update_listener_registration.cc
AlmostMatt: Done.

remote_config/src/config_update_listener_registration.cc
```diff
@@ -1,34 +1,145 @@
...
+    const ConfigUpdateListenerRegistration& registration) {
+  if (this == &registration) {
+    return *this;
+  }
+
+  // Copy assignment operator. Unregister cleanup for this object before
+  // changing the value of remote_config_ in case it was previously owned
+  // by a different instance of remote config.
+  CleanupFnConfigUpdateListenerRegistration::Unregister(this, remote_config_);
+  remote_config_ = registration.remote_config_;
```
cynthiajoan, line 93: This order change makes much more sense. Thanks for the clarification.

cynthiajoan (approved):

PR #1308, opened 2023-05-08
New Cache Config and LRU GC

wu-hui:


firestore/integration_test_internal/src/firestore_test.cc
```diff
@@ -1499,6 +1504,129 @@ TEST_F(FirestoreIntegrationTest, ClearPersistenceWhileRunningFails) {
...
+              ContainerEq(MapFieldValue{{"desc", FieldValue::String("eros")}}));
+}
+
+TEST_F(FirestoreIntegrationTest, NewCacheConfigForMemoryCacheWorks) {
+  auto* db = TestFirestore("new_memory_cache");
+  auto settings = db->settings();
+  settings.set_local_cache_settings(MemoryCacheSettings::Create());
+  db->set_settings(std::move(settings));
+
+  Await(db->Document("rooms/eros")
```
dconeybe, line 1546: Each test case has a lot of duplicated logic to verify that persistence is in effect or is _not_ in effect. Consider moving this logic into a helper function called something like VerifyPersistenceEnabled(bool) to perform the checks. It also has the benefit of increasing the readability the individual test cases.

firestore/integration_test_internal/src/settings_test.cc
```diff
@@ -94,6 +96,103 @@ TEST(SettingsTest, Equality) {
...
+  // Same as settings7
+  Settings settings9;
+  settings9.set_host("foo");
+  settings9.set_ssl_enabled(true);
+  settings9.set_local_cache_settings(
+      MemoryCacheSettings::Create().WithGarbageCollectorSettings(
+          MemoryLruGCSettings::Create().WithSizeBytes(kFiveMb)));
+
+  EXPECT_TRUE(settings1 == settings1);
+  EXPECT_TRUE(settings6 == settings4);
```
dconeybe, line 162: Shouldn't `settings6` and `settings4` compare _unequal_ since they have different persistence settings?

firestore/src/include/firebase/firestore/local_cache_settings.h
```diff
@@ -0,0 +1,308 @@
...
+ * limitations under the License.
+ */
+
+#ifndef FIREBASE_FIRESTORE_SRC_INCLUDE_FIREBASE_FIRESTORE_LOCAL_CACHE_SETTINGS_H_
+#define FIREBASE_FIRESTORE_SRC_INCLUDE_FIREBASE_FIRESTORE_LOCAL_CACHE_SETTINGS_H_
+
+#include <cstdint>
+#include <memory>
+
+#include "Firestore/core/src/api/settings.h"
```
dconeybe, line 23: I don't think we can include from the ios sdk in the "include" directory because this include also has to work with android. You may have to use a pimpl style like how `firestore/src/include/firebase/firestore/document_reference.h` does using `DocumentReferenceInternal`.

firestore/src/include/firebase/firestore/local_cache_settings.h
```diff
@@ -0,0 +1,308 @@
...
+ * pass it to an instance of `Settings` via `set_local_cache_settings()`, and
+ * use the `Settings` instance to configure the Firestore SDK.
+ */
+class PersistentCacheSettings final : public LocalCacheSettings {
+ public:
+  /** Create a default instance `PersistenceCacheSettings`. */
+  static PersistentCacheSettings Create();
+
+  /** Copy constructor. */
+  PersistentCacheSettings(const PersistentCacheSettings& other);
```
dconeybe, line 72: No move support?

firestore/src/include/firebase/firestore/local_cache_settings.h
```diff
@@ -0,0 +1,308 @@
...
+
+  /** Copy assignment. */
+  PersistentCacheSettings& operator=(const PersistentCacheSettings& other);
+
+  ~PersistentCacheSettings();
```
dconeybe, line 77: Declare the destructor as "override" to ensure that the base class' destructor is declared "virtual". I learned this trick recently from Mark :)

firestore/src/include/firebase/firestore/local_cache_settings.h
```diff
@@ -0,0 +1,308 @@
...
+ * This is the default cache type unless explicitly specified otherwise.
+ *
+ * To use, create an instance using `PersistentCacheSettings::Create`, then
+ * pass it to an instance of `Settings` via `set_local_cache_settings()`, and
+ * use the `Settings` instance to configure the Firestore SDK.
+ */
+class PersistentCacheSettings final : public LocalCacheSettings {
+ public:
+  /** Create a default instance `PersistenceCacheSettings`. */
+  static PersistentCacheSettings Create();
```
dconeybe, line 69: Should you make the default constructor private? I think the compiler-generated default constructor will be public, which doesn't seem to be desirable.

firestore/src/common/local_cache_settings.cc
```diff
@@ -0,0 +1,207 @@
...
+                const MemoryEagerGCSettings& rhs) {
+  return !(lhs == rhs);
+}
+
+bool operator==(const MemoryLruGCSettings& lhs,
+                const MemoryLruGCSettings& rhs) {
+  return &lhs == &rhs || (*lhs.settings_internal_ == *rhs.settings_internal_);
+}
+
+bool operator!=(const MemoryLruGCSettings& lhs,
```
dconeybe, line 201: Conventionally, the `operator!=` is defined in the .h file as "inline". Consider doing that here too.

firestore/src/common/settings.cc
```diff
@@ -51,12 +56,59 @@ void Settings::set_host(std::string host) { host_ = firebase::Move(host); }
...
+void Settings::set_local_cache_settings(const LocalCacheSettings& cache) {
+  if (used_legacy_cache_settings_) {
+    SimpleThrowIllegalState(
+        "Cannot mix set_local_cache_settings() with legacy cache api like "
+        "set_persistence_enabled() or set_cache_size_bytes()");
+  }
+
+  if (cache.kind() == api::LocalCacheSettings::Kind::kPersistent) {
+    local_cache_settings_ = std::make_shared<PersistentCacheSettings>(
+        static_cast<const PersistentCacheSettings&>(cache));
```
dconeybe, line 85: Yuck to the static_cast. Is there not a better way to do this? They have a common base class so can't local_cache_settings_ be a pointer to the base class type?

firestore/src/main/firestore_main.cc
```diff
@@ -188,8 +191,17 @@ Settings FirestoreInternal::settings() const {
   const api::Settings& from = firestore_core_->settings();
   result.set_host(from.host());
   result.set_ssl_enabled(from.ssl_enabled());
+
+  // TODO(wuandy): We use the deprecated API for default settings, but mark
+  // `used_legacy_cache_settings_` as false such that new settings API is not
+  // rejected by runtime checks. This should be removed when legacy API is
+  // removed.
+#pragma clang diagnostic push
```
dconeybe, line 199: What happens with gcc or msvc? Will this fail to build? I think there is a macro in absl to do this in a portable manner. If not, I'm pretty sure there's something in the ios sdk that does this exact thing in a portable manner.

wu-hui: Oh yeah, there is one macro from iOS SDK we could use. Thanks.

firestore/src/common/settings.cc
wu-hui: local_cache_settings_ IS a pointer to the base class. I am not sure how we can avoid this downcast given `cache` is a reference to the base type. Even if I create a copy-ctor for `LocalCacheSettings` in that ctor I'd still need to know which subclass of `LocalCacheSettings` it is and invoke the corresponding copy-ctor for that concrete type?

wu-hui (commented):
Thanks for the review!

wu-hui:
@dconeybe I took some nice stuff from your prototype and added to this PR, but I decided to keep the enum/static_cast around.

I did try to use absl::variant and `Impl` classes, and I do not quite like the result, mostly because in the end the approved api does not work well. I also think they add too much complexity for a small gain. enum/static_cast are very well understood and used in our code base in many places.

Please let me know your thoughts, we can discuss this further.

firestore/src/include/firebase/firestore/local_cache_settings.h
dconeybe: There are still includes from the "main" directory in `local_cache_settings.h`. This is fine for now but will fall apart once you implement android support. The APIs published in the "include" directory must be agnostic of the main/android implementation.

firestore/src/common/settings.cc
dconeybe: Fair enough. Thanks for considering the feedback. I still hate the static_cast but I can live with it.

dconeybe:
I took a quick look and left some minor comments. I'll do a more holistic review later today or early next week.

firestore/src/include/firebase/firestore/settings.h
```diff
@@ -137,13 +139,37 @@ class Settings final {
   void set_ssl_enabled(bool enabled);
 
   /**
+   * Returns a reference to the `LocalCacheSettings` instance
+   * used to configure this SDK.
+   */
+  const LocalCacheSettings& local_cache_settings();
```
dconeybe, line 145: Consider returning by value (i.e. returning a copy) from `local_cache_settings()` since the lifetime of the returned reference is... complicated. Specifically, the returned reference becomes invalid if there is a subsequent call to `set_local_cache_settings()` or `set_persistence_enabled()` or a few other functions. IMO in this particular instance the clarity and memory safety of returning by value far outweighs any performance benefit of returning by reference.

dconeybe (changes requested):
I'm still concerned about the "main" API bleeding into the public header file. I don't see how this will work on Android. Please correct me if I'm wrong about that. Other than that, I've just left a few comments. I'll do a deeper review once I feel that the API will work for both Android and main.

firestore/src/include/firebase/firestore/local_cache_settings.h
```diff
@@ -0,0 +1,308 @@
...
+
+class PersistentCacheSettingsInternal;
+class MemoryCacheSettingsInternal;
+class MemoryLruGCSettingsInternal;
+class MemoryEagerGCSettingsInternal;
+
+/**
+ * Abstract class implemented by all supported cache settings.
+ *
+ * `PersistentCacheSettings` and `MemoryCacheSettings` are the only cache types
```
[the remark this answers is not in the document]
wu-hui: Oh, learnt a trick today!

firestore/src/include/firebase/firestore/local_cache_settings.h
```diff
@@ -0,0 +1,308 @@
...
+
+ protected:
+  enum class Kind { kMemory, kPersistent };
+
+ private:
+  friend class FirestoreInternal;
+  friend class Settings;
+
+  virtual Kind kind() const = 0;
+  virtual const api::LocalCacheSettings& core_cache_settings() const = 0;
```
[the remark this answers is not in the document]
wu-hui: Addressed.

firestore/src/include/firebase/firestore/local_cache_settings.h
```diff
@@ -0,0 +1,308 @@
...
+   * for the on-disk data set to the given number in term of number of bytes,
+   * and return the new setting instance.
+   *
+   * If the cache grows beyond this size, Firestore SDK will start removing data
+   * that hasn't been recently used. The SDK does not guarantee that the cache
+   * will stay below that size, only that if the cache exceeds the given size,
+   * cleanup will be attempted.
+   *
+   * By default, persistence cache is enabled with a cache size of 100 MB. The
+   * minimum value is 1 MB.
```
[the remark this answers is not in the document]
wu-hui: Done.

firestore/src/include/firebase/firestore/settings.h
wu-hui: I agree with what you said, but I don't think it's possible to return a copy of `LocalCacheSetting` because it is abstract?

We could consider returning the underlying shared_ptr though..this is not in the approved proposal but seems like a good amendment?

wu-hui (commented):
Thanks! I have to admit I did not think about how to implement Android enough!

firestore/src/include/firebase/firestore/local_cache_settings.h
dconeybe: Remove all traces of the strings "main" and "api::" from `local_cache_settings.h`. It still has this line:

```
#include "firestore/src/main/local_cache_settings_main.h"
```

which won't work when targeting Android.

dconeybe (changes requested):
`local_cache_settings.h` still has an include line for `local_cache_settings_main.h`.  `local_cache_settings.h` must not rely on anything specific to the ios implementation, since the same .h file must be used for Android.

firestore/src/common/local_cache_settings.cc
```diff
@@ -0,0 +1,138 @@
...
+}
+
+bool operator==(const MemoryGarbageCollectorSettings& lhs,
+                const MemoryGarbageCollectorSettings& rhs) {
+  return lhs.internal() == rhs.internal();
+}
+
+bool operator==(const MemoryEagerGCSettings& lhs,
+                const MemoryEagerGCSettings& rhs) {
+  return &lhs == &rhs || (*lhs.settings_internal_ == *rhs.settings_internal_);
```
dconeybe, line 129: Is there a compelling reason to optimize `operator==()` with `&lhs == &rhs`? I can see it as a performance optimization, but I don't think that the performance benefit outweighs the detraction from consistency with the other `operator==` overloads. Also applies to the `operator==()` for `MemoryLruGCSettings`.

firestore/src/include/firebase/firestore/settings.h
dconeybe: Hmm fair enough. Maybe returning a const reference is fine. Consider documenting, though, the situations under which the returned reference becomes invalid.

firestore/src/include/firebase/firestore/settings.h
```diff
@@ -137,13 +139,37 @@ class Settings final {
...
+   *
+   * See documentation of `PersistentCacheSettings` to under the default
+   * settings.
+   *
+   * @param cache_settings Settings object to configue this SDK.
+   */
+  void set_local_cache_settings(const LocalCacheSettings& cache_settings);
+
+  /**
+   * NOTE: This method is deprecated in favor of `set_local_cache_settings()`.
```
dconeybe, line 160: nit: Move this deprecation note to _after_ the "Enables or disables local persistent storage" line. Otherwise, the summary documentation in generated docs and IDE hover-overs will simply read as the deprecation notices. Also in `set_cache_size_bytes()`.

firestore/src/common/settings.cc
```diff
@@ -44,18 +47,66 @@ std::string ToStr(int64_t v) {
...
 #endif
 
 void Settings::set_host(std::string host) { host_ = std::move(host); }
 
 void Settings::set_ssl_enabled(bool enabled) { ssl_enabled_ = enabled; }
 
+const LocalCacheSettings& Settings::local_cache_settings() {
+  if (cache_settings_source_ == CacheSettingsSource::kOld) {
+    if (is_persistence_enabled()) {
+      local_cache_settings_ = std::make_shared<PersistentCacheSettings>(
```
dconeybe, line 63: It seems pretty weird to mutate a member variable (i.e. `local_cache_settings_`) in this "getter" member function. Could you find a way to instead make this member function `const` and initialize `local_cache_settings_` when it is mutated?

firestore/src/common/settings.cc
```diff
@@ -70,6 +121,40 @@ std::ostream& operator<<(std::ostream& out, const Settings& settings) {
...
+bool operator==(const Settings& lhs, const Settings& rhs) {
+  if (lhs.host() != rhs.host()) {
+    return false;
+  }
+
+  if (lhs.is_ssl_enabled() != rhs.is_ssl_enabled()) {
+    return false;
+  }
+
+  if (lhs.cache_settings_source_ != rhs.cache_settings_source_) {
```
dconeybe, line 133: What if the settings sources are different but represent the same cache configuration? Do the other SDKs consider this "equal" or "unequal"? We should be consistent with the other SDKs here.

firestore/src/common/settings.cc
```diff
@@ -70,6 +121,40 @@ std::ostream& operator<<(std::ostream& out, const Settings& settings) {
...
+    return false;
+  }
+
+  if (*lhs.local_cache_settings_ != *rhs.local_cache_settings_) {
```
dconeybe, line 137: What happens if `lhs` and `rhs` have different types of persistence? For example, what if `*lhs.local_cache_settings_` is memory local cache and `*rhs.local_cache_settings_` is persistent local cache? Which `operator==` is used to compare them? I'm actually surprised this compiles, and doesn't crash if they are different types.

firestore/src/main/firestore_main.cc
```diff
@@ -203,8 +216,16 @@ void FirestoreInternal::set_settings(Settings from) {
   api::Settings settings;
   settings.set_host(std::move(from.host()));
   settings.set_ssl_enabled(from.is_ssl_enabled());
-  settings.set_persistence_enabled(from.is_persistence_enabled());
-  settings.set_cache_size_bytes(from.cache_size_bytes());
+  // TODO(wuandy): Checking `from.local_cache_settings_` is required, because
+  // FirestoreInternal::settings() overrides used_legacy_cache_settings_. All
```
dconeybe, line 220: What is "used_legacy_cache_settings_"? I think maybe the name has changed. Please update this comment.

firestore/src/main/firestore_main.h
```diff
@@ -48,6 +48,7 @@ class Firestore;
 class ListenerRegistrationInternal;
 class Transaction;
 class WriteBatch;
+class Settings;
```
dconeybe, line 51: nit: Put the forward declaration of `Settings` in its correct alphabetical ordering among the other forward declarations.

firestore/src/main/local_cache_settings_main.h
```diff
@@ -0,0 +1,177 @@
...
+#include "firebase/firestore/local_cache_settings.h"
+
+namespace firebase {
+namespace firestore {
+
+class LocalCacheSettingsInternal {
+ public:
+  friend bool operator==(const LocalCacheSettingsInternal& lhs,
+                         const LocalCacheSettingsInternal& rhs) {
+    return &lhs == &rhs || lhs.core_settings() == rhs.core_settings();
```
dconeybe, line 33: Is this checking for `&lhs == &rhs` strictly required? If no, I recommend removing it since it is a performance optimization that likely is not warranted. Applies here and elsewhere in this file.

firestore/src/main/local_cache_settings_main.h
```diff
@@ -0,0 +1,177 @@
...
+    : public LocalCacheSettingsInternal {
+ public:
+  explicit PersistentCacheSettingsInternal() = default;
+
+  friend bool operator==(const PersistentCacheSettingsInternal& lhs,
+                         const PersistentCacheSettingsInternal& rhs) {
+    return &lhs == &rhs || lhs.settings_ == rhs.settings_;
+  }
+
+  PersistentCacheSettingsInternal WithSizeBytes(int64_t size) {
```
dconeybe, line 49: Can `WithSizeBytes()` be `const` (as well as other "With" member functions declared/defined in this file)?

dconeybe (changes requested):
I think you got the "core" dependencies out of the public headers. Nice job! I've reviewed the code, but haven't looked deeply at the tests. I'll look at the tests in the next round of review.

PR #1309, opened 2023-05-08
Fix `credential.is_valid()` crash on iOS

jonsimantov:
- null check `Credential.impl_` in `CredentialFromImpl()` and `PhoneAuthCredentialFromImpl()`
- Check if Auth API received an invalid Credential
- Add a basic test to just verify if `Credential.is_valid()` crashes. More comprehensive test should be added through b/281590792

***

GHA
***

### Type of Change
***

#### Notes

### DIFF
--- a/auth/integration_test/src/integration_test.cc
+++ b/auth/integration_test/src/integration_test.cc
@@ -310,6 +310,7 @@ bool FirebaseAuthTest::WaitForCompletion(
     EXPECT_EQ(result_ptr->additional_user_info.provider_id, provider_id);
     EXPECT_EQ(result_ptr->user.uid(), auth_->current_user().uid());
     EXPECT_TRUE(auth_->current_user().is_valid());
+    EXPECT_TRUE(result_ptr->credential.is_valid());
   }
   return succeeded;
 }
--- a/auth/integration_test/src/integration_test.cc
+++ b/auth/integration_test/src/integration_test.cc
@@ -273,6 +273,9 @@ bool FirebaseAuthTest::WaitForCompletion(
         EXPECT_TRUE(auth_->current_user().is_valid());
         EXPECT_EQ(future.result()->user.uid(), auth_->current_user().uid())
             << "User returned by Future doesn't match User in Auth";
+        // Expect no crash.
+        // TODO(b/281590792): Properly validate AuthResult fields case-by-case.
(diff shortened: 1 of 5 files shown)


auth/integration_test/src/integration_test.cc
```diff
@@ -310,6 +310,7 @@ bool FirebaseAuthTest::WaitForCompletion(
     EXPECT_EQ(result_ptr->additional_user_info.provider_id, provider_id);
     EXPECT_EQ(result_ptr->user.uid(), auth_->current_user().uid());
     EXPECT_TRUE(auth_->current_user().is_valid());
+    EXPECT_TRUE(result_ptr->credential.is_valid());
```
chkuang-g, line 313: Ah, AuthResult.credential is not necessarily valid all the time, even if the request is successful. I think some of the API simply won't have credential information

auth/src/ios/common_ios.h
```diff
@@ -91,8 +91,8 @@ static inline FIRAuth *_Nonnull AuthImpl(AuthData *_Nonnull auth_data) {
 
 /// Convert from the void* credential implementation pointer into the Obj-C
 /// FIRAuthCredential pointer.
-static inline FIRAuthCredential *_Nonnull CredentialFromImpl(void *_Nonnull impl) {
-  return static_cast<FIRAuthCredentialPointer *>(impl)->get();
+static inline FIRAuthCredential *_Nullable CredentialFromImpl(void *_Nullable impl) {
+  return impl ? static_cast<FIRAuthCredentialPointer *>(impl)->get() : nil;
```
chkuang-g, line 95: Also need to change the function below this one,

chkuang-g: Also, many iOS objective-c API expect nonnull.
https://firebase.google.com/docs/reference/ios/firebaseauth/api/reference/Classes/FIRAuth
https://github.com/firebase/firebase-cpp-sdk/blob/main/auth/src/ios/auth_ios.mm#L460

auth/integration_test/src/integration_test.cc
chkuang-g: done

auth/src/ios/common_ios.h
chkuang-g: Done

PR #1313, opened 2023-05-10
Modify heartbeat local storage to use wide strings on Windows.

jonsimantov:
Two changes in this PR:
- Add Windows international path support to the heartbeat storage library.
- Ensure that multiple desktop apps don't write to the same heartbeat file by adding the package name to the heartbeat filename. Previously, only the app name (which is usually "__FIRAPPDEFAULT") was used, so if there were multiple apps installed on a machine (even completely unrelated apps), they would all use the same heartbeat file.

***

The integration test in this PR was run with added international characters in the path. (That test change has since been reverted.) Additionally, it was run manually *without* the wstring fix to verify that writing the heartbeat data to an international path returned an error prior to this fix, and this change actualy fixes it.
***

### Type of Change
***

#### Notes

### DIFF
--- a/app/src/heartbeat/heartbeat_storage_desktop.h
+++ b/app/src/heartbeat/heartbeat_storage_desktop.h
@@ -60,7 +60,7 @@ class HeartbeatStorageDesktop {
       const LoggedHeartbeats& heartbeats_struct) const;
 
   // local variables for state
-#ifdef FIREBASE_PLATFORM_WINDOWS
+#if FIREBASE_PLATFORM_WINDOWS
   // Use a wide string on Windows, to support international characters in the
   // path.
   std::wstring filename_;
--- a/app/src/heartbeat/heartbeat_storage_desktop.h
+++ b/app/src/heartbeat/heartbeat_storage_desktop.h
@@ -23,6 +23,7 @@
 #include <vector>
 
 #include "app/logged_heartbeats_generated.h"
+#include "app/src/include/firebase/internal/platform.h"
 #include "app/src/logger.h"
 
 namespace firebase {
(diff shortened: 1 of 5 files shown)


app/src/heartbeat/heartbeat_storage_desktop.h
```diff
@@ -61,7 +68,13 @@ class HeartbeatStorageDesktop {
       const LoggedHeartbeats& heartbeats_struct) const;
 
   // local variables for state
+#if FIREBASE_PLATFORM_WINDOWS
+  // Use a wide string on Windows, to support international characters in the
+  // path.
+  std::wstring filename_;
```
AlmostMatt, line 74: Does code like this just work as-is with wstring?
std::ifstream file(filename_,

AlmostMatt (approved):
Seems like an ok change, assuming it works correctly as tested.

It would probably be possible to have a type alias that is either wstring or string depending on platform to reduce the number of times that FIREBASE_PLATFORM_WINDOWS is checked, but then you would have to include whatever file defines this type alias, and it would be less obvious to the reader what type the variable is, so maybe not much benefit to that.

app/src/heartbeat/heartbeat_storage_desktop.h
jonsimantov: Yep, on Windows you can initialize an fstream with a wstring path. (You can't do this on non-Windows systems.)

jonsimantov:
> Seems like an ok change, assuming it works correctly as tested.
> 
> It would probably be possible to have a type alias that is either wstring or string depending on platform to reduce the number of times that FIREBASE_PLATFORM_WINDOWS is checked, but then you would have to include whatever file defines this type alias, and it would be less obvious to the reader what type the variable is, so maybe not much benefit to that.

I thought about doing that, and I had an older PR that did (in a different part of the code), but it felt more straightforward to just do it this way. Once we support C++17 we can just use std::path_t.

PR #1316, opened 2023-05-11
On Windows desktop, use wide path strings in Remote Config.

jonsimantov:
This PR moves the Remote Config cached data into the user's AppDataDir, under a directory containing the package name and the app name.

In doing so, it needs to update the way that it handles paths on Windows to support international characters via wide strings.
***

Integration tests in this PR, which were run with international characters added to the path. Also verified the fix locally on a Windows machine.
***

### Type of Change
***

#### Notes

### DIFF
--- a/remote_config/src/desktop/file_manager.cc
+++ b/remote_config/src/desktop/file_manager.cc
@@ -22,7 +22,7 @@
 #include <utility>
 
 #include "app/src/include/firebase/internal/platform.h"
-#include "app/src/util.h"
+#include "app/src/filesystem.h"
 #include "remote_config/src/desktop/config_data.h"
 
 #if FIREBASE_PLATFORM_WINDOWS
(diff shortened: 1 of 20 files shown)


remote_config/src/desktop/file_manager.cc
```diff
@@ -21,14 +21,33 @@
...
 namespace firebase {
 namespace remote_config {
 namespace internal {
 
-RemoteConfigFileManager::RemoteConfigFileManager(const std::string& file_path)
-    : file_path_(file_path) {}
+RemoteConfigFileManager::RemoteConfigFileManager(const std::string& filename,
+                                                 const firebase::App& app) {
+  std::string app_data_prefix = std::string(app.options().package_name()) +
+                                "/" + app.name() + "/" + "TẽstPạth";
```
AlmostMatt, line 40: Remove the /TestPath part of the prefix.

AlmostMatt (approved):
Looks good, just resolve the one comment

remote_config/src/desktop/file_manager.cc
AlmostMatt: Created https://github.com/firebase/firebase-cpp-sdk/pull/1323

PR #1318, opened 2023-05-11
Enable multi database in android

milaGGL:
- add `database_name_` private property to Firestoreinternal class to store database name on creation and read on delete
- update firestore `GetInstance` JNI method call in android folder to include database id string
- update validation test to fix github check failure


firestore/src/android/firestore_android.cc
```diff
@@ -260,14 +260,16 @@ Local<LoadBundleTaskInternal> CreateLoadBundleTask(Env& env,
...
-FirestoreInternal::FirestoreInternal(App* app) {
+FirestoreInternal::FirestoreInternal(App* app, const char* database_id) {
   FIREBASE_ASSERT(app != nullptr);
   if (!Initialize(app)) return;
   app_ = app;
 
   Env env = GetEnv();
   Local<Object> platform_app(env.get(), app_->GetPlatformApp());
-  Local<Object> java_firestore = env.Call(kGetInstance, platform_app);
+  Local<String> java_path = env.NewStringUtf(database_id);
```
tom-andersen, line 270: nit: could you rename `java_path` to `java_database_id`.

firestore/src/main/firestore_main.h
```diff
@@ -184,6 +186,7 @@ class FirestoreInternal {
   std::unordered_set<ListenerRegistrationInternal*> listeners_;
 
   std::shared_ptr<util::Executor> transaction_executor_;
+  const char* database_name_;
```
tom-andersen, line 189: When does `database_name_` delete? Because this is a pointer, you need to manage lifecycle yourself. I have traced logic, and it appears the pointer is created by customer, and can be deleted by customer as well during the lifetime of Firestore object. We need to atleast make a copy, but maybe just use `std::string` here instead.

In case I missed something, could you please explain the lifecycle of the `database_name_` pointer?

firestore/src/android/firestore_android.h
```diff
@@ -211,6 +213,8 @@ class FirestoreInternal {
   std::unique_ptr<PromiseFactory<AsyncFn>> promises_;
 
   CleanupNotifier cleanup_;
+
+  const char* database_name_;
```
tom-andersen, line 217: Please see comment in `firestore_main.h` for `database_name_`.

tom-andersen (changes requested):

firestore/integration_test_internal/src/firestore_test.cc
```diff
@@ -1402,6 +1404,62 @@ TEST_F(FirestoreTest, RestartFirestoreLeadsToNewInstance) {
   delete db1;
 }
 
+TEST_F(FirestoreTest, CanCreateMultipleDatabases) {
```
milaGGL, line 1407: 2 new test cases added: CanCreateMultipleDatabases, CanTerminateMultipleDatabases

tom-andersen (changes requested):

firestore/integration_test_internal/src/firestore_test.cc
```diff
@@ -1404,6 +1404,62 @@ TEST_F(FirestoreTest, RestartFirestoreLeadsToNewInstance) {
...
+  App* app = App::GetInstance();
+  Firestore* db1 = TestFirestoreWithDatabaseId(app->name(), "db1");
+  Firestore* db2 = TestFirestoreWithDatabaseId(app->name(), "db2");
+  EXPECT_NE(db1, db2);
+
+  // Create collections with same name in different databases.
+  DocumentReference doc1 = db1->Collection("abc").Document();
+  DocumentReference doc2 = db2->Collection("abc").Document();
+
+  // Databases can store and retrieve documents without interfering with each
```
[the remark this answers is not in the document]
milaGGL: If they interfere, or we created a same instance ignoring the db name, then it should get the same doc. hmmmm, I see ur point, i should rearrange the order of create and read of docs.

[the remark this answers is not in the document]
milaGGL: Re-arranged. If they interfere, then doc1 and doc2 should be actually the same doc, and the "foo" should be "bar2" for both line 1433 and 1437.

firestore/src/android/firestore_android.cc
tom-andersen: Last one, then I will approve.

milaGGL: This is somehow missed while committing

tom-andersen (approved):
LGTM

Issue #1247, opened 2023-03-27
[Question]  Segfault on firebase::gma::Initialize


DeexCoding:
### [REQUIRED] Please fill in the following fields:

  * Pre-built SDK from the [website](https://firebase.google.com/download/cpp) or open-source from this repo: Pre-built
  * Firebase C++ SDK version: 10.6.0
  * Main Firebase Components in concern: Google Mobile Ads (Auth, Database, etc.)
  * Other Firebase Components in use: None (Auth, Database, etc.)
  * Platform you are using the C++ SDK on: Windows (Mac, Windows, or Linux)
  * Platform you are targeting: Android (iOS, Android, and/or desktop)

### [REQUIRED] Please describe the question here:

I am trying to initialize GMA, but it just throws a segfault and nothing else (project name is AndroidStudioTest, weird that the segfault leaves "an" out):

`A/libc: Fatal signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x0 in tid 23276 (Thread-6), pid 27809 (droidstudiotest)`

And there is no other error

Here is the relevant code:

`int InitGMA(struct android_app* app)
{
    JNIEnv* env = 0;
	JavaVM* jniiptr = app->activity->vm;
	jint attachResult = jniiptr->AttachCurrentThread(&env, 0);

    firebase::InitResult result;
    firebase::Future<firebase::gma::AdapterInitializationStatus> future = 
        firebase::gma::Initialize(env, app->activity->clazz, &result);`

DeexCoding:
Call stack shows:
art_sigsegv_fault 0x00000076bc67e4b0
art::FaultManager::HandleFault(int, siginfo *, void *) 0x00000076bc67e9c4
art::SignalChain::Handler(int, siginfo *, void *) 0x000000567665aea4
<unknown> 0x0000007954307668
_JNIEnv::NewStringUTF(const char *) 0x00000076460493d8
firebase::util::FindOrLoadClassFromLoaders(_JNIEnv *, const char *) 0x00000076460493c4
firebase::util::FindClass(_JNIEnv *, const char *) 0x00000076460493c4
firebase::gma::Initialize(_JNIEnv *, _jobject *, firebase::InitResult *) 0x00000076460989dc
InitGMA(android_app *) AdsWrapper.cpp:50
...

AlmostMatt:
Just to clarify, you are using GMA without a FirebaseApp, right?

[...]

PR #1320, opened 2023-05-11
Fix crash on gma::Initialize without a Firebase App

AlmostMatt:
### Description
Resolves a crash on Android when calling gma::Initialize without a firebase App (#1247)

The crash occured because util::FindClass was called before util had been initialized. The fix is to reorder some lines so that util is initialized before calling util::FindClass.

Also updated existing tests to wait for the initialization future before calling terminate to resolve some flaky test failures.
***
### Testing
Added a new integration test case to initialize GMA without a Firebase App.
Ran integration tests locally on a real device and also on the github runners:
https://github.com/firebase/firebase-cpp-sdk/actions/runs/5731312895
***

### Type of Change
***

#### Notes


jonsimantov (commented):
Should we get this PR merged in? (With the release note moved to the next version, that is.)

AlmostMatt:
> Should we get this PR merged in? (With the release note moved to the next version, that is.)

The reason I didn't merge this previously is because the tests fail with this change on the github runners (despite passing for me locally at the time).
But I will try to revisit this and figure out why the tests were failing and see if this can be resolved.

gma/integration_test/src/integration_test.cc
```diff
@@ -340,6 +350,28 @@ void FirebaseGmaPreInitializationTests::SetUpTestSuite() {
...
+#else   // !defined(ANDROID)
+  ::firebase::gma::Initialize(&result);
+#endif  // defined(ANDROID)
+  EXPECT_EQ(result, ::firebase::kInitResultSuccess);
+  LogDebug("Successfully initialized GMA.");
+  // Workaround: GMA does some of its initialization in the main
+  // thread, so if you terminate it too quickly after initialization
+  // it can cause issues.  Add a small delay here in case most of the
+  // tests are skipped.
+  ProcessEvents(1000);
```
jonsimantov, line 370: Is it cleaner to wait on the Future that Initialize returns?

release_build_files/readme.md
```diff
@@ -678,6 +678,7 @@ code.
     - Storage (Desktop): Fixed a crash on Windows when uploading files from a
       path containing non-ANSI characters (Unicode above U+00FF).
     - Firestore: Added MultiDb support. ([#1321](https://github.com/firebase/firebase-cpp-sdk/pull/1321)).
+    - GMA (Android): Fixed a crash when Initializing GMA without a Firebase App.
```
jonsimantov, line 681: Remember to move this up

gma/integration_test/src/integration_test.cc
AlmostMatt: Yea, I updated this and the main testsuite teardown to wait for initialization futures instead of an arbitrary 1 second duration.

gma/src/android/gma_android.cc
```diff
@@ -298,7 +294,11 @@ Future<AdapterInitializationStatus> Initialize(JNIEnv* env, jobject activity,
     return Future<AdapterInitializationStatus>();
   }
 
-  if (!util::Initialize(env, activity)) {
+  // GMA requires Google Play services if the class
+  // "com.google.android.gms.ads.internal.ClientApi" does not exist.
+  if (!util::FindClass(env, "com/google/android/gms/ads/internal/ClientApi") &&
+      google_play_services::CheckAvailability(env, activity) !=
+          google_play_services::kAvailabilityAvailable) {
     if (init_result_out) {
```
jonsimantov, line 302: In this scenario, since util::Initialize was previously called, we need to call util::Terminate to ensure that the refcount is correct.

gma/integration_test/src/integration_test.cc
```diff
@@ -228,11 +234,11 @@ void FirebaseGmaTest::SetUpTestSuite() {
...
-  // Workaround: GMA does some of its initialization in the main
-  // thread, so if you terminate it too quickly after initialization
-  // it can cause issues.  Add a small delay here in case most of the
-  // tests are skipped.
-  ProcessEvents(1000);
+  // GMA does some of its initialization in the main thread, so if you terminate
+  // it before initialization has completed, it can cause issues. So wait for
+  // any pending GMA initialization to be completed before calling terminate.
+  auto initialize_future = firebase::gma::InitializeLastResult();
+  WaitForCompletion(initialize_future, "gma::Initialize");
```
jonsimantov, line 241: This could just be WaitForCompletion(firebase::gma::InitializeLastResult(), "gma::Initialize") for brevity.

gma/src/android/gma_android.cc
AlmostMatt: Good point. I see that Util::Terminate is called in another fail case later in this function, so I will call it in this fail case as well. Done.

gma/integration_test/src/integration_test.cc
AlmostMatt: Done

gma/integration_test/src/integration_test.cc
```diff
@@ -340,6 +350,28 @@ void FirebaseGmaPreInitializationTests::SetUpTestSuite() {
...
+#if defined(ANDROID)
+  ::firebase::gma::Initialize(app_framework::GetJniEnv(),
+                              app_framework::GetActivity(), &result);
+#else   // !defined(ANDROID)
+  ::firebase::gma::Initialize(&result);
+#endif  // defined(ANDROID)
+  EXPECT_EQ(result, ::firebase::kInitResultSuccess);
+  auto initialize_future = firebase::gma::InitializeLastResult();
+  WaitForCompletion(initialize_future, "gma::Initialize");
+  EXPECT_EQ(initialize_future.error(), 0) << initialize_future.error_message();
```
jonsimantov, line 368: One last nit, I believe WaitForCompletion actually already checks that the error is 0 (unless you pass in a different error to expect).

jonsimantov (approved):

gma/integration_test/src/integration_test.cc
AlmostMatt: True, done.

PR #1335, opened 2023-05-29
OR Query Implementation

tom-andersen:


firestore/src/include/firebase/firestore/filter.h
```diff
@@ -0,0 +1,96 @@
...
+  static Filter LessThan(const FieldPath& field, const FieldValue& value);
+  static Filter LessThanOrEqualTo(const FieldPath& field,
+                                  const FieldValue& value);
+  static Filter In(const FieldPath& field,
+                   const std::vector<FieldValue>& values);
+  static Filter NotIn(const FieldPath& field,
+                      const std::vector<FieldValue>& values);
+
+  template <class... Filters>
+  static Filter Or(const Filter& filter, const Filters&... filters);
```
ehsannas, line 65: I just realized that the final decision was to allow empty filters (e.g. `Or()`) on other platforms. This came out of an API discussion where folks argued that when the developer is using a UI to build composite filters, if all filters are unselected, then the composite filter would contain no filters, and it would greatly simplify the developers' code if they don't have to guard against the corner case of "empty composite filter".

So other platforms (android/ios/web) currently allow empty Or and empty And filters. We should do the same thing here. (I realize the API proposal has to be updated as well).

So I think we'll have to go with 
```cpp
  template <class... Filters>
  static Filter Or(const Filters&... filters);
```
and change all relevant functions and constructors to adhere to this.

firestore/src/common/filter.cc
```diff
@@ -0,0 +1,172 @@
...
+  return FilterInternal::Or(filter, filters...);
+}
+
+template <typename... Filters>
+Filter Filter::And(const Filter& filter, const Filters&... filters) {
+  return FilterInternal::And(filter, filters...);
+}
+
+bool operator==(const Filter& lhs, const Filter& rhs) {
+  return &lhs == &rhs || *lhs.internal_ == *rhs.internal_;
```
ehsannas, line 168: I think you should use the helper function that was created and used [elsewhere](https://cs.opensource.google/firebase-sdk/firebase-cpp-sdk/+/main:firestore/src/common/field_value.cc;l=344-346?q=field_value.cc&ss=firebase-sdk%2Ffirebase-cpp-sdk)

```cpp
return EqualityCompare(lhs.internal_, rhs.internal_);
```

firestore/src/main/filter_main.cc
```diff
@@ -0,0 +1,62 @@
...
+      case FilterInternal::Composite:
+        return *static_cast<const CompositeFilterInternal*>(&lhs) ==
+               *static_cast<const CompositeFilterInternal*>(&rhs);
+      case FilterInternal::Unary:
+        return *static_cast<const UnaryFilterInternal*>(&lhs) ==
+               *static_cast<const UnaryFilterInternal*>(&rhs);
+    }
+  }
+  return false;
+}
```
ehsannas, line 59: the idiom of using type a "type" enum in the class and using it in the comparator implementation SGTM.

The alternative to the static_cast would be to use an `absl::variant` (but I think it'd add unnecessary complexity here). After moving to C++17, we could use `std::variant`

firestore/src/main/filter_main.h
```diff
@@ -0,0 +1,138 @@
...
+  static Filter CompositeFilter(CompositeOperator op, const Filter& filter) {
+    return filter;
+  }
+
+  template <typename... Filters>
+  static Filter CompositeFilter(CompositeOperator op,
+                                const Filter& filter,
+                                const Filters&... filters);
+};
+
```
ehsannas, line 130: add
```cpp
bool operator==(const FilterInternal& lhs, const FilterInternal& rhs);
```

firestore/src/main/unary_filter_main.h
```diff
@@ -0,0 +1,63 @@
...
+
+ private:
+  UnaryFilterInternal* clone() override;
+
+  const bool allow_arrays_ = false;
+  const FieldPath path_;
+  const core::FieldFilter::Operator op_;
+  const FieldValue value_;
+};
+
```
ehsannas, line 54: add

```cpp
bool operator==(const UnaryFilterInternal& lhs, const UnaryFilterInternal& rhs);
```

firestore/src/main/composite_filter_main.h
```diff
@@ -0,0 +1,59 @@
...
+  friend bool operator==(const CompositeFilterInternal& lhs,
+                         const CompositeFilterInternal& rhs);
+
+ private:
+  CompositeFilterInternal* clone() override;
+
+  const core::CompositeFilter::Operator op_;
+  const std::vector<std::shared_ptr<FilterInternal>> filters_;
+};
+
```
ehsannas, line 50: add

```cpp
bool operator==(const CompositeFilterInternal& lhs, const CompositeFilterInternal& rhs);
```

ehsannas (changes requested):
The github actions lint warnings are also valid.

ehsannas (changes requested):
The github actions lint warnings are also valid.

firestore/src/main/composite_filter_main.cc
```diff
@@ -0,0 +1,54 @@
...
+
+namespace firebase {
+namespace firestore {
+
+template <typename... Ts>
+CompositeFilterInternal::CompositeFilterInternal(
+    core::CompositeFilter::Operator op, const Filter& filter, Ts... filters)
+    : FilterInternal(FilterType::Composite),
+      op_(op),
+      filters_(filter, filters...) {}
```
ehsannas, line 32: `filters_` is a vector of shared pointers. I'm pretty sure that the unpacking (`filters...`) will copy the `FilterInternal` values into that vector, but we should check this. (i.e. I'm worried that the shared pointers point to somewhere in the stack that'll be garbage after `filters...` goes out of scope).

ehsannas: uh.. wait, how is this initializing from Filter to shared_ptr<FilterInternal> ?

firestore/src/main/filter_main.h
```diff
@@ -0,0 +1,138 @@
...
+  using CompositeOperator = core::CompositeFilter::Operator;
+
+  static Filter UnaryFilter(const FieldPath& field_path,
+                            UnaryOperator op,
+                            const FieldValue& value);
+  static Filter UnaryFilter(const FieldPath& field_path,
+                            UnaryOperator op,
+                            const std::vector<FieldValue>& values);
+
+  static Filter CompositeFilter(CompositeOperator op, const Filter& filter) {
```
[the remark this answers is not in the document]
tom-andersen: When there is only 1 filter within composite, there is no need to wrap and return a composite. Instead simply return the filter. I believe this shortcut exists elsewhere.

firestore/src/main/composite_filter_main.cc
tom-andersen: I completely reworked this, passing a vector of filters to internal classes instead.

firestore/integration_test_internal/src/filter_test.cc
```diff
@@ -0,0 +1,472 @@
...
+  QuerySnapshot snapshot1 =
+      ReadDocuments(collection.Where(Filter::And(filter_xa, filter_yb)));
+  EXPECT_EQ(std::vector<MapFieldValue>({doc_ab}),
+            QuerySnapshotToValues(snapshot1));
+
+  QuerySnapshot snapshot2 =
+      ReadDocuments(collection.Where(Filter::Or(filter_xa, filter_yb)));
+  EXPECT_EQ(std::vector<MapFieldValue>({doc_aa, doc_ab, doc_bb}),
+            QuerySnapshotToValues(snapshot2));
+}
```
ehsannas, line 467: Can you please add a few more tests?

1- have nested composite filters
* `col.where(and(A, or(B, and(C, D))))`

2- have empty composite filters
* `query.where(and())` result set should be same as `query` result set
* `query.where(or())` result set should be same as `query` result set

3- Have composite filter with one filter inside it
* `query.where(and(A))` should have the same result set as `query.where(A)`
* `query.where(or(A))` should have the same result set as `query.where(A)`

4- mix of 1 and 2 and 3
* `query.where(and(or(and(A)),or(and(or())))` should have the same result set as `query.where(A)`
* `query.where(or(and(or(A)),and(or(and())))` should have the same result set as `query.where(A)`

There's some logic implemented in the C++ layer for these scenarios and so we should test them here.

firestore/src/android/wrapper.h
```diff
@@ -56,7 +56,7 @@ class Wrapper {
   // Similar to a copy constructor, but can handle the case where `rhs` is null.
   explicit Wrapper(Wrapper* rhs);
 
-  jni::Env GetEnv() const;
+  static jni::Env GetEnv();
```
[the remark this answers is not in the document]
ehsannas: reverting this change seems to be fine on my machine

ehsannas (commented):
Almost there.

firestore/src/include/firebase/firestore/filter.h
```diff
@@ -0,0 +1,378 @@
...
+
+ private:
+  friend class Query;
+  friend class QueryInternal;
+  friend class FilterInternal;
+  friend bool operator==(const Filter& lhs, const Filter& rhs);
+  friend struct ConverterImpl;
+
+  static Filter And(const std::vector<Filter>& filters);
+  static Filter Or(const std::vector<Filter>& filters);
```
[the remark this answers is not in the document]
ehsannas: Denver commented on the API doc that it might be better to take the vector by value (as follows) so that the callers can choose to move the vector.
```cpp
  static Filter And(std::vector<Filter> filters);
  static Filter Or(std::vector<Filter> filters);
  ```
I think that's reasonable.

File: firestore/src/android/query_android.h [hunk 12]
```diff
@@ -71,6 +71,8 @@ class QueryInternal : public Wrapper {
    */
   virtual AggregateQuery Count() const;
 
+  Query Where(const Filter& filter) const;
+
```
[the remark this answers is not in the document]
ehsannas: Agreed.

firestore/integration_test_internal/src/aggregate_count_test.cc
```diff
@@ -757,15 +751,5 @@ TEST_F(AggregateCountTest,
...
-#if defined(__ANDROID__)
-TEST(QueryTestAndroidStub, Construction) {
-  testutil::AssertWrapperConstructionContract<Query>();
-}
-
-TEST(QueryTestAndroidStub, Assignment) {
-  testutil::AssertWrapperAssignmentContract<Query>();
-}
-#endif  // defined(__ANDROID__)
-
```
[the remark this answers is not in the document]
ehsannas: I see. Thanks

ehsannas (approved):
Thanks! The release notes bot is failing. I think you need to add something to release_build_files/readme.md

markarndt (approved):
I don't see any docstrings, but the release note is fine.

PR #1357, opened 2023-06-21
Firestore: Improve Android tests to use the new env() helper function rather than creating an Env on the stack

dconeybe:
Many of the Firestore integration tests for Android need an instance of `jni::Env`. The usual solution is create an instance as a local variable by simply declaring `jni::Env env`. This is a bit annoying because it adds clutter to the source code. Additionally, `jni::Env` instances are not allowed to be shared between threads, and having a local variable makes it easy to accidentally share instances with other threads.

To fix this, `FirestoreAndroidIntegrationTest` gains a new static member function named `env()` that returns a reference to an instance of `jni::Env`. By using a `thread_local` variable, it ensures that each thread gets its own instance. Then, (almost) all occurrences of `jni::Env env` have been removed from the tests, and where an instance was needed it just calls the `env()` function.

### DIFF
--- a/firestore/integration_test_internal/src/android/firestore_integration_test_android.cc
+++ b/firestore/integration_test_internal/src/android/firestore_integration_test_android.cc
@@ -42,7 +42,6 @@ void PrintTo(const Object& object, std::ostream* os) {
 }  // namespace jni
 
 using jni::Constructor;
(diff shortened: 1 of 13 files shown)


wu-hui (approved):
LGTM.

PR #1358, opened 2023-06-22
Clean up JNI task listeners per object

a-maurice:
Add logic to the products that use objects to clean any pending Task listeners on destruction of that object.  Refactor the string creation logic from Auth to a utility function so all the products can use the same.
***

***

### Type of Change
***

#### Notes

### DIFF
--- a/app/src/util.cc
+++ b/app/src/util.cc
@@ -342,5 +342,17 @@ std::vector<std::string> SplitString(const std::string& s,
   return split_parts;
 }
 
+std::string CreateApiIdentifier(const char* api_id, void* object) {
+  std::string created;
+  created.reserve(strlen(api_id) +
+                  16 /* hex characters in the pointer */ +
+                  1 /* null terminator */);
+  snprintf(&(created[0]), created.capacity(), "%s0x%016llx",
+           api_id,
+           static_cast<unsigned long long>(  // NOLINT
+               reinterpret_cast<intptr_t>(object)));
+  return created;
+}
+
 // NOLINTNEXTLINE - allow namespace overridden
 }  // namespace firebase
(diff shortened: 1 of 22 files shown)


app/src/util.cc
```diff
@@ -342,5 +342,15 @@ std::vector<std::string> SplitString(const std::string& s,
   return split_parts;
 }
 
+std::string CreateApiIdentifier(const char* api_id, void* object) {
```
jonsimantov, line 345: const void* object, if it's only using the pointer itself

jonsimantov, line 345: Suggestion: similar to what we do with FutureHandle, what about also having an ever-incrementing unsigned integer appended to the name? Or do we need CreateApiIdentifier to always return the same name for the same API/object?

jonsimantov, line 345: Can you add a unit test for this?

jonsimantov (changes requested):

app/src/util.cc
a-maurice: Done, and it even found a problem with the string math.

a-maurice: We don't, but I do think there is value in doing it based on address, if only for debugging purposes.  It is guaranteed to be unique as well, which is nice.

a-maurice: So the nice thing about the string, beyond the debugging help, is the whole memory management part of it.  I'm not sure how the best way to have a generic way to return a unique pointer that could then be easily deleted by the caller later without using a string.

jonsimantov: It might be worth appending an incrementing number to the existing string, if only in case you delete a Firebase library and then it happens to get recreated in the same memory location.

a-maurice: Ah, that's a good argument.  Added it.

jonsimantov (approved):

PR #1360, opened 2023-06-24
Fix PhoneAuthOptions Timeout overflow on some platforms

AlmostMatt:
### Description
On some platforms there was previously an issue with the code to convert TimeoutMilliseconds from uint32_t to java long, causing it to become a negative value and then getting rejected by Android validation.
Example: https://github.com/firebase/firebase-unity-sdk/issues/764

I was able to reproduce the original issue with the Unity Auth testapp, which calls VerifyPhoneNumber with a PhoneAuthOptions.
The visible error was: "VerifyPhoneNumber: builder faild to create PhoneAuhtOptions"
Looking into the android logcat, the internal error was: "We only support 0-120 seconds for sms-auto-retrieval timeout"

This change should fix this by having an explicit cast from uint32_t to jlong before passing the timeout into the java long constructor. This should make the code work regardless of whether jlong is a 32 bit or 64 bit number.
***
### Testing

I verified the fix by building the unity sdk with this change included and running the Unity auth testapp. VerifyPhoneNumber ran successfully on my device.
***

### Type of Change
***

#### Notes

### DIFF
--- a/release_build_files/readme.md
+++ b/release_build_files/readme.md
@@ -627,6 +627,11 @@ workflow use only during the development of your app, not for publicly shipping
 code.
 
 ## Release Notes
+### Upcoming Release
+-   Changes
+    - Auth (Android): Fixed an issue where VerifyPhoneNumber's internal
+      builder failed to create PhoneAuthOptions.
+
 ### 11.2.0
 -   Changes
     - General (Android): Update to Firebase Android BoM version 32.1.1.
--- a/release_build_files/readme.md
+++ b/release_build_files/readme.md
@@ -630,7 +630,7 @@ code.
 ### Upcoming Release
 -   Changes
     - Auth (Android): Fixed an issue where VerifyPhoneNumber's internal
-      builder failed to create PhoneAuthOptions.
+      builder failed to create PhoneAuthOptions with certain compiler settings.
 
 ### 11.2.0
 -   Changes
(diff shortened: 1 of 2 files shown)


release_build_files/readme.md
```diff
@@ -627,6 +627,11 @@ workflow use only during the development of your app, not for publicly shipping
 code.
 
 ## Release Notes
+### Upcoming Release
+-   Changes
+    - Auth (Android): Fixed an issue where VerifyPhoneNumber's internal
```
jonsimantov, line 632: How would this actually occur to a developer using our SDK? Would it always fail or only if they specified a timeout? Probably best to describe this from the developer's perspective.

AlmostMatt: The issue was reported by a developer in https://github.com/firebase/firebase-unity-sdk/issues/764.
I tried to phrase the readme description similar to the error message that was shown to the user.

I think it would always fail. VerifyPhoneNumber takes a PhoneAuthOptions which always has a timeout milliseconds field. Regardless of what they set (0 or 60 seconds), it would still overflow to some invalid value and the user gets an error.

That said, when I tested our C++ integration tests VerifyPhoneNumber succeeded even without this fix. I think this must be something like a different compiler that uses 32 bits for a long, and therefore no overflow occured.

jonsimantov: Ah ok. So maybe add "with certain compiler settings" to the end of the explanation.

jonsimantov (approved):

release_build_files/readme.md
AlmostMatt: Done

PR #1365, opened 2023-06-27
Use C++'s smart pointers std::unique_ptr and std::shared_ptr instead of our own.

jonsimantov:
Switch from firebase::UniquePtr (and flatbuffers::UniquePtr) to std::unique_ptr, and from firebase::SharedPtr to std::shared_ptr. Also updates the use of MakeUnique / MakeShared to their std::make_unique and std::make_shared equivalents.

This also removes app/meta/move.h, which provided Move() and Forward<>(); it will now use std::move and std::forward, in the <utility> header, instead. Firestore uses these a lot.

The purpose of this PR is to clean up our code base, using modern C++ smart pointers instead of our own equivalents, to make it easier to maintain in the future.

This PR also replaces rare usage of firebase::compat::Atomic with std::atomic (mainly in a Firestore test, since the main usage used to be inside our own SharedPtr class that is now deleted).

Important: buried in this change is a [modification to database/src/desktop/query_desktop.cc](https://github.com/firebase/firebase-cpp-sdk/pull/1365/files/5355b5b871426987232fd410a4153495b3710f7a#diff-196c04498fc2dbafb32b5dd5f5076451a6cb3ee4a8e8bf40c1c348453a986947R220-R236), since the Callback mechanism does not work with move-only types (such as std::unique_ptr). This change wraps the unique_ptr inside a shared_ptr, thus making it copyable. Inside the callback, it releases the raw pointer from the unique_ptr and wraps it in a new unique_ptr. This ensures that there is never a time in which the pointer memory is not managed somewhere (it's either owned by the shared_ptr's unique_ptr prior to the callback, or by the newly created unique_ptr after the callback).

***

Running integration tests in this PR, as well as the unit tests.
***

### Type of Change
***

#### Notes


app/CMakeLists.txt
```diff
@@ -180,7 +180,7 @@ set(FLATBUFFERS_FLATC_SCHEMA_EXTRA_ARGS
     "--no-union-value-namespacing"
     "--gen-generated"
     "--gen-object-api"
-    "--cpp-ptr-type" "flatbuffers::unique_ptr")
+    "--cpp-ptr-type" "std::unique_ptr")
```
jonsimantov, line 183: There are a few places like this where we switch Flatbuffers over to use std::unique_ptr.

app/CMakeLists.txt
```diff
@@ -351,10 +351,7 @@ add_library(firebase_app STATIC
     ${utility_HDRS}
     ${app_platform_HDRS}
     ${FIREBASE_GEN_FILE_DIR}/app/google_services_generated.h
-    memory/atomic.h
```
jonsimantov, line 354: Remove the old memory library's header files entirely. (It was a header-only library, with tests.)

app/src/app_common.cc
```diff
@@ -289,14 +289,14 @@ App* AddApp(App* app, std::map<std::string, InitResult>* results) {
...
-  UniquePtr<AppData> app_data = MakeUnique<AppData>();
+  std::unique_ptr<AppData> app_data = std::make_unique<AppData>();
   app_data->app = app;
   app_data->cleanup_notifier.RegisterOwner(app);
   if (!g_apps) {
-    g_apps = new std::map<std::string, UniquePtr<AppData>>();
+    g_apps = new std::map<std::string, std::unique_ptr<AppData>>();
     created_first_app = true;
   }
-  (*g_apps)[std::string(app->name())] = app_data;
```
jonsimantov, line 299: Because the old UniquePtr actually had a copy constructor (which technically acted as a move constructor 😱), there are quite a few places where the code had to be modified to perform an explicit move where it wasn't happening before.

database/src/desktop/view/event.h
```diff
@@ -90,16 +91,13 @@ struct Event {
...
   //
-  // This field is not used to access any data, it's only used
-  // to take ownership. To keep the code streamlined, any time the pointer is
-  // needed, owned or not, it will go through the event_registration field
-  // above.
-  UniquePtr<EventRegistration> event_registration_ownership_ptr;
+  // This field is not used to access any data, it's only used to take
+  // ownership. To keep the code streamlined, any time the pointer is needed,
+  // owned or not, it will go through the event_registration field above.
+  std::shared_ptr<EventRegistration> event_registration_ownership_ptr;
```
jonsimantov, line 100: This has been changed from a UniquePtr to a shared_ptr, because Events are copied around all over the place in this library. Making it shared is much safer than trying to modify the entire desktop RTDB library to use move semantics for events. (And in fact there are certain places where we can't, such as a vector.insert(.....) that must be able to copy elements, and not move them.

database/src/desktop/query_desktop.cc
```diff
@@ -213,17 +214,26 @@ void QueryInternal::RemoveAllValueListeners() {
 }
 
 void QueryInternal::AddEventRegistration(
-    UniquePtr<EventRegistration> registration, void* listener_ptr) {
+    std::unique_ptr<EventRegistration> registration, void* listener_ptr) {
   database_->AddEventRegistration(query_spec_, listener_ptr,
                                   registration.get());
+  // Wrap the registration pointer into a shared pointer, so we can pass it
+  // safely to the callback.
+  std::shared_ptr<std::unique_ptr<EventRegistration>> reg_wrapped =
```
jonsimantov, line 222: Note: this is the bit that's called out in the PR description.

database/src/desktop/query_desktop.cc
```diff
@@ -217,13 +217,14 @@ void QueryInternal::AddEventRegistration(
...
-      [](Repo::ThisRef ref, std::unique_ptr<EventRegistration> registration) {
+      [](Repo::ThisRef ref, EventRegistration* registration_released) {
         Repo::ThisRefLock lock(&ref);
+        std::unique_ptr<EventRegistration> registration(registration_released);
         if (lock.GetReference() != nullptr) {
           lock.GetReference()->AddEventCallback(std::move(registration));
         }
       },
-      database_->repo()->this_ref(), std::move(registration)));
+      database_->repo()->this_ref(), registration.release()));
```
chkuang-g: Two things:

First, I think we can move  `std::unique_ptr<EventRegistration> registration(registration_released);` one line above. Still have that gap of time but just slightly shorter. 

Second, I think the future fellows, including me, will feel confused about this block of code.
Better to add a comment like:
```
// Workaround to pass EventRegistration into scheduler callback since currently NewCallback  "copy", instead of "move",  every parameters.
// TODO(some bug number): Refactor NewCallback to support move operator.
```

jonsimantov: Take a look at the [new version](https://github.com/firebase/firebase-cpp-sdk/pull/1365/files/5355b5b871426987232fd410a4153495b3710f7a#diff-196c04498fc2dbafb32b5dd5f5076451a6cb3ee4a8e8bf40c1c348453a986947R220-R236) of this fix, it's much safer than the workaround above, and I think it's probably fine to keep as-is.

jonsimantov: I added a more complete comment about what is happening here, and added the TODO.

dconeybe (commented):
LGTM for Firestore. I'll leave the formal LGTM to another reviewer with context to review the rest of the code.

jonsimantov:
This is ready for review, all integration and unit tests should now be working.

app/tests/callback_test.cc
```diff
@@ -290,13 +290,13 @@ TEST_F(CallbackTest, CallCallbackValue2String1) {
   EXPECT_THAT(value_and_string_.second, Eq("meaning"));
 }
 
-// Call a callback passing the UniquePtr
+// Call a callback passing the std::unique_ptr
```
a-maurice, line 293: I see this tests using a unique_ptr in the Callback logic, but we were concerned about that wrt how query_desktop was using it.  Any concerns?  Should we add a test for the whole shared wrapping unique thing?

jonsimantov: Good idea, I'll add it.

jonsimantov: Test added, PTAL.

a-maurice (approved):

PR #1368, opened 2023-06-28
Improve the iOS simulator install logic

a-maurice:
Add logic to check which versions of iOS/tvOS are available, and fallback to the latest if the requested one isn't available. Also change to using xcodes to install iOS simulators, since xcode-install has been deprecated in favor of it.

This is the same change as https://github.com/firebase/firebase-unity-sdk/pull/774, but for the C++ repo.
***

***

### Type of Change
Place an `x` the applicable box:
- [ ] Bug fix. Add the issue # below if applicable.
- [ ] New feature. A non-breaking change which adds functionality.
- [x] Other, such as a build process or documentation change.
***

#### Notes
- Bug fixes and feature changes require an update to the `Release Notes` section of `release_build_files/readme.md`.
- Read the contribution guidelines [CONTRIBUTING.md](https://github.com/firebase/firebase-cpp-sdk/blob/main/CONTRIBUTING.md).
- Changes to the public API require an internal API review. If you'd like to help us make Firebase APIs better, please propose your change in a feature request so that we can discuss it together.

PR #1371, opened 2023-06-28
Add a GitHub warning to the iOS simulator logic

a-maurice:
Adds a GitHub warning to the iOS simulator logic, as a follow up to https://github.com/firebase/firebase-cpp-sdk/pull/1368
***

***

### Type of Change
***

#### Notes

### DIFF
--- a/scripts/gha/test_simulator.py
+++ b/scripts/gha/test_simulator.py
@@ -512,6 +512,8 @@ def _create_and_boot_simulator(apple_platform, device_name, device_os):
     # If the requested version is available, use it, otherwise default to the latest
     if (device_os not in available_versions):
       logging.warning("Unable to find version %s, will fall back to %s", device_os, available_versions[-1])
+      if FLAGS.ci:
+        print("::warning ::Unable to find version %s, will fall back to %s" % (device_os, available_versions[-1]))
       device_os = available_versions[-1]
 
     args = ["xcodes", "runtimes", "install", "%s %s" % (apple_platform, device_os)]
--- a/scripts/gha/test_simulator.py
+++ b/scripts/gha/test_simulator.py
@@ -513,7 +513,7 @@ def _create_and_boot_simulator(apple_platform, device_name, device_os):
     if (device_os not in available_versions):
       logging.warning("Unable to find version %s, will fall back to %s", device_os, available_versions[-1])
       if FLAGS.ci:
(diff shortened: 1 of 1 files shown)


sunmou99 (commented):
`echo` works https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message
Not sure if `print()` works here. Should run some test or we have other examples in scripts.

Another nit, we could add check `if  available_versions:`.

jonsimantov (approved):

scripts/gha/test_simulator.py
```diff
@@ -512,6 +512,8 @@ def _create_and_boot_simulator(apple_platform, device_name, device_os):
     # If the requested version is available, use it, otherwise default to the latest
     if (device_os not in available_versions):
       logging.warning("Unable to find version %s, will fall back to %s", device_os, available_versions[-1])
+      if FLAGS.ci:
+        print("::warning ::Unable to find version %s, will fall back to %s" % (device_os, available_versions[-1]))
```
jonsimantov, line 516: Since this warning is outside of context, could you add text to explain that it's regarding iOS simulator version?

jonsimantov (changes requested):
(See comment about log message.)

a-maurice:
> `echo` works https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message Not sure if `print()` works here. Should run some test or we have other examples in scripts.
> 
> Another nit, we could add check `if available_versions:`.

The problem here is that this is the python script, so no access to echo.  We use print in a couple of other scripts, so hopefully that does actually work.

scripts/gha/test_simulator.py
a-maurice: Done

a-maurice:
> > `echo` works https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message Not sure if `print()` works here. Should run some test or we have other examples in scripts.
> > Another nit, we could add check `if available_versions:`.
> 
> The problem here is that this is the python script, so no access to echo. We use print in a couple of other scripts, so hopefully that does actually work.

I did a quick test, and confirmed that using print generates a warning card on the action level, similar to echo.

jonsimantov (approved):

PR #1379, opened 2023-06-30
Run iOS simulator install as root

a-maurice:
The xcodes install can fail when not run as root, so do so.
***

***

### Type of Change
***

#### Notes

### DIFF
--- a/scripts/gha/test_simulator.py
+++ b/scripts/gha/test_simulator.py
@@ -517,7 +517,7 @@ def _create_and_boot_simulator(apple_platform, device_name, device_os):
         print("::warning ::Unable to find %s version %s, will fall back to %s" % (apple_platform, device_os, available_versions[-1]))
       device_os = available_versions[-1]
 
-    args = ["xcodes", "runtimes", "install", "%s %s" % (apple_platform, device_os)]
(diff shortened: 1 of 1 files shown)


scripts/gha/test_simulator.py
```diff
@@ -517,7 +517,7 @@ def _create_and_boot_simulator(apple_platform, device_name, device_os):
         print("::warning ::Unable to find %s version %s, will fall back to %s" % (apple_platform, device_os, available_versions[-1]))
       device_os = available_versions[-1]
 
-    args = ["xcodes", "runtimes", "install", "%s %s" % (apple_platform, device_os)]
+    args = ["sudo", "xcodes", "runtimes", "install", "%s %s" % (apple_platform, device_os)]
```
jonsimantov, line 520: Update the info log message below to say it uses sudo; that way, if a user is running this on their machine, they won't be surprised by the password prompt.

a-maurice: It prints out the args array, so the sudo should be included in the message

jonsimantov: I guess that's fine, though not quite as user-friendly as it could be. :P

jonsimantov (approved):

jonsimantov:
BTW the time this failed was when I ran "full" tests so you might need to do that...

PR #1385, opened 2023-07-13
Fix for gma android failing assert

nakirekommula:
### Description
Fix for gma android failing assert

***
### Testing
None.

***

### Type of Change
***

#### Notes
NativeAd's getIcon() for android and icon() for iOS are nullable API methods and this indicates that ads returned will not always guarantee the availability of an icon asset. This PR relaxes the restriction which expects an icon asset at all times and processes the contents of an icon asset, only when a valid icon image asset is returned by andriod/iOS layers.
Ref: https://developers.google.com/admob/ios/api/reference/Classes/GADNativeAd#icon

Also updates the integration test to look for a large image asset rather than an icon asset, as the test ad unit ids are always expected to return a valid large image asset on a successful ad load.

### DIFF
--- a/gma/integration_test/src/integration_test.cc
+++ b/gma/integration_test/src/integration_test.cc
@@ -919,18 +919,15 @@ TEST_F(FirebaseGmaTest, TestNativeAdLoad) {
     EXPECT_FALSE(result_ptr->response_info().ToString().empty());
 
     // Check image assets.
-    EXPECT_FALSE(native_ad->icon().image_uri().empty());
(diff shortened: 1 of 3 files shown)


jonsimantov (commented):
Could you add a more detailed description to the PR about what was causing the assert and what this fix does?

nakirekommula:
> Could you add a more detailed description to the PR about what was causing the assert and what this fix does?

Added a bit of explanation in the description about the failing assert.

jonsimantov (approved):

PR #1399, opened 2023-07-21
Revert "OR Query Implementation (#1335)"

tom-andersen:
This reverts https://github.com/firebase/firebase-cpp-sdk/pull/1335 due to the Android crashes that cannot be fixed without a corresponding fix in the Android SDK.


dconeybe:
I'm withholding my approval until we determine that something less sledgehammer-ish won't work, such as forward-fixing or disabling the problematic tests.

dconeybe (approved):

PR #1404, opened 2023-07-25
fix: Disable OR Query test until follow up PR fixes problem

tom-andersen:
Failing tests will be disabled in this PR.

Before next release, a second PR needs to be merged: https://github.com/firebase/firebase-cpp-sdk/pull/1405

Otherwise, OR Query needs to be rolled back prior to release.

b/293326241

### DIFF
--- a/firestore/integration_test_internal/src/filter_test.cc
+++ b/firestore/integration_test_internal/src/filter_test.cc
@@ -23,9 +23,7 @@ namespace {
 
 using FilterTest = FirestoreIntegrationTest;
 
-#if false
-
-TEST_F(FilterTest, CopyConstructorReturnsEqualObject) {
+TEST_F(FilterTest, DISABLED_CopyConstructorReturnsEqualObject) {
   const Filter filter1a = Filter::EqualTo("foo", FieldValue::Integer(42));
   const Filter filter2a = Filter::ArrayContainsAny(
       "bar", {FieldValue::Integer(4), FieldValue::Integer(2)});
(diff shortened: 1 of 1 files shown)


firestore/integration_test_internal/src/filter_test.cc
```diff
@@ -23,6 +23,8 @@ namespace {
 
 using FilterTest = FirestoreIntegrationTest;
 
+#if false
+
 TEST_F(FilterTest, CopyConstructorReturnsEqualObject) {
```
jonsimantov, line 28: This is OK, but a better way to do this is to rename the test to DISABLED_CopyConstructorReturnsEqualObject, this will ensure gtest reports there being 1 disabled test.

dconeybe: Please create a ticket in TaskFlow to track the work required to fix this test. Then, use `GTEST_SKIP` and specify a message that includes the buganizer ticket.

e.g. https://github.com/firebase/firebase-cpp-sdk/blob/b9a53cf29fefe8e351e8b6b53a238a8365aa8993/firestore/integration_test_internal/src/source_test.cc#L402-L404

tom-andersen: I have updated the PR with better way of disabling tests.

This is a temporary measure until another PR fixes the problem.

dconeybe (approved):

jonsimantov (approved):

firestore/integration_test_internal/src/filter_test.cc
```diff
@@ -39,6 +39,8 @@ TEST_F(FilterTest, CopyConstructorReturnsEqualObject) {
 }
 
 TEST_F(FilterTest, CopyAssignementReturnsEqualObject) {
+  GTEST_SKIP() << "b/293326241: The Android implementation is failing.";
```
dconeybe, line 42: I just remembered an even better solution: `SKIP_TEST_ON_ANDROID`. Please use that instead so that the test continues to run in the other target platforms.

https://github.com/firebase/firebase-cpp-sdk/blob/b9a53cf29fefe8e351e8b6b53a238a8365aa8993/testing/test_framework/src/firebase_test_framework.h#L185

PR #1422, opened 2023-08-08
Internal Updates.

nakirekommula:
### Description
Remove query parameters from native image urls that may download a webp image

***
### Testing
None.

***

### Type of Change
***

#### Notes
Remove query parameters from native image urls that may download a webp image

### DIFF
--- a/gma/src/android/native_ad_image_android.cc
+++ b/gma/src/android/native_ad_image_android.cc
@@ -19,7 +19,6 @@
 #include <jni.h>
 
 #include <memory>
-#include <regex>
 #include <string>
 
 #include "gma/src/android/gma_android.h"
@@ -74,7 +73,10 @@ NativeAdImage::NativeAdImage(
 
   // Images requested with an android user agent may return webp images. Trim
   // webp parameter from image url to get the original JPG/PNG image.
-  internal_->uri = std::regex_replace(internal_->uri, std::regex("-rw"), "");
+  std::size_t webp_pos = internal_->uri.find("-rw");
+  if (webp_pos != std::string::npos) {
+    internal_->uri.replace(webp_pos, 3, "");
+  }
 
   // NativeAdImage scale.
   jdouble j_scale =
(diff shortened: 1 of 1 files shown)


gma/src/android/native_ad_image_android.cc
```diff
@@ -71,6 +72,10 @@ NativeAdImage::NativeAdImage(
   FIREBASE_ASSERT(j_uri);
   internal_->uri = util::JniUriToString(env, j_uri);
 
+  // Images requested with an android user agent may return webp images. Trim
+  // webp parameter from image url to get the original JPG/PNG image.
+  internal_->uri = std::regex_replace(internal_->uri, std::regex("-rw"), "");
```
jonsimantov, line 77: Is there a reason to use a regex for something that's just a text replacement?

nakirekommula: No particular reason. regex_replace syntax seemed easier to read and performance is similar to a regular replace. 
Would you prefer something like this:  internal_->uri.replace(internal_->uri.find("-rw"),3,"");

jonsimantov: Yes, I'd prefer that - regex is a lot of overhead for just a text replacment.

(Also, if you do it the other way, definitely also check that "-rw" is in the string first to avoid errors.)

nakirekommula: That makes sense. Updated it in the latest commit. Thanks for the review!

jonsimantov (approved):
LGTM. Just to be sure, is there anyway that the URL could contain "-rw" in a different spot that should *not* be replaced? (meaning, does this need to replace only the last occurrence?)

nakirekommula:
> LGTM. Just to be sure, is there anyway that the URL could contain "-rw" in a different spot that should _not_ be replaced? (meaning, does this need to replace only the last occurrence?)

We checked internally to make sure that "-rw" is reserved for webp, but it's a valid point that there is a chance of "-rw" occurence elsewhere in the url. Updated it to find the last occurence and also verify that last occurence appears after "=". Hopefully this gates any unwanted "-rw" trimming. Thanks again!

Issue #793, opened 2021-12-15
[Bug] Linker errors when using LLVM toolchain on Windows


triplef:
### [REQUIRED] Please fill in the following fields:

  * Pre-built SDK from the [website](https://firebase.google.com/download/cpp) or open-source from this repo: pre-built from website
  * Firebase C++ SDK version: 8.8.0
  * Problematic Firebase Component: App, Auth
  * Other Firebase Components in use: –
  * Platform you are using the C++ SDK on: Windows
  * Platform you are targeting: Windows

### [REQUIRED] Please describe the issue here:

Using the LLVM toolchain (specifically lld-link) instead of MSVC (link.exe) results in linker errors "relocation against symbol in discarded section: ..." from lld-link (see logs below).

Using link.exe instead of lld-link works fine, and it does not matter whether the source was built with MSVC or clang-cl.

As we have technical requirements to use lld-link in our projects and this is blocking us from using the Firebase C++ SDK.

#### Steps to reproduce with sample project:

1. Download Visual Studio sample project: [FirebaseCppTest.zip](https://github.com/firebase/firebase-cpp-sdk/files/7717600/FirebaseCppTest.zip)
2. Build app

#### Steps to reproduce manually:

1. Create new C++ project in Visual Studio
2. Select "LLVM (clang-cl)" as Platform Toolset in project settings
3. Link against at least firebase_app.lib and optionally firebase_auth.lib (linking both results in more errors of the same kind)
4. Add some minimal code to reference these projects, e.g.
```
    App *app = App::Create();
    if (app) {
        Auth *auth = Auth::GetAuth(app);
        printf("Created auth: %p", auth);
    }
```
5. Build app

#### Error Logs

```
1>lld-link : error : relocation against symbol in discarded section: $LN10
1>>>> referenced by C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\VC\Tools\MSVC\14.29.30133\include\vector:803
[...]

PR #1427, opened 2023-08-16
Use llvm-objcopy to package Windows binaries

triplef:
### Description
Changes the packaging script to use LLVM binutils. This fixes object files corrupted by using `objcopy` by using `llvm-objcopy` instead. While the corrupted files were working fine when using the Visual Studio toolchain they were leading to errors when using the LLVM toolchain.

See https://github.com/firebase/firebase-cpp-sdk/issues/793#issuecomment-1676632194 for details.

***
### Testing
Needs to be run as part of the GitHub actions packaging. I will test builds once they are available.

***

### Type of Change
***

#### Notes


.github/workflows/cpp-packaging.yml
```diff
@@ -561,8 +561,8 @@ jobs:
               # MacOS: use LLVM binutils for both X64 and ARM64
               additional_flags+=(-L)
             elif [[ "${{ matrix.sdk_platform }}" == "windows" && "${variant}" == *"/x64/"* ]]; then
-              # Windows x64: force input and output target format
-              additional_flags+=(-f pe-x86-64,pe-bigobj-x86-64)
+              # Windows x64: force input and output target format and use LLVM binutils for correct objcopy (issue #793)
+              additional_flags+=(-f pe-x86-64,pe-bigobj-x86-64 -L)
             elif [[ "${{ matrix.sdk_platform }}" == "windows" && "${variant}" == *"/x86/"* ]]; then
               # Windows x86: force input and output target format
               additional_flags+=(-f pe-i386,pe-bigobj-i386)
```
aganea, line 568: I would add `-L` for the 32-bit version as well.

.github/workflows/cpp-packaging.yml
```diff
@@ -561,8 +561,8 @@ jobs:
               # MacOS: use LLVM binutils for both X64 and ARM64
               additional_flags+=(-L)
             elif [[ "${{ matrix.sdk_platform }}" == "windows" && "${variant}" == *"/x64/"* ]]; then
-              # Windows x64: force input and output target format
-              additional_flags+=(-f pe-x86-64,pe-bigobj-x86-64)
+              # Windows x64: force input and output target format and use LLVM binutils for correct objcopy (issue #793)
+              additional_flags+=(-f pe-x86-64,pe-bigobj-x86-64 -L)
```
aganea, line 565: I think you need to remove `-f pe-x86-64,pe-bigobj-x86-64` because `llvm-objcopy` don't support triples written this way, and it works without that anyway.

.github/workflows/cpp-packaging.yml
```diff
@@ -561,8 +561,8 @@ jobs:
               # MacOS: use LLVM binutils for both X64 and ARM64
               additional_flags+=(-L)
             elif [[ "${{ matrix.sdk_platform }}" == "windows" && "${variant}" == *"/x64/"* ]]; then
-              # Windows x64: force input and output target format
-              additional_flags+=(-f pe-x86-64,pe-bigobj-x86-64)
+              # Windows x64: force input and output target format and use LLVM binutils for correct objcopy (issue #793)
```
jonsimantov, line 564: Rather than installing llvm-objcopy from binary, I'd rather compile it from source as we do on Mac above.

aganea: Makes sense. Not why we're building with `-DLLVM_ENABLE_PROJECTS=clang` though, `-DLLVM_ENABLE_PROJECTS=llvm` would be enough since only llvm binaries are being built.

triplef: Happy to change this as needed as I’m not sure I fully get the whole setup, but what’s the benefit of building these LLVM tools from source vs. simply installing the LLVM package in `install_prereqs_desktop.py`?

jonsimantov: Because this way we can be sure exactly which binary formats the objcopy command supports. In some cases, objcopy on platforms only supports a subset of binary formats (which would be relevant to the platform). By building it ourselves, we can be sure that it supports all of the binary formats we need.

triplef: Thanks, I’ve changed it so it should be using self-built LLVM binutils.

triplef:
@jonsimantov I hope I made the right changes based on your feedback - if not please let me know. Any chance you could trigger the CI to run with this branch?

jonsimantov:
I'll go ahead and test the packaging process with these changes. To do that I believe I'll have to copy these changes into a branch within this repo.

jonsimantov:
Invalid workflow file: .github/workflows/cpp-packaging.yml#L130
The workflow is not valid. .github/workflows/cpp-packaging.yml (Line: 130, Col: 16): Unexpected symbol: '['. Located at position 43 within expression: matrix.tools_platform == 'darwin' && join(['-', env.xcodeVersion]) || ''

triplef:
Thanks for checking – can you try again with the latest changes now please?

jonsimantov:
OK, here goes: https://github.com/firebase/firebase-cpp-sdk/actions/runs/6190110457

jonsimantov:
Fixed a missing quote, round 2: https://github.com/firebase/firebase-cpp-sdk/actions/runs/6190946332

triplef:
Thanks @jonsimantov! I tried the build and it works great for me using the LLVM toolchain.

One thing I noticed is that the firebase_cpp_sdk_windows.zip artifact itself contains another ZIP file, but I don’t think that stems from these changes?

jonsimantov:
Unfortunately, the integration tests failed to build against the newly packaged SDK on Windows.

I've attached the log file.

[8_Build integration tests.txt](https://github.com/firebase/firebase-cpp-sdk/files/12622409/8_Build.integration.tests.txt)

triplef:
Hmm ok, so the issue here is all these `error LNK2001: unresolved external symbol` errors, right? Any idea how they could be caused by this change?

Is this also a new issue: `Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed`?

triplef:
@aganea would you be able to take a look at these errors? Unfortunately I’m a bit lost about how this change might be causing them.

triplef:
I rebased the branch onto the latest main.

@jonsimantov could you please run the tests once more to see if these linker errors are still happening? 🙏

jonsimantov:
It's still failing to link. Log attached:
[8_Build integration tests.txt](https://github.com/firebase/firebase-cpp-sdk/files/13895048/8_Build.integration.tests.txt)

triplef:
Thank you @jonsimantov! It seems like the packaged library is missing a bunch of symbols.

Since the `-L` flag we’re adding in this patch is also making the packaging script use `llvm-nm` and `llvm-ar` (in addition to `llvm-objcopy`), I’m wondering if those are somehow behaving differently to cause this. Do they require some extra flags maybe – any idea @aganea?

Here is the same log without timestamps as well as the one from the latest run off the main branch for comparison:
- [8_Build.integration.tests-llvm.txt](https://github.com/firebase/firebase-cpp-sdk/files/13899221/8_Build.integration.tests-llvm.txt)
- [8_Build integration tests-main.txt](https://github.com/firebase/firebase-cpp-sdk/files/13899222/8_Build.integration.tests-main.txt)

The `Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed` is also present in the latest run and thus not a new issue.

PR #1439, opened 2023-09-07
Internal Updates

nakirekommula:
### Description
Add API to fetch adchoices attribution logo for native.

***
### Testing
N/A

***

### Type of Change
***

#### Notes
Internal / Experimental Change.


gma/src/common/native_ad.cc
```diff
@@ -85,6 +85,10 @@ const std::vector<NativeAdImage>& NativeAd::images() const {
   return internal_->images();
 }
 
+const NativeAdImage& NativeAd::adchoices_icon() const {
+  return internal_->adchoices_icon();
+};
```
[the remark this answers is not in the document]
nakirekommula: Ack.

gma/src/android/gma_android.cc
```diff
@@ -922,18 +923,30 @@ void JNI_NativeAd_completeLoadedAd(JNIEnv* env, jclass clazz, jlong data_ptr,
...
-                                           true, true);
+                                           "icon", true);
     env->DeleteLocalRef(j_icon);
   }
 
+  // getAdChoicesInfo().getImages() can return an empty list and a valid ad can
+  // exist without an adchoices icon image.
+  if (j_adchoices_icon != nullptr) {
+    NativeAdImageInternal adchoices_icon_internal;
+    adchoices_icon_internal.native_ad_image = j_adchoices_icon;
```
jonsimantov, line 934: Since j_adchoices_icon is passed into this callback, it's probably a local reference - does this need to become a global reference instead?

nakirekommula: While we perform InsertNativeInternalImage below on line 937, it creates a GlobalRef for j_adchoices_icon.
ref: https://github.com/firebase/firebase-cpp-sdk/blob/main/gma/src/android/native_ad_image_android.cc#L61

So, i believe we can have j_adchoices_icon as a local reference here and safely delete it, after a succesful InsertNativeInternalImage.

jonsimantov: OK, sounds good.

jonsimantov:
LGTM, just want to make sure - is there anything that we can add to the integration test for this?

jonsimantov (approved):

nakirekommula:
> LGTM, just want to make sure - is there anything that we can add to the integration test for this?

Thanks for the review Jon! Initially added a test for ad_choices icon as well, but excluded it later as we already have tests for image and icons which run a similar workflow as adchoices icon.

PR #1588, opened 2024-05-07
Revert support for avx2

kendalharland:
### Description

Passing `/arch:AVX2` to MSVC breaks the libcurl build in a strange way. It causes the error at the end curl's [nonblock.c](https://github.com/curl/curl/blob/master/lib/nonblock.c) file to throw, indicating our build is missing one of the defines in that file.  This change is being reverted until we can diagnose.

***
### Testing

- https://github.com/thebrowsercompany/firebase-cpp-sdk/actions/runs/8993073149
- https://github.com/thebrowsercompany/firebase-cpp-sdk/actions/runs/8993056021

***

### Type of Change
***


kendalharland:
Whoops, wrong repo

Issue #1703, opened 2025-03-31
[FR] Add `SUM` Functionality to AggregateQuerySnapshot


KKimj:
### Feature proposal

**Firebase Component:** Firestore  

Hello,
The Firestore C++ SDK currently supports only the `count()` method in the `firebase::firestore::AggregateQuerySnapshot` class but lacks a `sum()` method (and `avg()`), which is essential for aggregating numeric fields across documents.  

### Current State  
The public functions of `AggregateQuerySnapshot` are:  
- `count()`  
- `is_valid()`  
- Assignment operators (`operator=`)  
- `query()`  

Reference:  
- [API Documentation](https://github.com/firebase/firebase-cpp-sdk/blob/main/firestore/src/include/firebase/firestore/aggregate_query_snapshot.h#L109)  

### Flutter SDK Implications  
This enhancement would also benefit the Flutter SDK, which relies on the Firestore C++ SDK for Windows platform support. For example, in the [FlutterFire plugin](https://github.com/firebase/flutterfire/blob/main/packages/cloud_firestore/cloud_firestore/windows/cloud_firestore_plugin.cpp#L1424-L1426), unsupported aggregation types like `sum` currently result in:  
```dart
...
      case AggregateType::sum:
        std::cout << "Sum is not supported on C++" << std::endl;
        break;
...
```

Adding `sum()` to the C++ SDK would also benefit Flutter developers using Firestore on Windows.

### Benefits  
Implementing a `sum()` method would:  
1. Provide access to Firestore's aggregation capabilities in the C++ SDK.
2. Support more use cases for C++ and Flutter developers.

Looking forward to hearing thoughts and feedback!  
Thanks,

cherylEnkidu:
Hi @KKimj , thank you for filling out the ticket. This has been added to the feature request list.

jwinarske:
Heads up for anyone picking this up: on desktop the core already does the work.
The Firestore core the SDK vendors (firebase-ios-sdk, pinned in
`cmake/external/firestore.cmake`) carries aggregations end to end. From the
built `libfirestore_core.a`:

    api::Query::Aggregate(std::vector<model::AggregateField>&&) const
    api::AggregateQuery::GetAggregate(callback<StatusOr<model::ObjectValue>>)
    core::FirestoreClient::RunAggregateQuery(...)

and `model::AggregateField` is `enum class OpKind { Sum, Avg, Count }`.

[...]

PR #1705, opened 2025-04-01
[WIP] feat: Add AggregateField and Sum with Query

KKimj:
https://github.com/firebase/firebase-cpp-sdk/issues/1703

***

***

### Type of Change
***

#### Notes

### DIFF
--- a/firestore/integration_test_internal/src/aggregate_query_test.cc
+++ b/firestore/integration_test_internal/src/aggregate_query_test.cc
@@ -337,6 +337,21 @@ TEST_F(AggregateQueryTest, TestHashCode) {
             AggregateQueryHash(query1.Count()));
 }
 
+TEST_F(AggregateQueryTest, SumReturnsCorrectSum) {
+  CollectionReference collection =
+      Collection({{"a", {{"value", FieldValue::Integer(1)}}},
+                  {"b", {{"value", FieldValue::Integer(2)}}},
+                  {"c", {{"value", FieldValue::Integer(3)}}}});
+
+  AggregateQuery aggregate_query =
+      collection.Sum("value");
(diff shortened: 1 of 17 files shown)


jwinarske:
One thing to fix before this goes further — `QueryInternal::Aggregate` in
`firestore/src/main/query_main.cc`:

```cpp
AggregateQuery QueryInternal::Aggregate(const AggregateField& aggregate_field) {
  // TODO: Implement Aggregate
  // return MakePublic(aggregate_field);
}
```

That falls off the end of a non-void function, which is UB — with optimization
on you get whatever happens to be in the return slot, and `AggregateQuery`'s
destructor then runs on it. It's also the desktop path, so as it stands this PR
adds SUM for Android and leaves desktop worse than untouched.

The good news is that the desktop implementation is small. The vendored core
already has everything:

    api::Query::Aggregate(std::vector<model::AggregateField>&&) const
    api::AggregateQuery::GetAggregate(callback<StatusOr<model::ObjectValue>>)
    model::AggregateField::OpKind { Sum, Avg, Count }

So `Aggregate` is a matter of building the `AggregateField` vector with an
alias, calling `query_.Aggregate(...)`, and reading the result back out of the
`ObjectValue` by that alias in `AggregateQueryInternal::Get` — the existing
`Get` uses the count-specific `Get(StatusOr<int64_t>)` callback, which won't
serve a sum.

Two details that cost me time: a sum of integers comes back as
`integer_value` and a sum of doubles as `double_value`, so you need both; and
a missing alias means the aggregation matched nothing rather than summed to
zero.

I have this working on desktop against the Firestore emulator as a downstream
patch. It's desktop-only and doesn't use the `AggregateField` public shape this
PR introduces, so it isn't a drop-in, but happy to share it if it's useful.

PR #1737, opened 2025-06-18
Refactor ForEachAppDelegateClass for iOS into a new function that swizzles [UIApplication setDelegate:] to obtain App Delegate classes.

jonsimantov:
This commit refactors the `ForEachAppDelegateClass` function in `app/src/util_ios.mm`. Instead of scanning all runtime classes to find UIApplicationDelegate implementers, it now relies on method swizzling.

The `[UIApplication setDelegate:]` method is swizzled at startup. When `setDelegate:` is called, the class of the actual application delegate is captured and stored globally (static to the .mm file).

`ForEachAppDelegateClass` now uses this stored class. If it's called before `setDelegate:` has been invoked, the block passed to `ForEachAppDelegateClass` is queued. This queued block is then executed once the delegate is set via the swizzled `setDelegate:` method.

This approach is more efficient and directly targets the actual App Delegate class used by the application.

Key changes:
- Added `Firebase_setDelegate` C function as the swizzled implementation.
- Introduced `UIApplication(FirebaseAppDelegateSwizzling)` category with a `+load` method to perform the swizzling.
- Uses `method_setImplementation` for swizzling and stores the original IMP.
- Global static variables `g_app_delegate_class`, `g_original_setDelegate_imp`, and `g_pending_app_delegate_block` manage the state within `util_ios.mm`.
- Modified `ForEachAppDelegateClass` to use the new mechanism and queue blocks if the delegate is not yet known.

This change also allows you to set a special key in your `Info.plist` file to specify your App Delegate class name. If you set the `FirebaseAppDelegateClassName` to the name of your App Delegate class, it will bypass this whole setDelegate swizzling logic and simply use the App Delegate class you provided.

The purpose of this fix is to remove the dependency on `objc_copyClassList`, which was causing a crash on iOS 15. This crash was occurring because this function was forcing all Objective-C classes to be realized during the +load of the Firebase classes, which caused an issue with a common unrelated iOS framework. With this fix, `objc_copyClassList` is no longer used, which should also improve startup times on iOS.

***

Manually tested in C++ and Unity testapps.

***

### Type of Change
***

#### Notes


app/src/util_ios.mm
```diff
@@ -27,6 +25,98 @@
 #import <UIKit/UIKit.h>
 #import <objc/runtime.h>
 
+#define MAX_PENDING_APP_DELEGATE_BLOCKS 8
+#define MAX_SEEN_DELEGATE_CLASSES 32
+
+static IMP g_original_setDelegate_imp = NULL;
+// static Class g_app_delegate_class = nil; // Removed
```
jonsimantov: remove this

app/src/invites/ios/invites_ios_startup.mm
```diff
@@ -286,7 +286,7 @@ @implementation UIApplication (FIRFBI)
 + (void)load {
   // C++ constructors may not be called yet so call NSLog rather than LogDebug.
   NSLog(@"Loading UIApplication category for Firebase App");
-  ::firebase::util::ForEachAppDelegateClass(^(Class clazz) {
+  ::firebase::util::RunOnAppDelegate(^(Class clazz) { // Renamed here
```
jonsimantov: remove comment.

app/src/util_ios.h
```diff
@@ -188,7 +188,7 @@ typedef BOOL (
 // Call the given block once for every Objective-C class that exists that
 // implements the UIApplicationDelegate protocol (except for those in a
 // blacklist we keep).
-void ForEachAppDelegateClass(void (^block)(Class));
+void RunOnAppDelegate(void (^block)(Class));
```
jonsimantov: Maybe this should be renamed to RunOnAppDelegateClasses since it technically runs on the *classes*, not the delegate objects?

app/src/util_ios.h
```diff
@@ -188,7 +188,7 @@ typedef BOOL (
 // Call the given block once for every Objective-C class that exists that
```
jonsimantov: This comment should be changed to reflect the new behavior - calling the block for every objective-C class that's passed into [UIApplication setDelegate:]

Jules.md
```diff
@@ -305,6 +305,7 @@ API documentation.
     as mentioned in `CONTRIBUTING.md`.
 *   **Formatting**: Use `python3 scripts/format_code.py -git_diff -verbose` to
     format your code before committing.
+*   **Naming Precision for Dynamic Systems**: Function names should precisely reflect their behavior, especially in systems with dynamic or asynchronous interactions. For example, a function that processes a list of items should be named differently from one that operates on a single, specific item captured asynchronously. Regularly re-evaluate function names as requirements evolve to maintain clarity.
```
jonsimantov: Please word wrap this Markdown file to 80 characters, and put a self-referential note about this file that it should always be wrapped at 80 characters.

Jules.md
```diff
@@ -397,6 +399,10 @@ API documentation.
     integration, it can occasionally be a factor to consider when debugging app
     delegate behavior or integrating with other libraries that also perform
     swizzling.
+    When implementing or interacting with swizzling, especially for App Delegate methods like `[UIApplication setDelegate:]`:
+        *   Be highly aware that `setDelegate:` can be called multiple times with different delegate class instances, including proxy classes from other libraries (e.g., GUL - Google Utilities). Swizzling logic must be robust against being invoked multiple times for the same effective method on the same class or on classes in a hierarchy. An idempotency check (i.e., if the method's current IMP is already the target swizzled IMP, do nothing more for that specific swizzle attempt) in any swizzling utility can prevent issues like recursion.
```
jonsimantov: I think this idempotency check was removed during the iteration process, do you think we should add it back in?

app/src/util_ios.mm
```diff
@@ -76,41 +158,63 @@ - (BOOL)application:(UIApplication *)application
...
-      }
-      if (!blacklisted) {
-        if (GetLogLevel() <= kLogLevelDebug) {
-          // Call NSLog directly because we may be in a +load method,
-          // and C++ classes may not be constructed yet.
-          NSLog(@"Firebase: Found UIApplicationDelegate class %s", class_name);
-        }
-        block(clazz);
+void RunOnAppDelegateClasses(void (^block)(Class)) {
+  if (g_seen_delegate_classes_count > 0) {
```
a-maurice, line 274: Since a lot of these are running at start up time, any concern about multiple threads having problems here?  Like, can RunOnAppDelegateClasses be called by one thread while another is still iterating over the AppDelegates?

jonsimantov: No, I don't believe there is a risk here, all of the +load functions will only execute in the main thread, and [UIApplication setDelegate:] must be called from the main thread.

PR #1757, opened 2025-06-27
Implement UseUserAccessGroup for Firebase Auth C++ SDK

jonsimantov:
This commit adds the UseUserAccessGroup method to the Firebase Auth C++ SDK.

- On iOS, this method calls the native [FIRAuth useUserAccessGroup:error:] method to specify a user access group for iCloud keychain access.
- On other platforms (desktop, Android), this method is a no-op and returns kAuthErrorNone as it's an iOS-only feature.

Public API has been added to firebase::auth::Auth in auth.h, with implementations in auth_ios.mm (iOS) and auth_desktop.cc (stub for other platforms).

***

***

### Type of Change
***

#### Notes

### DIFF
--- a/auth/integration_test/src/integration_test.cc
+++ b/auth/integration_test/src/integration_test.cc
@@ -1513,4 +1513,18 @@ TEST_F(FirebaseAuthTest, TestLinkFederatedProviderBadProviderIdFails) {
 
 #endif  // defined(ENABLE_OAUTH_TESTS)
 
+#if TARGET_OS_IPHONE
+TEST_F(FirebaseAuthTest, TestUseUserAccessGroup) {
+  // This is a simple smoke test to ensure the method can be called
+  // without crashing on iOS and returns the expected default success.
+  // Deeper testing of keychain access group functionality would require
+  // more complex setup and is typically done manually or with UI tests.
+  EXPECT_EQ(auth_->UseUserAccessGroup(nullptr),
+            firebase::auth::kAuthErrorNone);
+  EXPECT_EQ(auth_->UseUserAccessGroup("test-group"),
+            firebase::auth::kAuthErrorNone);
+  EXPECT_EQ(auth_->UseUserAccessGroup(""), firebase::auth::kAuthErrorNone);
+}
+#endif  // TARGET_OS_IPHONE
+
 }  // namespace firebase_testapp_automated
--- a/auth/integration_test/src/integration_test.cc
+++ b/auth/integration_test/src/integration_test.cc
@@ -1516,14 +1516,14 @@ TEST_F(FirebaseAuthTest, TestLinkFederatedProviderBadProviderIdFails) {
 #if TARGET_OS_IPHONE
 TEST_F(FirebaseAuthTest, TestUseUserAccessGroup) {
   // This is a simple smoke test to ensure the method can be called
-  // without crashing on iOS and returns the expected default success.
+  // without crashing on iOS.
   // Deeper testing of keychain access group functionality would require
   // more complex setup and is typically done manually or with UI tests.
-  EXPECT_EQ(auth_->UseUserAccessGroup(nullptr),
-            firebase::auth::kAuthErrorNone);
-  EXPECT_EQ(auth_->UseUserAccessGroup("test-group"),
-            firebase::auth::kAuthErrorNone);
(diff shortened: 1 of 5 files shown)


auth/src/desktop/auth_desktop.cc
```diff
@@ -575,6 +575,11 @@ void Auth::UseEmulator(std::string host, uint32_t port) {
   auth_impl->assigned_emulator_url.append(std::to_string(port));
 }
 
+AuthError Auth::UseUserAccessGroup(const char* access_group) {
```
jonsimantov, line 578: We need an Android stub as well.

auth/src/include/firebase/auth.h
```diff
@@ -517,6 +517,21 @@ class Auth {
   /// not available on the current device.
   static Auth* GetAuth(App* app, InitResult* init_result_out = nullptr);
 
+  /// @brief Specifies a user access group for iCloud keychain access.
+  ///
+  /// This method is only functional on iOS. On other platforms, it is a no-op
+  /// and will always return `kAuthErrorNone`.
+  ///
+  /// If you are using iCloud keychain synchronization, you will need to call
```
jonsimantov: Remove this sentence.

auth/src/include/firebase/auth.h
```diff
@@ -517,6 +517,21 @@ class Auth {
...
+  /// this method to set the user access group.
+  ///
+  /// @param[in] access_group The user access group to use. Set to `nullptr` or
+  /// an empty string to use the default access group.
```
jonsimantov: Let's pass in an empty string verbatim instead of treating it as nil.

auth/src/ios/auth_ios.mm
```diff
@@ -608,5 +608,23 @@ void DisableTokenAutoRefresh(AuthData *auth_data) {}
 void InitializeTokenRefresher(AuthData *auth_data) {}
 void DestroyTokenRefresher(AuthData *auth_data) {}
 
+AuthError Auth::UseUserAccessGroup(const char* access_group_str) {
+  if (!auth_data_) {
+    return kAuthErrorUninitialized;
```
jonsimantov: No such error, this one doesn't exist, please check the AuthError enum and suggest a better option.

auth/integration_test/src/integration_test.cc
```diff
@@ -1513,4 +1513,18 @@ TEST_F(FirebaseAuthTest, TestLinkFederatedProviderBadProviderIdFails) {
 
 #endif  // defined(ENABLE_OAUTH_TESTS)
 
+#if TARGET_OS_IPHONE
+TEST_F(FirebaseAuthTest, TestUseUserAccessGroup) {
+  // This is a simple smoke test to ensure the method can be called
+  // without crashing on iOS and returns the expected default success.
+  // Deeper testing of keychain access group functionality would require
+  // more complex setup and is typically done manually or with UI tests.
+  EXPECT_EQ(auth_->UseUserAccessGroup(nullptr),
```
jonsimantov: Let's just run the function but not check its results since it may not be configured for keychain sharing.

auth/integration_test/src/integration_test.cc
```diff
@@ -1513,4 +1513,18 @@ TEST_F(FirebaseAuthTest, TestLinkFederatedProviderBadProviderIdFails) {
 
 #endif  // defined(ENABLE_OAUTH_TESTS)
 
+#if TARGET_OS_IPHONE
```
jonsimantov, line 1516: Since we have this stubbed out on other platforms, let's make sure that works and run this test on all platforms.

auth/src/include/firebase/auth.h
```diff
@@ -517,6 +517,19 @@ class Auth {
...
   static Auth* GetAuth(App* app, InitResult* init_result_out = nullptr);
 
+  /// @brief Specifies a user access group for iCloud keychain access.
+  ///
+  /// This method is only functional on iOS. On other platforms, it is a no-op
+  /// and will always return `kAuthErrorNone`.
+  ///
+  /// @param[in] access_group The user access group to use. Set to `nullptr`
+  /// to use the default access group. An empty string will be passed as an
+  /// empty string.
```
jonsimantov, line 527: We don't need to call out that the empty string is passed in special.

PR #1775, opened 2025-07-16
feat: Remove trackingID from iOS

ncooke3:
Firebase 12 Apple SDK has been released and has removed the deprecated, unused `trackingID` property.
***

Did not test locally, relying on CI for full coverage.
***

### Type of Change
***

#### Notes

### DIFF
--- a/app/src/app_ios.mm
+++ b/app/src/app_ios.mm
@@ -71,10 +71,6 @@ static void PlatformOptionsToAppOptions(FIROptions* platform_options, AppOptions
     const char* value = platform_options.databaseURL.UTF8String;
     if (value) app_options->set_database_url(value);
   }
-  if (!strlen(app_options->ga_tracking_id())) {
-    const char* value = platform_options.trackingID.UTF8String;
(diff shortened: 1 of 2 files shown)


ncooke3:
@a-maurice, I see there are likely some other changes needed in https://github.com/firebase/firebase-cpp-sdk/blob/main/app/src/include/firebase/app.h

This seems potentially trickier than intended if the other platforms also expose this property. Does this PR seem feasible to do if it still exists on other platforms? I don't want to introduce too much complexity, so I'd be fine adding the property back upstream.

a-maurice:
> @a-maurice, I see there are likely some other changes needed in https://github.com/firebase/firebase-cpp-sdk/blob/main/app/src/include/firebase/app.h
> 
> This seems potentially trickier than intended if the other platforms also expose this property. Does this PR seem feasible to do if it still exists on other platforms? I don't want to introduce too much complexity, so I'd be fine adding the property back upstream.

Will see what needs to be done for the field on other platforms, but for now this should be good for iOS. Thanks!

Issue #1724, opened 2025-06-09
[Bug] [iOS][Storage] Download fails with NSCocoaErrorDomain Code=262 in GetFile() (C++ SDK 12.7.0–12.8.0, 13.0.0)


pbs0512:
### [REQUIRED] Please fill in the following fields:

  * Pre-built SDK from the [website](https://firebase.google.com/download/cpp) or open-source from this repo: **Pre-built SDK from website**
  * Firebase C++ SDK version: **12.7.0 / 12.8.0 / 13.0.0**
  * Problematic Firebase Component: **Storage**
  * Other Firebase Components in use: **Auth, Firestore, Functions, Analytics, Messaging, Remote Config**
  * Platform you are using the C++ SDK on: **macOS**
  * Platform you are targeting: **iOS**

---

## ❗️[iOS][Storage] `GetFile()` download fails with NSCocoaErrorDomain Code=262 on valid file path

Calling `StorageReference::GetFile()` on iOS fails consistently when attempting to download to a valid sandboxed absolute path. This does **not happen** on Android or Windows using the same C++ code and Firebase bucket, suggesting the problem lies in the iOS native layer of the C++ SDK.

---

### 📄 Error Log Example

```
Could not remove previous file at /private/var/mobile/Containers/Data/Application/1AADB80E-8E59-4531-83DF-FD292F7BD24C/tmp/CFNetworkDownload_6CZmV6.tmp due to Error Domain=NSCocoaErrorDomain Code=262 "The specified URL type isn’t supported for opening the file." UserInfo={NSFilePath=/var/mobile/Containers/Data/Application/1AADB80E-8E59-4531-83DF-FD292F7BD24C/Documents/patch/asset/100000_100001.zip, NSURL=/var/mobile/Containers/Data/Application/1AADB80E-8E59-4531-83DF-FD292F7BD24C/Documents/patch/asset/100000_100001.zip}
```

---

### 🔍 Relevant Code Snippet (C++ - Firebase Storage Download)

```cpp
auto app = firebase::App::GetInstance();
if (app == nullptr) return false;
auto storage = firebase::storage::Storage::GetInstance(app);
if (storage == nullptr) return false;
auto storage_ref = storage->GetReference(file_path);
firebase::storage::Controller controller;

auto download_task = storage_ref.GetFile(local_path.c_str(), g_storageListener, &controller);
download_task.OnCompletion([local_path, outputPath](const firebase::Future<size_t>& result) {
[...]

PR #1782, opened 2025-07-30
[Storage] Improve how GetFile handles paths

a-maurice:
On iOS, allow absolute paths to be passed in to GetFile, instead of just file URLs.
***

***

### Type of Change
***

https://github.com/firebase/firebase-cpp-sdk/issues/1724

#### Notes

### DIFF
--- a/storage/src/ios/storage_reference_ios.mm
+++ b/storage/src/ios/storage_reference_ios.mm
@@ -147,9 +147,12 @@ Future<size_t> StorageReferenceInternal::GetFile(const char* path, Listener* lis
     future_impl->Complete(handle, kErrorUnknown, "Path cannot be empty.");
     return GetFileLastResult();
   }
-  NSURL* local_file_url = [NSURL URLWithString:path_string];
-  if (local_file_url == nil) {
-    // If it failed to load as a URL, try loading it is a path.
+  NSURL* local_file_url = nil;
+  if ([path_string hasPrefix:@"file://"]) {
+    // If it starts with the prefix, load it assuming a URL string.
+    local_file_url = [NSURL URLWithString:path_string];
+  } else {
+    // Otherwise, assume it is a file path.
     local_file_url = [NSURL fileURLWithPath:path_string];
   }
   // If we still failed to convert the path, error out.


storage/integration_test/src/integration_test.cc
```diff
@@ -647,7 +647,8 @@ TEST_F(FirebaseStorageTest, TestPutFileAndGetFile) {
     std::string path = PathForResource() + kGetFileTestFile;
     // Cloud Storage expects a URI, so add file:// in front of local
     // paths.
-    std::string file_path = kFileUriScheme + path;
+    // TODO: Revert, just testing to see if this will pass without.
+    std::string file_path = /*kFileUriScheme +*/ path;
```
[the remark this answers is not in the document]
a-maurice: Added another test for both

storage/src/ios/storage_reference_ios.mm
```diff
@@ -142,7 +142,21 @@
...
+  } else {
+    // Otherwise, assume it is a file path.
+    local_file_url = [NSURL fileURLWithPath:path_string];
+  }
+  // If we still failed to convert the path, error out.
+  if (local_file_url == nil) {
+    future_impl->Complete(handle, kErrorUnknown,
+      "Unable to convert provided path to valid URL");
+    return GetFileLastResult();
+  }
```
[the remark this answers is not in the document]
a-maurice: Added a check for empty

storage/src/ios/storage_reference_ios.mm
```diff
@@ -142,7 +142,25 @@
...
+  NSString* path_string = @(path);
+  if (path_string.length == 0) {
+    future_impl->Complete(handle, kErrorUnknown, "Path cannot be empty.");
+    return GetFileLastResult();
+  }
+  NSURL* local_file_url = nil;
+  if ([path_string hasPrefix:@"file://"]) {
+    // If it starts with the prefix, load it assuming a URL string.
+    local_file_url = [NSURL URLWithString:path_string];
+  } else {
```
paulb777, line 154: Can this be breaking since before it always used `URLWithString`  and now it only uses `URLWithString` if there is a `file://` prefix?

a-maurice: I think previously it would fail later if you tried to use a different url type, since it is calling into Storage ObjC's writeToFile, which is documented to take in a fileUrl (though looking over the code, I don't immediately spot anything that enforces that).  The other possibility I was thinking of doing was trying `URLWithString` first, and if that returns `nil` fall back to `fileURLWithPath`.  Do you think that would make more sense (and it would resolve the problem you are describing)?

paulb777: Yep, that would seem like a safer change to do URLWithString first and fall back. However, I'm also ok as is, if you've tested the different combinations.

paulb777 (approved):

storage/src/ios/storage_reference_ios.mm
a-maurice: I'll change it to that then, just in case there is some odd case I'm not aware of.

a-maurice: So, I tried that, and it has it's own weird problems of URLWithString succeeding and generating not nil, but incorrect file urls from the full paths. So I think I will go with checking for "file://", since it does seem to be necessary for the writeToFile logic anyway.

PR #1795, opened 2025-10-21
Feat: add UseEmulator to storage api

AustinBenoit:
Add support for Firebase Storage emulator via `UseEmulator`.

***

Tested this manually via unity for Desktop, Android, iOS. See the photos below:

image

image

image
 
***

### Type of Change
***

#### Notes

### DIFF
--- a/release_build_files/readme.md
+++ b/release_build_files/readme.md
@@ -613,7 +613,7 @@ workflow use only during the development of your app, not for publicly shipping
 code.
 
 ## Release Notes
-### 13.3.0
+### Upcoming
 -   Changes
     - Storage: Add support for Firebase Storage emulator via `UseEmulator`.
         The `UseEmulator` method should be called before invoking any other
--- a/storage/src/desktop/storage_desktop.cc
+++ b/storage/src/desktop/storage_desktop.cc
@@ -23,6 +23,7 @@
 #include "app/src/app_common.h"
 #include "app/src/function_registry.h"
 #include "app/src/include/firebase/app.h"
+#include "app/src/log.h"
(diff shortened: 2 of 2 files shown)


release_build_files/readme.md
```diff
@@ -613,6 +613,12 @@ workflow use only during the development of your app, not for publicly shipping
 code.
 
 ## Release Notes
+### 13.3.0
```
a-maurice: We typically put "Upcoming" instead of a version number, just to make it clearer that these haven't gone out yet.

storage/src/desktop/storage_desktop.cc
```diff
@@ -134,6 +136,27 @@ void StorageInternal::CleanupCompletedOperations() {
   }
 }
 
+void StorageInternal::UseEmulator(const char* host, int port) {
+  if (host == nullptr || host[0] == '\0') {
+    throw std::invalid_argument("Emulator host cannot be null or empty.");
```
a-maurice: We generally don't use C++ exceptions, with the philosophy that using them requires users to enable them on their end, which isn't guaranteed. I think the better thing to do would be to LogError, and return.

release_build_files/readme.md
AustinBenoit: Changed

storage/src/desktop/storage_desktop.cc
AustinBenoit: Changed to return and send errors

a-maurice (approved):
LGTM, but should wait until the API review has been handled before submitting

cynthiajoan (approved):

